PR #11, opened 2015-08-13
added link to source code on password generator page

alexbooker:


alexbooker:
My editor automatically changed some formatting so the differential is a bit noisy - hope that's OK.

alexbooker:
Whops.

PR #13, opened 2015-08-14
Moved password generation to the client

alexbooker:
Hello,

I have ported the server-side [password generator](https://github.com/privacytoolsIO/privacytools.io/blob/master/pw.php) to JS.  It uses [*`RandomSource.getRandomValues()`*](https://developer.mozilla.org/en-US/docs/Web/API/RandomSource/getRandomValues) which is cryptographically secure - it *should* work in all [modern browsers](http://caniuse.com/#feat=getrandomvalues). I *could* make it work in older-browsers by falling back to [*ISAAC*](http://burtleburtle.net/bob/rand/isaacafa.html) or something but that would not be totally secure.

I also improved the UI and UX (imo):

(The alert is a normal Bootstrap `alert alert-info` but due to the GIF's limited colour palette, it looks different...)

I know you were concerned about users who disable JS. If someone visits this website without JS enabled, they are shown this message:

If you want, we can link those users the server-side password generator (e.g. append to that alert the sentence: *"If you want, you can use the server-side based solution that does not require JS"*.)

I am taking a bit of a risk here because you did not explicitly say: *"Yes, please do this"* but I feel **strongly** that password generation should be done on the client when possible, so do other people in the community. 

An additional benefit of doing this on the client-side is that, users can download `password.html` and use it locally without the need to run a local PHP server.

I am looking forward to hear what you think, @privacytoolsIO.


alexbooker:
I think that is good enough for a version 1.0.

alexbooker:
What do you think, @privacytoolsIO? I know I am taking a bit of a risk here because you did not give the go-ahead but I am hopeful that we can make this work.

privacytoolsIO:
Great work, alexbooker! I just changed a couple of things, please review it. Your new generator is already online, and old links redirect also to the new generator. Thanks :)

alexbooker:
_Awesome!_

There is only one more thing to do, I think - make the "Source code" link point to the [source file](https://github.com/privacytoolsIO/privacytools.io/blob/master/password.html). I'll do that now and commit directly to `master`. I'll ping you here once it's done so you can update the server.

alexbooker:
OK, @privacytoolsIO - I did it.

PR #15, opened 2015-08-26
add a tab_undo suggestion to firefox tweaks

matthewdavis:
I don't know of any way to pull this info remotely via javascript or otherwise. But it was a bit surprising to see a list of my previouslly closed tabs even while using private browsing. As seen here.

blah

### DIFF
--- a/index.html
+++ b/index.html
@@ -812,7 +812,7 @@ This is a collection of privacy related <strong>about:config</strong> tweaks. We
 <ul><li>Website owners can track the battery status of your device. <a href="https://www.reddit.com/r/privacytoolsIO/comments/3fzbgy/you_may_be_tracked_by_your_battery_status_of_your/" target="_blank">Source</a></li></ul>
 
 <li>browser.sessionstore.max_tabs_undo = 0</li>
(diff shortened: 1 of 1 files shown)


alexbooker:
Indeed, that is quite disconcerting - thank you for bringing it up.

I do not use Firefox personally, so I'll leave it up to @privacytoolsIO  to review.

alexbooker:
Just a couple of suggestions:

- Use an upper-case *F* for the proper noun *FireFox*.
- Could be nice to mention where the history surfaces (i.e. in the recently closed tab list).

And one concern:

- Hm. It sounds like the history is stored for the duration of the session. I *imagine* that this is required to power the "undo" button. If you set `max_tabs_undo = 0` does the "undo" button still work? If it does, someone could see your session history by hitting the "undo"  button *still*. It would also mean the history is stored in memory still. If the "undo" button does *not* work, then this sounds like it comes at too high of a cost to usability and *probably* is not something to recommend by default.

matthewdavis:
Thanks for the suggestions Alex, I modified the wording a bit.

PR #16, opened 2015-08-28
Update ProtonMail Storage

TonyZTan:
ProtonMail Storage is now 500 MB for new users during the Beta, with the possibility to request more if needed. (It is still 1 GB for users who registered before the change.)

More space will be available in the future as a premium option.

https://support.protonmail.ch/knowledge-base/increase-storage-space/


privacytoolsIO:
thanks!

PR #26, opened 2016-01-19
Free option column

OKNoah:
Adds “free option” column to emails and clarifies data.

PR #30, opened 2016-02-25
Free option column

OKNoah:
Adds the class label-primary to values in email table.

See PR #26

### DIFF
--- a/index.html
+++ b/index.html
@@ -208,17 +208,17 @@ Over the last 16 months, as I've debated this issue around the world, every sing
 	<ul><li>Browsing history and open websites.</li></ul>
 	<ul><li>Apps settings.</li></ul>
 	<ul><li>WiFi hotspot names and passwords.</li></ul>
-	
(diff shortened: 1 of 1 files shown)


alexbooker:
@OKNoah I'm only here to assist @privacytoolsIO with the workload. I would like to know his/her thoughts before we proceed. Sorry about all this churn.

@privacytoolsIO I will handle the merge conflict if you agree this should be merged? I'm not sure if your objection was to the change _in general_, or the way in which the change was applied in #26.

alexbooker:
Ping @privacytoolsIO

alexbooker:
Sorry, @OKNoah but I don't think this pull request is going anywhere. I really appreciate you taking the time to contribute to [privacytools.io](https://privacytools.io) but I'm going to close this now.

In the future please create an issue or post to Reddit before writing any code to make sure this is the direction we (the community) want to take.

PR #38, opened 2016-04-12
Fix JonDo Live - mention end of support

m3t:
https://anonymous-proxy-servers.net/blog/index.php?/archives/438-End-of-Live-DVD.html


alexbooker:
Thank you, @m3t :smile:!

What do you think about removing this link entirely?

m3t:
As long as the download is still available ...
Maybe some folks want to use it for research :smirk: (or exchange psc to bitcoin with its tools on their  [website](https://shop.anonymous-proxy-servers.net/bin/btc-shop))

PR #40, opened 2016-04-18
Updating logos

conorohiggins:
Encryptr and Signal didn't have the right logos next to them.


alexbooker:
Thanks :smile: 

It looks like the Signal logos are different for Android and iOS.

Android:



iOS:



I think they're similar enough we can pick one for the website so I'll merge this for now.

It looks like Encryptr was acquired by SpiderOak and the logo changed -- thanks for the update. I also updated the link in commit b4493c26b747dd1d9c9ca2145877c49b89d03346.

PR #46, opened 2016-04-27
menor fix

victorhck:
just only a double ".." 
sorry for the noise!! ;)

've phun!


alexbooker:
No contribution is too small. Thanks :smile:!

PR #51, opened 2016-05-12
Update site title to be more consistent across pages

mitchellcash:
Sorry, I use a text editor add-on that removes white space automatically. It seems I have inadvertently done that here, let me know if you want me to rebase leaving the whitespace but I think it's good to remove in this commit.

### DIFF
--- a/donate.html
+++ b/donate.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-<title>Please Donate to privacytools.io</title>
+<title>Please Donate - privacytools.io</title>
 <link href="css/bootstrap.min.css" rel="stylesheet">
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
 <link rel="shortcut icon" href="favicon.ico" type="image/ico">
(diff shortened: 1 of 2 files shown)


alexbooker:
Hi again, @mitchellcash :wave: 

I agree that the site title should be consistent across pages so I merged your branch. Thanks for the PR :smile: 

I think I'll make some tweaks later. For example, the title used to be _"privacy tools - encryption against global mass surveillance 🔒"_ and now it's _"Encryption against global mass surveillance 🔒 - privacy tools"_. I don't think the _"-"_ was stylistic - I think @privacytoolsIO used grammar intentionally.

Moreover, I think the site name, _"Privacy Tools"_, should be visible on tabs. As you can see, most browsers truncate the tab title:



By putting the text _"Privacy Tools"_ at the end of a long title it almost certainly won't be seen.

To be honest, I'm not even sure the text _"Encryption against global mass surveillance"_ even belongs in `<title>` - it's not a `<title>`, it's a description. Perhaps it would better to make the `<title>` just _"Privacy Tools"_ and and in the case of the donate page, _"Donate | Privacy Tools"_.

What do you think?

mitchellcash:
Completely agree. That was my original thought, but as I am new to committing to this project I thought I would ease myself in with smaller changes :)

My thoughts would be:
Main page: Privacy Tools
Subsequent pages: "*Page Title*" | Privacy Tools (e.g. Secure Password Generator | Privacy Tools)

We should also consider updating the site description, as it's currently truncated on Google's search engine:



So maybe this is where we instead add:
Site description: `You are being watched! Knowledge, encryption and privacy tools to protect you against  global mass surveillance.`

PR #53, opened 2016-05-15
Introduce flag-icon-css and display select country flags

mitchellcash:


alexbooker:
I think this is a nice addition. 

:+1: for using SVG images but can you please put them in the `img` folder and remove the ones we don't reference? Also, can you please link the source of the flags that I can check the licence. 

Thanks :smile:

mitchellcash:
I can do.

With removing the ones we don't reference, I left them all there in case there were other parts of the project that could use them. What do you think?

It's an MIT licence. I'm no expert but I think it means we are safe to use?

https://github.com/lipis/flag-icon-css/blob/master/LICENSE

alexbooker:
> With removing the ones we don't reference, I left them all there in case there were other parts of the project that could use them. What do you think?

My concern is that there are **500** icons which require 6mb. We use just 14 :scream: 

... It's a bit too unwieldy (relatively speaking) in terms of **1,** deployment and **2,** managing the Git repository.

Adding new flags in the future shouldn't be too much bother.

mitchellcash:
Okay, agreed. Let me fix this commit

mitchellcash:
@alexbooker updated and squashed commit, can you please review?

alexbooker:
More awesome work :fire:. 
Merged :smile:!

mitchellcash:
Woo, thanks @alexbooker

privacytoolsIO:
Good job, guys! Thanks

PR #54, opened 2016-05-19
Remove foreign link

mitchellcash:
Not sure what this link was doing in the first place, but as it’s not in English (which I understand is the target audience), I think we should remove it.

Also I guess remove a tonne of white space but see line 648 for link removal.


alexbooker:
Thanks :+1:

PR #57, opened 2016-07-20
Minor edit to 'ChatSecure' description

beardog108:
Facebook is no longer an option for ChatSecure, because they discontinued their XMPP service in April 2015. I have changed the ChatSecure description to reflect this. [Source](https://developers.facebook.com/docs/chat/).

Side note: It says that one should 'post suggestions' in the subreddit, which I do use, however I take that to mean suggestions for adding or removing software/advice, not minor edits like this. I apologize if minor edits should also be discussed in the sub.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1515,7 +1515,7 @@
 						<h3 class="panel-title">Off-the-Record: ChatSecure</h3>
 					</div>
 					<div class="panel-body">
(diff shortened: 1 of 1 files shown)


mitchellcash:
@beardog108 anecdotally minor edits have been fine without posting in the subreddit :)

alexbooker:
Thanks :smile:

Merged and should be online soon!

Issue #80, opened 2016-10-21
Remove CloudNS recommendation


johnnagro:
re: https://www.privacytools.io/#dns

A couple issues here. First, they are an Australian company (Five Eyes)! Second, if you've tried to hit up their homepage since June, you were presented with an expired SSL certificate (https://cloudns.com.au/). Rookie.

How did they become the recommendation? Can we get that recommendation removed? There are plenty of other DNS and DNSCrypt providers out there.

PR #87, opened 2016-11-11
remove CloudNS #80

johnnagro:
re #80
re: https://www.privacytools.io/#dns

A couple issues here. First, they are an Australian company (Five Eyes)! Second, if you've tried to hit up their homepage since June, you were presented with an expired SSL certificate (https://cloudns.com.au/). Rookie.

How did they become the recommendation? Can we get that recommendation removed? There are plenty of other DNS and DNSCrypt providers out there.


Shifterovich:
You can't even visit it with Chrome due to expired SSL + HSTS.

Issue #76, opened 2016-10-04
ChatSecure Android has reached end-of-life


AntiSurveillance:
ChatSecure officially dropped support of Android platform and recommends ZomApp or Conversations. So Android should be removed (https://twitter.com/ChatSecure/status/780848326002429953)

privacytoolsIO:
Is it possible for ChatSecure iOS users to chat with Conversations Android users?

ghariib:
Quotes from their (ChatSecure) FAQ and their blog

"For a **similar** user experience and the **best compatibility** with the latest features of ChatSecure iOS, we recommend downloading Conversations." (https://chatsecure.org/faq/)

"Soon we will be implementing OMEMO Encryption in ChatSecure iOS,..." (https://chatsecure.org/blog/chatsecure-conversations-zom/)
Conversations is also using the OMEMO encryption.

Other than that they are also recommending Zom (https://zom.im/) (https://github.com/zom) (iOS and Android)

kewde:
address in #192

PR #88, opened 2016-11-11
replace ChatSecure with Conversations.im #76

johnnagro:
#76

> ChatSecure officially dropped support of Android platform and recommends ZomApp or Conversations. So Android should be removed (https://twitter.com/ChatSecure/status/780848326002429953)


pExeen:
i agree to this proposal

BlauerHunger:
I think ChatSecure should be kept for iOS, but Conversations should be recommended for Android.

kewde:
I've heard several recommendations and praises for Conversations.im so I'm fine with replacing chatsecure.

Shifterovich:
Apparently ChatSecure is in Worth Mentioning now, along with Conversations.im. Should we close @kewde?

kewde:
Hmm, I'd switch Conversations.im with ChatSecure in the worth mentioning. I see that we replaced ChatSecure with Wire.

Shifterovich:
@kewde Better make a new PR?

kewde:
Yeah

PR #89, opened 2016-11-12
Link to searx instances

IDKwhattoputhere:
I think a link to other instances is advisable. That's one of the ideas behind searx and most users looking for searx will probably not see that there's more than one instance. Searx.me isn't neccessarily the best instance for everyone either (it doesn't proxy images by default for example). On the other hand linking to only [github.io](https://asciimoo.github.io/searx/) or [github.com](https://github.com/asciimoo/searx/wiki/Searx-instances) will keep the user from immediatly being able to try searx out and might scare them off. I therefore added a second button linking to the list of instances.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1376,7 +1376,7 @@
 								<button type="button" class="btn btn-success">Website: searx.me</button>
 							</a>
 							<a href="https://github.com/asciimoo/searx/wiki/Searx-instances">
-								<button type="button" class="btn btn-success">Instances: github.com</button>
+								<button type="button" class="btn btn-success">List of Instances</button>
 							</a>
 						</p>
             </div>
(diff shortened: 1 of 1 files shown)


IDKwhattoputhere:
Sure, I just wanted to keep the formatting the same.

Shifterovich:
It's nice to keep consistent formatting, but I think it's more clear like this.

IDKwhattoputhere:
Thanks for merging. Any particular reason why you chose `stats.searx.oe5tpo.com` over the GitHub link? I'm just asking since the site is down occasionally and the list of instances on there is not updated very frequently. For example `opengo.nl` and `searx.ch` were added to the GitHub list over a month ago and `search.koehn.com` over a year ago but all three are missing from the site.

privacytoolsIO:
Agreed and fixed.

PR #91, opened 2016-11-14
Replace dead I2P-Bote link with working one

l-n-s:
i2pbote.i2p website is down and old maintainer is missing for a year, new maintainer @str4d have set up bote.i2p webpage. 

Also, i2p.us seems to be down, replace it with i2p.xyz in-proxy.


str4d:
I've just set up https://i2pbote.xyz/ which you could use instead :smile:

privacytoolsIO:
fixed.

PR #109, opened 2016-11-27
Switch Bitmessage and I2P-Bote

Shifterovich:
I2P-Bote is literally a decentralised e-mail alternative, Bitmessage is less of an alternative to email.


Shifterovich:
Created a new, non-conflicting branch, will create a new PR.

PR #111, opened 2016-11-28
Fix email provider table sorting

Shifterovich:
Price `data-value`'s are in USD. As currencies' values change, I rounded the converted values.


privacytoolsIO:
Thanks @Shifterovich

PR #112, opened 2016-11-28
Fix VPN provider table sorting

Shifterovich:
All values (apart from AzireVPN which is not rounded to a whole number to not interfere with NordVPN's $48 / year) rounded, in dollars.

### DIFF
--- a/index.html
+++ b/index.html
@@ -343,7 +343,7 @@
 								<td data-value="AzireVPN">
 									<a href="https://www.azirevpn.com/"><img src="img/provider/AzireVPN.gif" width="200" height="70"></a>
 								</td>
-								<td data-value="45">45 €</td>
+								<td data-value="47.5">45 €</td> <!-- Nord costs $48, 47.5 rounded is 48 => didn't round this one -->
 								<td><span class="label label-warning">No</span></td>
 								<td>5</td>
 								<td><span class="flag-icon flag-icon-se"></span> Sweden</td>
@@ -354,7 +354,7 @@
 								<td data-value="blackVPN">
 									<a href="https://www.blackvpn.com/"><img src="img/provider/blackVPN.gif" width="200" height="70"></a>
 								</td>
-								<td data-value="99">99 €</td>
+								<td data-value="105">99 €</td>
 								<td><span class="label label-warning">No</span></td>
 								<td>27</td>
 								<td><span class="flag-icon flag-icon-hk"></span> Hong Kong</td>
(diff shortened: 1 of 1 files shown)


IDKwhattoputhere:
Shouldn't yearly price be the price of the 1 year contract? Otherwise it's not clear that you're making a bigger commitment with the 2 year plan (referring to NordVPN).

Shifterovich:
@IDKwhattoputhere I don't fully understand what you're talking about, but I did not change prices, and yes, NordVPN's price is not correct.

IDKwhattoputhere:
NordVPN
The price currently listed for NordVPN is the price you pay per year with a 2 year subscription (see picture above). In my opinion the price listed should be the one you pay per year with a 1 year subscription.

Shifterovich:
@IDKwhattoputhere Creating a new PR fixing this would create a conflict once either of these PRs gets merged, so I should fix that, thanks for telling me.

Shifterovich:
Current normal prices:
**AirVPN**: 54 € ---------- matches
**AzireVPN**: 45 € ---------- matches
**blackVPN**: 99 € ---------- matches
**cryptostorm**: $52 ---------- matches
**doublehop**: $33 ---------- matches
**EarthVPN**: 39,99 € ---------- matches
**FrootVPN**: $35,88 ---------- **doesn't match. off by $0.12**
**hide.me**: $65.04 ---------- **doesn't match. off by $0.04**
**IVPN**: $99.96 ---------- **doesn't match. off by $0.04**
**Mullvad**: 60 € ---------- matches
**NordVPN**: $69 ---------- **doesn't match. off by $21 (wrong pricing)**
**OVPN.se**: $84 ---------- **matches, but wrong currency**
**Perfect Privacy**: $ 124.95 € ---------- **doesn't match. off by $25.05**
**proxy.sh**: $90 ---------- matches (*SOLID* plan)
**TRUST.ZONE**: $39.95 ---------- **doesn't match. off by $4.07**
**vpn.ht**: $39.99 ---------- matches
**VPNTUNNEL*: $35.88 ---------- **matches, but wrong currency**

Shifterovich:
Maybe amount of servers should be re-checked too?

IDKwhattoputhere:
Does that number actually matter? You don't know if the servers are dedicated or virtual, right? At least BlackVPN [claims](https://medium.com/@blackVPN/we-only-use-dedicated-servers-not-virtual-cloud-servers-for-maximum-security-893d6f110ebd) to have dedicated ones.

Shifterovich:
It's supposed to mean that they have a lot of servers in a lot of countries. And it doesn't matter too much, but I'd prefer the values on the website to be correct.

privacytoolsIO:
Thank you so much guys!

PR #148, opened 2017-01-04
Changed viewport to make the site mobile responsive

beardog108:
I changed the viewport to make the website responsive in mobile browsers, @Shifterovich agreed with me.


privacytoolsIO:
We've tried that a year ago. But our VPN and E-Mail comparison ended up looking weird. Did you guys double checked that? How does it look like?

Shifterovich:
@privacytoolsIO It just doesn't fit and you have to scroll to the right, but it doesn't look too weird and can't look any better afaik. The original version was much less usable than this one imho.

Issue #84, opened 2016-11-02
Restore DuckDuckGo


bakku:
Hi guys,

Recently I began searching for a search engine (pun intended). Certainly I came across DuckDuckGo and searched for information since a lot of people regard it as a search engine which respects privacy.

I came across a few problems (relevant source, sadly in german: http://www.zeit.de/digital/datenschutz/2014-01/duckduckgo-startpage-ixquick-nsa) :

- I don't know how up to date this is but DuckDuckGo uses Amazon Webservices for its service. Amazon works volunteerly together with the US government according to the source
- DuckDuckGo is based in the US. On privacytools.io it states that people should not use US services and explains why. To me this seems like a contradiction

I suggest removing DuckDuckGo from the list and maybe taking startpage.com as a candidate. I have not found information regarding startpage which shows that it is not trust worthy regarding privacy

EDIT: I would be delighted to create a PR if others agree

IDKwhattoputhere:
I'm not sure how trustworthy they are either. Specifically because they seem more secretive about certain topics instead of open like one would expect (see [here](http://securityspread.com/2016/10/24/duckduckgo-startpage-2016-update/) under "DuckDuckGo and Yahoo").

johnnagro:
+1 value point re: american company subject to american law (national security letters, etc)

justafatalerr0r:
+1. IF DDG remains, StartPage should at least be added.

["DDG Privacy,](https://duckduckgo.com/privacy) we may add an affiliate code to some eCommerce sites (e.g. Amazon & eBay) that results in small commissions being paid back to DuckDuckGo when you make purchases at those sites"

IDKwhattoputhere:
@justafatalerr0r Could you elaborate further on that quote? What's the problem with adding their affiliate tag to links in search results?

privacytoolsIO:
@bakku Good point. I have no problem with removing DDG. Should it be replaced with StartPage?

IDKwhattoputhere:
Could someone take a look at [this](https://np.reddit.com/r/privacy/comments/599ia8/duckduckgo_or_startpage_2016_update/d986wwo/) discussion? As far as I can see these requests should be blocked by most adblockers but it still made me think if SP is as trustworthy as they'd like you to believe. On the other hand they could probably hide this data collection from the user if it really were problematic. (?)

bakku:
@IDKwhattoputhere Good thread. I will have a deeper look at this and come back here when I have some results

[...]

PR #157, opened 2017-01-07
Add DuckDuckGo to worthwhile mentions

Devices:
I thought DDG would fit in the same list as other privacy focused search engines. Their privacy policy is simple enough for anyone to understand, but they are clear in the workings of the search engines.

https://duckduckgo.com/privacy

### DIFF
--- a/index.html
+++ b/index.html
@@ -1483,6 +1483,9 @@
 				<a href="https://www.ixquick.eu/">ixquick.eu</a> - Returns the top ten results from multiple search engines. Based in the
 				USA and the Netherlands.
 			</li>
+			<li>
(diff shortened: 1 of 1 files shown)


bakku:
Related to this thread https://github.com/privacytoolsIO/privacytools.io/issues/84.

I would agree to put it to the mentions section.

Devices:
I tried to make an unrelated change and it pushed it to this pull request as well and I see no way to undo that.

e5f3c2ea895af0f27667:
You should also mention their onion service: https://3g2upl4pq6kufc4m.onion

privacytoolsIO:
#84

PR #167, opened 2017-01-20
Adding "SMS Verification" section in email list (to be completed)

e5f3c2ea895af0f27667:


Shifterovich:
ProtonMail does not require SMS verification.

e5f3c2ea895af0f27667:
@Shifterovich Yes it does, 

> Too many ProtonMail accounts have been created from your connection.
> Thus, we are requesting additional verification to ensure you are human and not a spam bot.
> Because Tor is frequently abused by spammers, this check may be triggered because of the Tor exit node you are using.

Shifterovich:
The quote says that the verification isn't required *always*. Therefore saying that ProtonMail *requires* SMS verification is not completely true and should be specified further.

PR #181, opened 2017-02-19
Add Taiwan flag

ymhuang0808:
Replace the flag into Taiwan in Traditional Chinese version.


bugtender:
Agree with that. Traditional Chinese is used in Taiwan or Hong Kong.

ymhuang0808:
Thanks @Shifterovich

privacytoolsIO:
Taiwan Number 1

Issue #155, opened 2017-01-06
Remove Ubuntu Mobile


soaringowl2145:
It seems that the project will no longer be continued.  Source:

http://www.omgubuntu.co.uk/2017/01/ubuntu-phone-ota-15-ditched

bookercodes:
Although Cannonical will continue to release security patches for Oxide, etc., I do not think we should promote sunset tools.


What do you recommend we replace Ubuntu Mobile with? [Replicant](http://www.replicant.us/)?

soaringowl2145:
By sunset tools do you mean dead projects?

Shifterovich:
#196

PR #196, opened 2017-03-08
Replace Ubuntu Mobile with CopperheadOS

kewde:
Ubuntu Mobile is discontinued [0] so I swapped it out with CopperheadOS.
When looking for sources I realized Ubuntu Mobile in some way is continued but under a different name [1].
Nonetheless I think that the focus of CopperheadOS is primarily on security and privacy and that suits the website better than Ubuntu Touch.

[0] : https://en.wikipedia.org/wiki/Ubuntu_Mobile
[1] : https://en.wikipedia.org/wiki/Ubuntu_Touch

### DIFF
--- a/index.html
+++ b/index.html
@@ -2664,8 +2664,8 @@
 						<h3 class="panel-title">CopperheadOS</h3>
 					</div>
 					<div class="panel-body">
-						<p><img src="img/tools/CopperheadOS.png" alt="CopperheadOS" align="right" style="margin-left:5px;">CopperheadOS (COS) is a fork of Android (AOSP) developed by Copperhead. It is designed primarily for providing increased security and privacy, the operating system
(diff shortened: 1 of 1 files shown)


C-O-M-P-A-R-T-M-E-N-T-A-L-I-Z-A-T-I-O-N:
If conflicts are solved this should solve: https://github.com/privacytoolsIO/privacytools.io/issues/155

kewde:
@Shifterovich CopperheadOS is already added
https://privacytoolsio.github.io/privacytools.io/#mobile_os

Optionally it could be moved to the first spot, but I do want to point out that COS is open source but _not_ free anymore.

Shifterovich:
We're closing this one then. Feel free to add that in another PR.

PR #199, opened 2017-03-10
Add guidelines for testing

kewde:
@Shifterovich I added a guideline to testing PRs. Might want to merge this one as this does not require an update to the site,


mitchellcash:
I like the changes, before merging it would be nice to clean up those commits a little though (i.e. reduce it down to a single clean commit).

kewde:
Fair enough, I'm currently using the browser to deal with the maintenance.
I'll rebase it and squash the commits into one when I get on a computer with git, if anyone wants to expedite this then they can take the liberty to do it themselves as it may take a while.

PR #214, opened 2017-04-13
Reference Mastodon

ilpianista:
### Description
Mastodon is a GNU Social-compatible microblogging server

### HTML Preview

http://htmlpreview.github.io/?https://github.com/ilpianista/privacytools.io/blob/master/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -2250,6 +2250,11 @@
 			</div>
 		</div>
 
+                <h3>Worth Mentioning</h3>
+		<ul>
+			<li><a href="https://mastodon.social/about">Mastodon</a> - An open sourced, free and decentralized social network built on top of GNU Social.</li>
+		</ul>
+
 		<h3>Related Information</h3>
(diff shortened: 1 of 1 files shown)


Shifterovich:
I suggest making it the first recommendation, as it seems quite popular.

vaygr:
I second this.

Jbastardov:
There should be a line recommending Mastodon, an open-source self-managed instance-based microblogging social network. The network of federated Mastodon instances forms one part of the Fediverse, a federated network that includes servers running any social network software that uses OStatus.

Gargron:
Mastodon is [quite popular](https://mastopeek.app-dist.eu/), more so than GNU social and diaspora* combined. But more importantly, it's not built on top of GNU social in any meaningful way (it used to use the same protocol that GNU social used, but never shared any codebase and no longer uses that protocol for about a year now).

PR #223, opened 2017-05-27
Replaces Firefox Add-on Self-Destructing Cookies with Cookie AutoDelete

sts10:
### Description

It's my understanding that the currently recommended Firefox add-on for automatically deleting cookies, [Self-Destructing Cookies](https://addons.mozilla.org/en-US/firefox/addon/self-destructing-cookies/?src=ss), will not work with Firefox 57 and later. This is because "by the end of 2017, and with the release of Firefox 57, [Mozilla will] move to WebExtensions exclusively, and will stop loading any other extension types on desktop." ([source](https://blog.mozilla.org/addons/2016/11/23/add-ons-in-2017/)). The developer(s) of Self-Destructing Cookies has made it plain in [the add-on's description](https://addons.mozilla.org/en-US/firefox/addon/self-destructing-cookies/?src=ss) that he/she does not "have the time for a full rewrite as a WebExtension. Enjoy it while it lasts."  

By contrast, [Cookie AutoDelete](https://addons.mozilla.org/en-US/firefox/addon/cookie-autodelete/) is apparently prepared for this brave new world. From [a ghacks.net article](https://www.ghacks.net/2017/05/15/cookie-autodelete-for-firefox-webextension/) [posted to the privacytoolsio subbreddit](https://www.reddit.com/r/privacytoolsIO/comments/6bde8q/cookie_autodelete_for_firefox_webextension/?st=j36klypz&sh=426e5e9c): "Cookie AutoDelete is a new extension for the Firefox web browser that has been inspired by the popular Self-Destructing Cookies add-on. Cookie AutoDelete is a WebExtension, which means that it will work in Firefox 57 and newer versions of the web browser, while Self-Destructing Cookies won't."

It's functionality appears to be the same, but the UI is not identical (e.g. it looks like users can only URLs to a white list by typing them out or uploading a list of them). I could understand if there were objections here!

It looks like Firefox 57 is [slated for release in November 2017](https://wiki.mozilla.org/RapidRelease/Calendar). But I think it's best to get users, especially new users, to use Cookie AutoDelete sooner rather than later.

### You Didn't Size the Logo Image Correctly

If this is indeed the case I'm really truly sorry (here's [the contributions policy](https://github.com/privacytoolsIO/privacytools.io/blob/master/README.md#images)-- the add-on images on the site seem to be 200px by 70px which I did not do). I just grabbed [a PNG from the repo](https://github.com/mrdokenny/Cookie-AutoDelete/blob/master/src/icons/icon_128.png) and tweaked the inline CSS to make it look like the other images. I have next to zero image manipulation skills... if anyone wants to edit this PR or just redo it that'd be awesome! (Though, if I may, writing CSS to accept square PNG files for add-ons might be easier in the long run?)

### HTML Preview

http://htmlpreview.github.io/?https://github.com/sts10/privacytools.io/blob/master/index.html


mitchellcash:
The Self-Destructing Cookies add-on is still under active development, which means that it could still conform with the new WebExtension world before the cutoff date. 

Not sure if it's worth replacing based on speculation? We could maybe instead make a note against current add-on of the impending change and an alternative choice, being your proposed Cookie AutoDelete.

sts10:
Yeah, on second thought these are good points. (a) 6+ months is a good amount of time and (b) there's no guarantee that Cookie AutoDelete will be the best option at that point-- Self Destructing could certainly adapt, despite that FAQ answer in the description. 

I'll close this PR and trust that the community will keep an eye out over the summer!

mitchellcash:
@sts10 thanks for pointing out the FAQ, I hadn't read that! It does seem like the developer behind Self-Destructing Cookies is not going to make the move to WebExtensions. 

Maybe reopen this and see what some other think, you might have a good point with replacing it now.

sts10:
Thanks for the encouragement to reopen. Let's see if others chime in! 

So I guess some questions to consider are: 

1. How sure are we that Self Destructing Cookies will not be modified to work in Firefox 57 and above? The FAQ seems pretty clear, but it's not certain.
2. If indeed Self Destructing Cookies will not work in Firefox 57 (scheduled for November 2017), how much time before that should privacytools.io replace it with an alternative?
3. If, in the answer #2 we determine that replacing it sooner is best, how do we know that Cookie AutoDelete will be the best replacement? Maybe a more suitable alternative will appear in, say, October as we get closer to the release of 57. Will we regret displaying Cookie AutoDelete for a few months and then switching it out again for another add-on?
4. Does Cookie AutoDelete, in its current state, meet the requirements and expectations of the privacytools.io community? For example, from my understanding, it seems like users have to turn on "Active Mode" in order to get the same frequency of cookie destruction that is enabled by default with Self Destructing Cookies.
5. Is anyone else really craving a cookie right now?

mrdokenny:
Hi, developer of Cookie AutoDelete here. I found this PR through the magic of google.

So to answer some of your questions.

> How sure are we that Self Destructing Cookies will not be modified to work in Firefox 57 and above? The FAQ seems pretty clear, but it's not certain.

SDC's developer is most likely not going write SDC to [WebExtensions](https://gist.github.com/ssokolow/2fb8debc7dfdbf1d9a0da90284f5d73a#sdc1)

> If, in the answer Q2 we determine that replacing it sooner is best, how do we know that Cookie AutoDelete will be the best replacement? 

Out of all the extensions that are currently [compatible with Firefox 57,](https://addons.mozilla.org/en-US/firefox/search/?q=cookie&platform=all&appver=55.0&tag=firefox57&page=) Cookie AutoDelete seems to be the only one that's the closest to SDC's behavior.

>Does Cookie AutoDelete, in its current state, meet the requirements and expectations of the privacytools.io community?

I am currently in the process of closing the gap with SDC with more two features: [setting site cookies to autodelete on browser restart](https://github.com/mrdokenny/Cookie-AutoDelete/issues/20) and [localstorage support](https://github.com/mrdokenny/Cookie-AutoDelete/issues/44) 

>it seems like users have to turn on "Active Mode" in order to get the same frequency of cookie destruction that is enabled by default with Self Destructing Cookies.

"Active Mode" is turning on and off the automatic cleanup operation that happens on tab close. By default, it is off to let users add sites to their whitelist before letting the cleanup happen. In my opinion, it is better than having SDC deleting all cookies on install because people might want to save some of their cookies.

Shifterovich:
@kewde @privacytoolsIO Can you resolve the image issue?

privacytoolsIO:
done.

Issue #207, opened 2017-03-25
Cryptocurrencies recommendations are dangerous


hyc:
The Shadowcash project has been abandoned by its developers, and should be removed from the site.
https://steemit.com/cryptocurrency/@tonylondon/particl-a-new-privacy-friendly-market-platform

The Zcash listing is misleading at best, and because of Zcash' trusted setup, the project itself is not cryptographically sound. It should be removed from the site. The key point here "Unlike Bitcoin, Zcash transactions automatically hide the sender, recipient, and value of all transactions on the blockchain." is false, Zcash does *not* automatically hide anything. Users must explicitly choose to use Zcash' private transactions; the default is not private and the majority of users never change the default. Possibly because they've been misled into believing it's already private by default.

The only cryptocurrency in existence that is private automatically, by default, is Monero. It should be raised from just a "mention" to a more prominent listing. It is the only one that actually protects users' privacy.

hyc:
Listing Bitcoin is problematic as well. It is well proven to be fully traceable and offers zero privacy protection. http://www.coindesk.com/danish-police-claim-breakthrough-bitcoin-tracking/

loganmarchione:
I had a PR open, but it sat for over a month until I eventually closed it. PR's aren't automatically merged via cron job or anything, so we're at the mercy of the repo owner to recognize and approve them.
https://github.com/privacytoolsIO/privacytools.io/pull/156

Shifterovich:
Why should PRs be merged automatically?

YuFanLovezYou:
I have also wrote a post on the subreddit, addressing the misinformation on privacytools's Cryptocurrencies recommendation. 

https://www.reddit.com/r/privacytoolsIO/comments/6499qm/cryptocurrencies_recommended_on_privacytoolsio/

Especially ShadowCash/Particl, recommending ShadowCash/Particl, is dangerous, this project has raised red flags.

Josexv1:
Any news on this? i still see that the cryptocurrencies have misleading information.

ixnari:
I would just like to add that in addition to the Shadowcash devs announcing that the project is finished, it is also nigh impossible to get SDC, as most, if not all currency exchanges have delisted it. It no longer has any practical use.

3k2:
@YuFanLovezYou 

Shadowcash is also dead project so it should be removed imo..

> The ShadowCash team have stopped working on the Shadow Project.
[...]

PR #241, opened 2017-06-30
Update Zcash and Dash text

SamsungGalaxyPlayer:
I updated the Zcash text to be more accurate, and I updated the Dash text to be more descriptive.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2156,7 +2156,7 @@
 						<h3 class="panel-title">Zcash</h3>
 					</div>
 					<div class="panel-body">
-						<p><img src="img/tools/zcash.png" alt="Zcash" align="right" style="margin-left:5px;">Zcash offers total payment confidentiality, while still maintaining a decentralized network using a public blockchain. Unlike Bitcoin, Zcash transactions have the ability to hide the sender, recipient, and value of transactions in z-addresses.</p>
+						<p><img src="img/tools/zcash.png" alt="Zcash" align="right" style="margin-left:5px;">Zcash offers total payment confidentiality, while still maintaining a decentralized network using a public blockchain. Unlike Bitcoin, Zcash transactions can optionally hide the sender, recipient, and value of transactions in z-addresses.</p>
 						<p>
 							<a href="https://www.z.cash/">
 								<button type="button" class="btn btn-info">Website: z.cash</button>
--- a/index.html
+++ b/index.html
@@ -2156,7 +2156,7 @@
 						<h3 class="panel-title">Zcash</h3>
 					</div>
 					<div class="panel-body">
-						<p><img src="img/tools/zcash.png" alt="Zcash" align="right" style="margin-left:5px;">Zcash offers total payment confidentiality, while still maintaining a decentralized network using a public blockchain. Unlike Bitcoin, Zcash transactions optionally hide the sender, recipient, and value of transactions in z-addresses.</p>
+						<p><img src="img/tools/zcash.png" alt="Zcash" align="right" style="margin-left:5px;">Zcash offers total payment confidentiality, while still maintaining a decentralized network using a public blockchain. Unlike Bitcoin, Zcash transactions have the ability to hide the sender, recipient, and value of transactions in z-addresses.</p>
 						<p>
 							<a href="https://www.z.cash/">
 								<button type="button" class="btn btn-info">Website: z.cash</button>
--- a/index.html
+++ b/index.html
@@ -2184,10 +2184,6 @@
 			</div>
 		</div>
 		
-		<h3>Worth Mentioning</h3>
-		<ul>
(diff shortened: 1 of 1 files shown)


Shifterovich:
Could you provide sources of that data?

SamsungGalaxyPlayer:
[Information on Dash](https://www.dash.org/darksend/) (their website seems to be down at the moment, so you may need to fetch an [archived version](https://webcache.googleusercontent.com/search?q=cache:F70HpzJsEQIJ:https://www.dash.org/darksend/+&cd=1&hl=en&ct=clnk&gl=us)).

> PrivateSend is the feature that gives Dash users full privacy when they use it. It is an improved and extended version of the CoinJoin.

[Information on Zcash](https://z.cash/technology/index.html):

> Zcash also enables users to send public payments which work similarly to Bitcoin. With the support for both shielded and transparent addresses, users can choose to send Zcash privately or publicly. Zcash payments sent from a shielded address to a transparent address reveal the received balance, while payments from a transparent address to a shielded address protect the receiving value.

I disagree with many of the claims in these sources, but the facts as I have updated them are accurate.

afighttilldeath:
I agree with SGP that Dash and Zcash's text is inaccurate. 

Zcash does not hide transactions automatically as most transactions are transparent, in fact, I'm not aware of an exchange that supports z-addresses (shielded addresses). The shielded addresses are used approximately 10% of time (I'm being generous).

Dash is not "privacy-centric" as most of their marketing is towards fast transactions as apparent through the original text where the writer talked about "instant transactions" and 'making transactions without waits' on a privacy-centric website.  In addition, Dash's privacy is hinged on the improved and extended version of CoinJoin which relies on a small percentage of users who have "Masternodes". One thousand Dash is required to have a Masternode and the current supply of Dash is 7.4 million. It is important to note that in the first 48 hours of Dash launching approx. two million Dash was mined so there is a good possibility a small percentage of people own Masternodes.

My recommendation for Dash:
Dash is an open-source, Bitcoin-based cryptocurrency that offers an improved version of CoinJoin called PrivateSend through the use of Masternodes.

Zcash Source:
https://z.cash/support/faq.html (Under "Are only a small fraction of Zcash users using shielded addresses?"
https://explorer.zcha.in/statistics/usage (this is referenced on FAQ in regards to transaction statistics)
https://www.reddit.com/r/zec/comments/5hyyc1/do_any_exchanges_give_shielded_addresses_for/

Dash Source:
https://www.dash.org/wp-content/uploads/2015/04/Dash-WhitepaperV1.pdf
https://dashpay.atlassian.net/wiki/display/OC/Dash+Instamine+Issue+Clarification

Shifterovich:
> Dash is an open-source, Bitcoin-based cryptocurrency that offers an improved version of CoinJoin called PrivateSend through the use of Masternodes.

Explain what those buzzwords mean, though.

hyc:
I think the main point is as I outlined it in issue #207 none of these others are private by default.

But even calling Dash "privacy-centric" is inaccurate as it's clear that the CoinJoin protocol it uses has zero privacy guarantees.

Shifterovich:
We can easily remove Dash. It's just in the Worth Mentioning section.

afighttilldeath:
@Shifterovich Sorry, I forgot this isn't a cryptocurrency website. lol, it's hard to explain without going too into it. I'd recommend removing Dash.

Shifterovich:
I meant

> CoinJoin-based mixing system called PrivateSend

instead of

> improved version of CoinJoin called PrivateSend through the use of Masternodes

Anyway, I'm no cryptocurrency expert, but I haven't heard good things about Dash either.

SamsungGalaxyPlayer:
@Shifterovich @afighttilldeath I also recommend its removal, though I imagine that should go in a different pull request. This request only intends to make the current wording more accurate.

Shifterovich:
Done :P

Shifterovich:
> Unlike Bitcoin, Zcash transactions optionally hide the sender

Doesn't sound very good. Maybe just point out that it *has the ability* to hide the sender.

afighttilldeath:
:+1:

hyc:
Not to nitpick, but I kinda prefer "may optionally" to "have the ability". Feel free to disregard this comment if everyone else is happy with the latter.

Shifterovich:
OP changed one word and it doesn't fit very well into the context. I'd at least change "optionally" to "can optionally" or like you suggested "may optionally". And indeed, "have the ability" doesn't fit in very well either given the context. I wasn't being literal when I said "point out that it *has the ability* to hide the sender."

daira:
"Unlike Bitcoin, Zcash transactions can optionally hide the sender, recipient, and value of transactions in z-addresses."

This is not accurate. Hiding of sender, recipient, and value of transactions in z-addresses is not optional. Use of shelled transactions (between z-addresses) is optional. I suggest the following correction:

"Unlike Bitcoin, Zcash supports shielded transactions, which hide the sender, recipient, and value."

Issue #259, opened 2017-07-09
Replace keepassx with keepassxc


pExeen:
Hello I think that keepassxc should be mentioned at the end of the keepass description because keepassx hasn't been updated since 8 months.

KhalidVIP:
But you have to wait because the program will have some Bugs

PR #261, opened 2017-07-10
Add KeePassXC (KeePassX's community fork)

hugoncosta:
Title, I've also added a little note saying that KeePassX doesn't have regular updates, feel free to use a shorter term to say that, I couldn't remember it. Brought up by #259.

Sidenote: The box is too small for all the text, should we just remove the update comment?

### HTML Preview

http://htmlpreview.github.io/?https://github.com/hugoncosta/privacytools.io/blob/patch-3/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -1896,7 +1896,7 @@
 					</div>
 					<div class="panel-body">
 						<p><img src="img/tools/KeePass.png" alt="KeePass" align="right" style="margin-left:5px;">KeePass is a free open source password manager, which helps you to manage your passwords in a secure way. All passwords in one database, which is locked with one
-							master key or a key file. The databases are encrypted using the best and most secure encryption algorithms currently known: AES and Twofish. See also: <a href="http://www.keepassx.org/">KeePassX</a>.</p>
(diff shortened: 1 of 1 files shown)


kewde:
Hmm,
The community edition does seem to have a more active development, I don't know what @Shifterovich thinks but replacing it completely might be more interesting route.

Shifterovich:
Wanted to suggest that, but that might require some additional feedback. I know that some people prefer KeePassX over KeePass2 because they don't like mono. But I don't know anything about the community edition.

hugoncosta:
Maybe put the replace the See also with "KeePassXC and KeePassX (older)"?

kewde:
I'm fine with approving this as a temporary measure but how many forks of KeePass are there?

KeePassX?
KeePass2?
KeePassXC?

Shifterovich:
KeePass2 isn't a fork iirc.

kewde:
I suppose we could poll on reddit about this one?
I like the active development of KeePassXC but I hadn't heard of it before and don't know any of the devs..

Shifterovich:
> I like the active development of KeePassXC but I hadn't heard of it before and don't know any of the devs..

Exactly. Non-maintained software is better than ill-maintained software.

PR #265, opened 2017-07-13
Update Zcash description

alvinjoelsantos:
### Description

Updated Zcash description to include a caution to highlight the fact that the privacy feature of Zcash is not automatic and requires users to generate a z-addr. Although some service providers like Shapeshift [support z-addr](https://twitter.com/ShapeShift_io/status/857639222487994368), the majority of pools, light wallets, and exchanges only use t-addr, so new users should be warned of the difference.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2246,7 +2246,7 @@ switch(document.readyState){case'complete':case'loaded':case'interactive':onDomC
 						<h3 class="panel-title">Zcash</h3>
 					</div>
 					<div class="panel-body">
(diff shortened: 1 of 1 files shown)


kewde:
Hmm, this a tough one..

The information is accurate, I would however reword it to a much simpler sentence:
"Use z-addresses for anonymity."

I think it's _generally_ a bad idea to go down this path of giving what I call "opsec" advice.
There are a thousand ways to screw up using any privacy application..


Should we add a warning "do not log into facebook and other personal websites" to Tor Browser?
Should we add a warning "Stores your phone number on their server" to Signal et al?
Should we add a warning "Requires you to wait X amount of days before spending coins again" to Monero (timing analysis attacks)?

Shifterovich:
Indeed, nearly finished with the theory (like explaining crypto terms), so I'll start writing step-by-step tutorials for specific tools soon.

privacytoolsIO:
#269

Issue #256, opened 2017-07-08
Switch Bitcoin and Monero


Shifterovich:
See the #247 comment thread

@hyc:
> Monero is still the only coin that is private by default - it's the one that should be recommended first.

kewde:
If any coins deserves to be first, it's either Bitcoin (usability) or Zcash (technical superiority). I have been fascinated by ring signatures and RingCT (albeit for the wrong reasons, the small anonymity subset per transactions makes for interesting attacks) and I'll honestly admit: it's theoretically inferior to the anonymity provided by zk-SNARKs. I'm by no means saying RingCT is bad or doesn't work, but on an individual basis, a transaction by Monero is less anonymous than a Zcash transaction.

We already had a lengthy discussion about this in 207. I know we'll see a small flood of people supporting Monero making baseless claims, but the opinion of the biased masses don't automagically make it the truth.

hyc:
Zcash is not private by default. It is this way for solid technical reasons - constructing a shielded txn takes several GB of RAM and multiple minutes of CPU time. Quit pushing for the "theoretically superior" solution when it's known not to actually work in the real world.

lethos3:
> Monero (although the zcash proponents note that a ring signature is a "smaller" anonymity set, they usually don't mention that the stealth address factor actually means that each transaction is masked, whereas the ring signatures provide additional plausible deniability, furthermore, since keys appear in different ring signatures in different blocks in time, the anonymity set for when a given key is spent grows infinitely, and could eventually grow larger than the zcash anonymity set at any fixed instant in time) vs Zcash (anonymity set is the entire blockchain )

- Shen Noether

DiMiTri101:
Monero first. Zcrash is shit.

BitOfWisdon:
How ZcRash understands 'privacy' from Zooko's own incompetent tweets:
https://twitter.com/zooko/status/863202798883577856 
...and their other tweets: https://twitter.com/zcashgpumining/status/863547851565084672 
On top of that ZcRash is a U.S. COMPANY, sponsored and developed by the Israeli government link and in an "enterprise partnership" with JP Morgan link...

xmr-eric:
@BitOfWisdon Your "link"s are not linked ;)

c789:
Monero should be listed first. It is private by default. The other coins are not private by default.

Monero transactions are almost instantaneously constructed, vs. an impractical amount of time required to construct a private Zcash transaction. 

[...]

PR #247, opened 2017-07-04
Move bitcoin

kewde:
bitcoin might not be the most private but you'll need it to buy other coins, makes sense to put it first.

### HTML Preview

https://htmlpreview.github.io/?https://github.com/kewde/privacytools.io/blob/bitcoin/index.html#cryptocurrency

PR #239, opened 2017-06-29
Replaced Shadowcash with Monero

afighttilldeath:
### Description
First time creating a pull request (or doing anything on GitHub) so please bear with me. 

In response to issue #207, I have deleted Shadowcash from the cryptocurrency privacy listing and added Monero instead. I believe Monero has a proven track record for protecting people's privacy and would be a great addition to privacytools.io. 

edit: changed HTML Preview to correctly redirect you.
### HTML Preview

https://htmlpreview.github.io/?https://github.com/afighttilldeath/privacytools.io/blob/patch-1/index.html

PR #267, opened 2017-07-14
Switch Bitcoin and Monero

Shifterovich:
### Description

#247 #256 

### HTML Preview

http://htmlpreview.github.io/?https://github.com/privacytoolsIO/privacytools.io/blob/Shifterovich-switch-btc-xmr/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -2227,16 +2227,16 @@ switch(document.readyState){case'complete':case'loaded':case'interactive':onDomC
 			<div class="col-sm-4">
 				<div class="panel panel-success">
 					<div class="panel-heading">
-						<h3 class="panel-title">Bitcoin</h3>
-					</div>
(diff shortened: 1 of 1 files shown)


Shifterovich:
Reverting an unapproved change.

kewde:
The initial PR #239 wasn't approved either, what I reverted with move Bitcoin was an addition that I did not agree with.

Shifterovich:
> kewde reviewed 15 days ago 
> :+1:

kewde:
I did not agree with the switch of Bitcoin, that went under my radar, hence I reverted it.

Shifterovich:
Except you're not the one who decides which CCs go on the website only because you develop Particl.

kewde:
I don't decide indeed, I'm trying to achieve consensus and there is an active debate going on.

PR #301, opened 2017-07-31
Telegram insecurity

Nalapl3:
### Description
https://www.reddit.com/r/privacy/comments/6qmzkx/best_desktop_app_for_instant_messaging_with/dkyk2an/
https://security.stackexchange.com/a/49802

Link two posts from Reddit and Stackoverflow. The posts are linking to various sources, so it's just easier to add the posts instead of linking all of the sources, the posts are also shorter than the original sources, making it easier to read for average person.
I think it's pretty important to describe why Telegram isn't secure, cause their marketing promoting it as such might be misleading for many people reading the website.

### HTML Preview

http://htmlpreview.github.io/?https://github.com/nalapl3/privacytools.io/blob/patch-1/index.html


kewde:
The information and the complaints date back from 2015 - so I was not sure if it was still relevant, I hoped they may have fixed the issues by now. I spend a good hour last week going through the research and checking if it's still relevant and some issues are definitely not fixed.

https://core.telegram.org/mtproto

1. MTProto is _STILL NOT_ IND-CCA secure. The encryption scheme still does not authenticate the padding (length nor content).
2. Uses the broken SHA-1 hash function

PR #304, opened 2017-08-02
Remove bad email providers

Shifterovich:
### Description

CryptoHeaven doesn't use HTTPS, CounterMail requires Java to even register.

### HTML Preview

http://htmlpreview.github.io/?https://github.com/privacytoolsIO/privacytools.io/blob/Shifterovich-email-patch/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -1190,22 +1190,6 @@
 						</tr>
 
 						<tr>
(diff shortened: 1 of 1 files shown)


privacytoolsIO:
good catch, mate.

PR #313, opened 2017-08-09
Updated servers for OVPN and country for VPNTunnel

captaindoe:
### Description
OVPN launched six new servers this week. Three in Los Angeles and three in Miami. Confirmation address: https://www.ovpn.com/en/status

Also updated country for VPNTunnel.com since according to VPNTunnel.com's footer, the company is Edelino Commerce Inc which is located in Seychelles on the following address:

Suite 1, Second Floor, Sound & Vision House, 
Francis Rachel Street, Victoria, Mahe, Seychelles 

### HTML Preview

http://htmlpreview.github.io/?https://github.com/ovpn-se/privacytools.io/blob/master/index.html


privacytoolsIO:
thx

PR #315, opened 2017-08-10
Group and sort Firefox about:config tweaks

sbennett1990:
### Description

This makes it easier to apply the settings to Firefox, since you shouldn't have
to type in the full setting each time. I have applied these to probably a dozen
Firefox instances, and each time I have wished that the setting names were
sorted and grouped together.

I haven't removed any of the settings nor changed any wording.

I left the privacy settings at the top of the list, since they are the most
relevant.

### HTML Preview

http://htmlpreview.github.io/?https://github.com/sbennett1990/privacytools.io/blob/master/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -940,29 +940,24 @@
 		<h3>Getting started:</h3>
 
 		<ol>
-			<li>privacy.trackingprotection.enabled = true</li>
-			<ul>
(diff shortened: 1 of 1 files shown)


privacytoolsIO:
Well done, mate.

PR #325, opened 2017-08-25
Removed Muonium (duplicate entry)

p2635:
As far as I know Muonium is just encrypted cloud storage but not a piece of self hosted software. I noticed this entry in 2 places so I've removed it only from the Self-Hosted Cloud Server Software section as it doesn't apply (already appears in Encrypted Cloud Storage Services)

If I'm incorrect about this, please ignore me.

### HTML Preview
http://htmlpreview.github.io/?https://github.com/p2635/privacytools.io/blob/master/index.html


p2635:
Html preview doesn't seem to work, did I do something wrong?

p2635:
I made a second change, I went through the links and replaced http with https where possible. I've never done a pull request before, it looks like everything is going into one thread now! (not sure how to separate)

p2635:
Made a third change (how should I rename 'Removed Muonium (duplicate entry)' as the title only relates to 1 change now...)

Shifterovich:
> As far as I know Muonium is just encrypted cloud storage but not a piece of self hosted software.

https://github.com/muonium/core

> We haven't really packaged Muonium, but you can modify the code to run it on your own server.

PR #334, opened 2017-09-08
Switch Red/Green in KDL section

### LINKED ISSUE #333: Red and Green box position switch in Key disclosure law
In "Key disclosure law" the green box is to the right instead of the left. In all other cases the green (best option) is the left.

Solution: red and green box switch position.

shifterovich2:
### Description

**Don't merge until the design is resolved.**

Resolves #333 

Not sure whether I like it more now. Maybe use the same layout as #ukusa?

### HTML Preview

http://htmlpreview.github.io/?https://github.com/shifterovich2/privacytools.io/blob/switch-red-green-kdl/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -2028,7 +2028,8 @@
 						1. <a href="https://en.wikipedia.org/wiki/Key_disclosure_law#New_Zealand">New Zealand</a> (unclear) <div class="pull-right"><span class="flag-icon flag-icon-nz"></span></div>
 						<br /> 2. <a href="https://en.wikipedia.org/wiki/Key_disclosure_law#Sweden">Sweden</a> (proposed) <div class="pull-right"><span class="flag-icon flag-icon-se"></span></div>
 						<br /> 3. <a href="https://en.wikipedia.org/wiki/Key_disclosure_law#The_Netherlands">The Netherlands (*)</a> <div class="pull-right"><span class="flag-icon flag-icon-nl"></span></div>
-						<br /> 4. <a href="https://en.wikipedia.org/wiki/Key_disclosure_law#United_States">United States</a> (see related information) <div class="pull-right"><span class="flag-icon flag-icon-us"></span></div>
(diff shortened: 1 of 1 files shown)


kewde:
@Shifterovich 
Is @shifterovich2 an account that you own?

kewde:
Special key disclosure laws exist for Belgian citizens. I've moved it to the right section.

Shifterovich:
Yes, it's the account I use on public/school computers.

shifterovich2:
@kewde What do you think about the layout? Use the same as #ukusa or keep it like this?

Shifterovich:
@hugoncosta Your opinion on the layout?

hugoncosta:
The preview says it cannot load, could you give me another link?

Shifterovich:
Right, I apparently deleted the branch. I'll update the link once I have time to upload the files to my website.

hugoncosta:
btw I was thinking, I don't believe we should change the layout. We use this on the number eyes section too, and we have the worst offenders first. I believe that the current layout should stay, since both KDL and #Eyes are warnings and not necessarily tools/advices. Our philosophy of putting the "best" option with a green box and to the left should remain, by putting the "worst" offenders (in this case) to the left and in a red box.

Shifterovich:
@kewde?

kewde:
@Shifterovich the link is dead so I can't see it anymore. Can you snap a pic of it?

Shifterovich:
https://htmlpreview.github.io/?https://raw.githubusercontent.com/privacytoolsIO/privacytools.io/89ce79e4637bc8349f091bbc5e538feafbb1384e/index.html#kdl

hugoncosta:
I'm still not conviced. I believe that the panel on the left should be the most important - either the "stay away" or "you should really use this". If we change the KDL from "Good" to "Bad", we should also change the Global Mass Surveillance and put the Fourteen Eyes first (switch like in here).

Shifterovich:
Agree, I think I liked it better before aesthetically as well. Just wanted to know what you think. @kewde @beardog108 what do you think? Merge or close?

kewde:
I don't really care about this change, it's just moving boxes :package: and the order of worst -> best is clear here.

PR #340, opened 2017-09-27
add Firefox about_config values for Referer

groovecoder:
### Description
Add suggested values for `network.http.referer.XOriginPolicy` and `network.http.referer.XOriginTrimmingPolicy`

### HTML Preview

http://htmlpreview.github.io/?https://github.com/groovecoder/privacytools.io/blob/about-config-referer/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -1031,6 +1031,30 @@
 				<li>3 = Accept for N days</li>
 			</ul>
 
+      <li>network.http.referer.trimmingPolicy = 2</li>
+      <ul>
+        <li>Send only the scheme, host, and port in the <code>Referer</code> header</li>
+        <li>0 = Send the full URL in the <code>Referer</code> header</li>
+        <li>1 = Send the URL without its query string in the <code>Referer</code> header</li>
+        <li>2 = Send only the scheme, host, and port in the <code>Referer</code> header</li>
+      </ul>
+
+      <li>network.http.referer.XOriginPolicy = 2</li>
+      <ul>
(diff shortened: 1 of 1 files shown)


kewde:
I wasn't aware about these configuration settings, thanks!

kewde:
Is there a particular reason why you set `network.http.referer.XOriginPolicy` to 1 instead of 2?

1: a.example.com and b.example.com would be allowed
2: only b.example.com & b.example.com would be allowed

Seems like the second option is even more secure, but maybe at the expense of functionality.

groovecoder:
Right; `1` would seem to break fewer things. When we studied its effect on user-reported breakage, it seemed to be minimal. We didn't study `2` yet, but we probably will. I'll update here again if `2` seems just as good.

groovecoder:
Sorry I had my settings crossed. We didn't actually study `XOriginPolicy` effects on breakage. But the assumption is the same - referers to eTLD will presumably break fewer sites.

Shifterovich:
Not sure whether `1` really helps. Maybe recommend `2` and explain that if it breaks sites a lot, the user should try changing it to `1`?

groovecoder:
Depends on the adversary and the risk. eTLD's would able to combine/track across the sub-domains if they wanted to. But I see the other privacy settings tweaks maximize privacy, so `2` makes sense here.

Updated commit to suggest `2` and also to suggest a global `Referer` trimming policy.

PR #344, opened 2017-10-05
Add a warning about GApps

hugoncosta:
### Description
Small alert in the Mobile OS area with concern to the use of apps made by Google.

### HTML Preview

http://htmlpreview.github.io/?https://github.com/hugoncosta/privacytools.io/blob/patch-1/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -2685,7 +2685,7 @@
 		</div>
 		<!-- alert message about GApps -->
 		<div class="alert alert-warning" role="alert">
-			<strong>Even though the Source Code of the following OS are provided, installing Google Apps may compromise your setup</strong>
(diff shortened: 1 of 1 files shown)


Shifterovich:
Maybe add a similar notice to Lineage too? IIRC it can be downloaded both with and without Google services.

hugoncosta:
Changed everything. I was in doubt with the "are" too, since I wanted to make reference to the 3 alternatives, but in hindsight, it does sound better with "is".

The GApps warning is supposed to be to Android OS in general, since you have the option to install them on every distro.

Shifterovich:
Been nearly a week, still no approval of another contributor. Merging.

kewde:
Please ping me with @kewde in cases like these (or on IRC), I have tons of messages on github

PR #350, opened 2017-10-21
Update pricing of IVPN

Heliosmaster:
### Description
Update pricing of IVPN according to: https://www.ivpn.net/pricing (semi-permanent sales is considered, as it was already the case for other VPNs in this list)

### HTML Preview

http://htmlpreview.github.io/?https://github.com/Heliosmaster/privacytools.io/blob/patch-1/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -358,7 +358,7 @@
 							<tr>
 								<td data-value="AirVPN">
 									<a href="https://airvpn.org/"><img src="img/provider/AirVPN.gif" width="200" height="70"></a></td>
-								<td data-value="54">54 €</td>
+								<td data-value="63">54 €</td>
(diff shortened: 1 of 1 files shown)


Shifterovich:
Note that the value of `data-value` is in USD. Should be 53, then. `60` makes it more expensive than AirVPN, which is false.

We should update all `data-value` values for prices in EUR, anyway.

Heliosmaster:
I'm not sure what you mean, IVPN is 60$, not €.

I updated the data-values to be in $ :)

Shifterovich:
Ah, sorry, for some reason I thought it said `€`.

The values seem right, thanks. Merging.

Issue #339, opened 2017-09-26
Firefox SafeBrowsing should be okay to enable


groovecoder:
Is there a link or an article to the privacy concerns of enabling SafeBrowsing? It [should not be a privacy concern](https://feeding.cloud.geek.nz/posts/how-safe-browsing-works-in-firefox/), and the security risk seems very high.

Atavic:
Is useful if you're not using anything else. Moreover it is a service based on Amazon servers (3rd party). See discussion [here](https://github.com/ghacksuserjs/ghacks-user.js/issues/102).

groovecoder:
Only the Tracking Protection service ([shavar](https://github.com/mozilla-services/shavar)) is hosted on Amazon servers. (I'm the maintainer of that code-base, in fact.)

SafeBrowsing lists are served by Google's servers directly to Firefox clients. It should not be a privacy concern (as explained by the article I linked ... which was written by the Firefox SafeBrowsing maintainer).

gitbugged:
> SafeBrowsing lists are **served by Google's servers directly** to Firefox clients. It should not be a privacy concern...

And that's [precisely](https://stallman.org/google.html) why it is.

groovecoder:
I'm not a big fan of big G either, but given that ...

>Query string parameters are stripped from URLs we check as part of the download protection feature.
>Cookies set by the Safe Browsing servers to protect the service from abuse are stored in a separate cookie jar so that they are not mixed with regular browsing/session cookies.
>When requesting complete hashes for a 32-bit prefix, Firefox throws in a number of extra "noise" entries to obfuscate the original URL further.

... the only privacy threat I can see with Firefox clients contacting Google's servers is that it leaks an IP address to Google's servers. But this only tells Google that there's a Firefox browser at the IP address and nothing more.

Compare that to the security (and corollary privacy) risks of completely disabling malware, unwanted software, and social engineering threat checks in Firefox, and it seems that disabling SafeBrowsing is based on Google-paranoia - not a realistic privacy concern.

Shifterovich:
> Compare that to the security (and corollary privacy) risks of completely disabling malware, unwanted software, and social engineering threat checks in Firefox

This can be solved with common sense which doesn't leak anything to Google.

groovecoder:
[...]

PR #353, opened 2017-10-29
for #339: Notes about uBlock and SafeBrowsing list updates

groovecoder:
### HTML Preview

*Replace [GITHUB_USERNAME] with your GitHub username and [BRANCH] with the branch name.*

http://htmlpreview.github.io/?https://github.com/[GITHUB_USERNAME]/privacytools.io/blob/[BRANCH]/index.html


zero77:
@groovecoder 
I think that warning should also be added to some of the other addons, like HTTPS Everywhere , Privacy Badger and uMatrix.

Because, they also update from remotely hosted block lists.

groovecoder:
Sorry I've lost track of these changes, but someone else can re-submit them if they want.

PR #361, opened 2017-11-14
add ipfs to the self contained networks section

emanresusername:
### Description

add ipfs to the self contained networks section

### HTML Preview

http://htmlpreview.github.io/?https://github.com/emanresusername/privacytools.io/blob/master/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -2260,7 +2260,7 @@
 			<li><a href="https://zeronet.io/">ZeroNet</a> - Open, free and uncensorable websites, using Bitcoin cryptography and BitTorrent network.</li>
 			<li><a href="http://retroshare.net/">RetroShare</a> - Open Source cross-platform, Friend-2-Friend and secure decentralised communication platform.</li>
 			<li><a href="https://gnunet.org/">GNUnet</a> - GNUnet provides a strong foundation of free software for a global, distributed network that provides security and privacy.</li>
-			<li><a href="https://ipfs.io/">IPFS</a> - A peer-to-peer hypermedia protocol to make the web faster, safer, and more open. (<a href="https://github.com/privacytoolsIO/privacytools.io/pull/361">warning/discussion</a>)</li>
+			<li><a href="https://ipfs.io/">IPFS</a> - A peer-to-peer hypermedia protocol to make the web faster, safer, and more open. (<a href="https://github.com/privacytoolsIO/privacytools.io/pull/361#issuecomment-344414022">important privacy warning</a>)</li>
 		</ul>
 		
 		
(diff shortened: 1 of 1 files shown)


Shifterovich:
@kewde

beardog108:
@IPFS, while being open source, and utilizing encryption for traffic, was not designed with anonymity or privacy in mind (unlike Freenet, which is kind of similar in that they're both data store programs).

You can kind of form a darknet with IPFS if you set IPFS to only bootstrap with friends, (and your friends do the same) but this is not as good as Retroshare or Freenet when it comes to anonymity & privacy. **This requires some technical knowledge so we shouldn't expect normal users to do this** This can also be done with traditional tormenting to an extent, since private trackers or a self-contained DHT and enabled encryption would essentially do this. 

Like Bittorrent, you can see who is seeding/sharing any given file on the public IPFS network, although VPNs can help with this to an extent. I don't believe IPFS supports Tor very well, but I could be wrong. I know OpenBazaar ended up creating an [addon](https://github.com/OpenBazaar/go-onion-transport) for onion support, but this was for OpenBazaar only.

In addition to traditional torrent-like concerns, IPFS also includes a web gateway to access files from your browser. This is enabled by default, but I believe it can be disabled. Using an "attack" (not really an attack so much as it is an abuse of features) I came up with early this year **websites can create supercookies which persist even if your browser is wiped or a different browser is used**. [Link to this attack, here.](https://www.chaoswebs.net/blog/tracking-ipfs-users-via-cache-probing.html)

I realize not everyone's threat model includes complete anonymity, so I guess it would be fine to add IPFS (as you are) to a worth mentioning, but I think we should put a warning.


To summarize:
* IPFS is not much better than open source Bittorrent clients
* IPFS was not really designed with privacy in mind (although it does use encryption for traffic)
* Some features can be abused to actually harm user privacy, even when they're not actively using IPFS.

kewde:
I'm checking this out.

IPFS is indeed not made for anonymity but I have seen moves towards Tor support.

Some interesting GitHub issues & repos that are about IPFS & Tor.
https://github.com/ipfs/notes/issues/37
https://github.com/OpenBazaar/go-onion-transport

emanresusername:
:astonished: whoa! y'all are **way** more knowledgable here than i, i defer
[relevant thread](https://github.com/ipfs/faq/issues/18) before i disappear 
disappear

kewde:
@beardog108 

IPFS makes use of an node keypair (I don't know the exact terminology, but it will do) and it persist across reboots. This key is used in the protocol to identify itself & maintain a reputation with other nodes through an internal ledger. 

A silly implementation of IPFS and Tor together, would still result in a persistent node keypair, essentially serving as a fingerprint. 
I wonder if the current Tor implementation of IPFS makes use of ephemeral (temporary) keys in those cases.

Shifterovich:
Not private by default, though. Are we closing @kewde @beardog108?

Shifterovich:
What about Worth Mentioning with a warning @kewde?

kewde:
@Shifterovich 

A worth mentioning with a warning seems more appropriate.

Shifterovich:
@emanresusername

emanresusername:
how's that last commit for the warning? (just linked to the convo here)

beardog108:
I would say something along the lines of "important warning regarding privacy" or just "important warning" and specifically link to #issuecomment-344414022

emanresusername:
@beardog108 hows [that](https://github.com/privacytoolsIO/privacytools.io/pull/361/commits/016fc0167f7e11ad2e198c0e5aa407002d0b4def)?

beardog108:
Yeah looks good to me, thanks.

Shifterovich:
@kewde

PR #367, opened 2017-11-26
List open source means to prevent WebRTC leaks in Chrome.

aghorler:
### Description

Although it's not possible to disable WebRTC in Chrome, an API exists to change its routing settings.

The API can be used to prevent leaks using extensions.

*index.html* currently mentions that only a single extension exists, and that it can be easily circumvented. 

This is no longer true. 

The old extension (which has been removed by its developer) attempted to block WebRTC by injecting JavaScript into web pages (a content script). This solution is flawed because it's not possible to inject code into a sandboxed iframe.

Current extensions use an API provided by Chrome, and cannot be circumvented.

### HTML Preview

https://htmlpreview.github.io/?https://github.com/aghorler/privacytools.io/blob/master/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -832,7 +832,7 @@
 
 		<!-- steps to manually disable WebRTC in Chrome -->
 		<h3>How to fix the WebRTC Leak in Google Chrome?</h3>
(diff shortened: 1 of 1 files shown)


Shifterovich:
Thanks

Issue #384, opened 2018-01-09
DNSCrypt Proxy v1 replaced by v2


NPN:
DNSCrypt has been abandoned by its creator, Frank Denis. On November 10, 2017, Denis [tweeted](https://twitter.com/jedisct1/status/928942292202860544) that the project was looking for new maintainers. Some time later, the [project repository](https://github.com/dyne/dnscrypt-proxy) was moved to [Dyne](https://github.com/dyne), the new maintainers (edit: here is a [tweet](https://twitter.com/DyneOrg/status/949957701114761216) from them).

The [README](https://github.com/dyne/dnscrypt-proxy/blob/master/README.markdown) on Dyne's repository explains that DNSCrypt will continue to be maintained until "a viable and mature alternative arises," because Dyne relies on it for their project [Dowse.eu](https://dowse.eu/). However, no new features will be added. Similarly, the [first commit](https://github.com/dyne/dnscrypt-proxy/commit/45912417738ce38ce6dd04cc4173d276a17ef2d8) after the transfer states that:

> We will likely work to keep dnscrypt around for another year at least, while evaluating if to shift our code to use other alternatives.

The domain dnscrypt.org is still registered under Denis, but now redirects to https://dnsprivacy.org/wiki/, the site for the "DNS Privacy Project." As the Dyne README explains:

> The old webpage dnscrypt.org now points to a new domain, endorsing the usage of competing protocol "DNS-over-TLS" and competing software in particular the "getdns" library and an immature implementation that could substitute dnscrypt-proxy, called "stubby".
> The new website also links a [critical analysis of DNSCrypt vs DNS-over-TLS protocols](https://tenta.com/blog/post/2017/12/dns-over-tls-vs-dnscrypt) by a company marketing their own open-source Android web browser and offering a new DNS resolver implemented in Go.

Denis has created a [new repository](https://github.com/jedisct1/dnscrypt-proxy) under the same name as before, and is writing "a new implementation [of DNSCrypt] that sucks less."

I am uncertain of the status of the DNSCrypt resolvers. DNSCrypt is still working for me right now, so they probably haven't all shut down. The [resolver list page](https://dnscrypt.org/dnscrypt-resolvers.html) is no longer available because the website is gone, but the [CSV resolver list](https://github.com/dyne/dnscrypt-proxy/blob/master/dnscrypt-resolvers.csv) is still in Dyne's repository.

--------

[...]

PR #388, opened 2018-01-13
Remove DNSCrypt

NPN:
### Description

Remove DNSCrypt. Explanation is in #384.

### HTML Preview

http://htmlpreview.github.io/?https://github.com/NPN/privacytools.io/blob/remove-dnscrypt/index.html


zero77:
The problem with DNS over TLS is that it leaks the hostname in plain text by the Server Name Indication (SNI) extension for TLS.

kewde:
#393

Issue #399, opened 2018-02-08
Suggestion: mention XPrivacyLua


jj-pietro:
XPrivacyLua is an Android open source xposed module which blocks access to personal data by feeding fake data to apps instead of revoking permissons, I think it should be added to the "worth mentioning" section of "Mobile Operating Systems".
Github: https://github.com/M66B/XPrivacyLua
Website: https://lua.xprivacy.eu/
Support: https://forum.xda-developers.com/xposed/modules/xprivacylua6-0-android-privacy-manager-t3730663
It requires the [Xposed Framework](https://forum.xda-developers.com/showthread.php?t=3034811) which is open source too.

hugoncosta:
This is in line with the #338. The plugin seems great though, I can't understand how it's done, but I am a fan of security by obfuscation. 

@kewde, @Shifterovich, what do you think if we create a subsection under Mobile OS with useful plugins/add-ons, just like with Firefox (https://www.privacytools.io/#addons)?

Shifterovich:
Sounds great, but someone has to do it. I'm very busy lately.

hugoncosta:
Awesome, I'll start hacking at it, I'll see which other add-ons there are that we've been speaking about in the other issues.

pietropacchio:
@hugoncosta while you are at it you may want to add [AFWall+](https://github.com/ukanth/afwall) and/or [Netguard](https://github.com/M66B/NetGuard) as a no-root alternative.
There's also an [hosts file](https://github.com/jawz101/MobileAdTrackers/) created specifically to block some mobile tracking libraries.

hugoncosta:
@pietropacchio I'm currently a user of AFWall, would you say that the big difference is that Netguard requires no root? 

I think I'll do two trees - one with and one without root access.

pietropacchio:
@hugoncosta well with NetGuard you can load a custom hosts file instead of blocking apps, a thing which you can't do using AFWall+ but for doing its things it needs to use the VPN service, so due to Android limitations you can't use another one.

CHEF-KOCH:
The problem here is that XLUA needs root, or in other words, the Xposed framework installed, this triggers Google Play Store and it's SafetyNET check. Currently, there is no workaround to bypass this, even if it takes only some small changes in order to re-detect this. 

NetGuard should be preferred from all mentioned solutions cause it simply works with Android's native VPN API which doesn't need you to root your device and works for everyone without triggering security mechanism by Google.

[...]

PR #400, opened 2018-02-16
Add Android Add-ons

hugoncosta:
### Description
Creation of a new section focused on Privacy Add-ons for the Android OS. 
There's a problem with the CSS and one of the images overlaps the text, any idea on how to fix it?

### HTML Preview

http://htmlpreview.github.io/?https://github.com/hugoncosta/privacytools.io/blob/master/index.html


Shifterovich:
You added the pictures to `/`. Move them to `img/addons/`.

kewde:
Image sizes are OK :+1:, would use pngs rather than gifs but not a blocker imo.

I really like this PR.
But before we merge this, what is the exact purpose of this section?
Is this section for _any_ android application that improves the security and privacy of the _user_.
Or is it limited to applications that make the _android system_ more secure and private.

hugoncosta:
I thought the same, but the imgs used on the Firefox Add-ons (where I based the category) were all gifs, so it was just to keep the flow.

I'm more inclined to do the 2nd. If we include the 1st, we'd move into Android Apps, and that's a whole other world that would easily fill the page, and that's not the purpose imo, it's to provide people with basic/ground level tools.

kewde:
@hugoncosta I'm more inclined to do the 2nd too.

jj-pietro:
@hugoncosta may you add AFWall+ (firewall) and Adaway (for hosts files) to the root section? If you are rooted they're better than Netguard because you can keep using a vpn (Netguard used a local vpn to block unwanted traffic, and Android only supports one vpn service, AFWall+ uses iptables instead)

CHEF-KOCH:
Please revert this pull request.

1. The description is wrong, Xposed is a pre-requirement (the framework). 
2. Magiks is not needed, however you can install it in order to flash modules, I guess that is what the author liked to say. But Xposed repo installation process is the most common method.
3. The category and it's apps doesn't explain the consequences rooting the device, which is needed, even if the apps might directly not need su, the framework or the upper layer they using need higher permissions.
4. There good arguments that such apps not hold what they promise, in fact there wrongly advertised and there was never a code audit, or evidence that it holds what it promises, which makes me question why we need to recommend it. 
5. There is no guide or mention that XPL needs pro/additional hooks in order to address several things, that said there is no proper guide which tells you how to use it (also not in the official documentation). And it does hide an important fact, that there several things are not 'fakable' or not hida (everything) there is only one or two examples given. However it's worth nothing to hide/fake your IP when your Mac address transmits your MAC which currently isn't covered by XPL.
6. XPL is not an security application or manager, that's a marketing phrase. It's simply a Xposed Module which has ability to hook and intercept into several things.
7. Some AV products might not like Xposed / hooking / root because it can be a security risk (in the wrong hands).

There more things but I guess it's clear what we need to check and explain. If the original author doesn't correct or like this I would offer myself to make a PR in order to 'pimp' this.

Namnodorel:
_Note: For anyone reading only this PR, please also do consider to read the discussion in issue #399. Although some of CHEF_KOCHs points are true, a few of them are based on misguided assumptions that have been discussed before._

CHEF-KOCH:
* Xprivacy is not a tool. 
* XprivacyLUA is to make money to scare people with no proof that it helps. I remember everyone on the pokemon xposed modules. None of them worked anymore since they triangulate your position now, even if you fake your position they still see your ID and real-location. there was an Interview how this was/is done. Just google it.
* No one should be forced to root/unlock bootloader (call it what you want) to install a module from someone which can't provide any evidence that this has a security benefit. Xposed itself is detected by safetynet for a reason, some malware uses to hook into other system related or app related settings (that's btw malware behavior) same like XprivacyLUA can do. Most AV companies would see this as critical.
* Defeating surveillance with spoofing is already known as ineffective. 

My research paper is avbl. [here](https://github.com/CHEF-KOCH/Online-Privacy-Test-Resource-List/tree/master/Documents). You maybe start with 'Why mac spoofing is not enough". 

Snowdens apps and Xposed modules should be banned from this repo since there questionable, one hooks (which is malware behavior) and the other app uses surveillance in order to defeat surveillance (which is in some countries even illegal to use without confirmation/approval).

I would do a pull but only in order to revert it and remove the module.

PR #413, opened 2018-03-26
Add org mode to digital notebooks section (worth mentioning)

Akyshnik:
Add org-mode as it is a great TODO and note-taking tool.

### Description
Org mode is simple yet powerful mode for emacs for writing notes, maintaining TODO lists, and, as the name suggests, organizing your work.

### HTML Preview
http://htmlpreview.github.io/?https://github.com/Akyshnik/privacytools.io/blob/patch-1/index.html


Hillside502:
Noticed that you're using "org mode" and "org-mode" inconsistently.

Suggest the Wikipedia "org-mode" throughout.
https://en.wikipedia.org/wiki/Org-mode

PR #422, opened 2018-04-01
Change Open-Whisper-Systems.png to Signal.png

NinebitX:
### Description

* Deleted Open-Whisper-Systems.png (jagged, about 2 KB)
* Added Signal.png (smooth, about 3 KB)
* Changed Open-Whisper-Systems.png to Signal.png (under index.html)

### HTML Preview

http://htmlpreview.github.io/?https://github.com/NinebitX/privacytools.io/blob/master/index.html


c0rdis:
@Shifterovich looks ok to me although I'd probably leave the previous img name and wouldn't change index.html :)

Shifterovich:
I think Signal is better as it refers to the app, not the organization behind it. Either way, merging.

PR #425, opened 2018-04-03
Replace Android with BSD / Solaris under i2p

Avizini:
### Description
This PR removes F-Droid in OS line under i2p. Check commits for better understanding.

F-Droid is not an OS, it will be confusing for new users. Instead can we add another line saying, "Available on F-Droid"?

### HTML Preview
http://htmlpreview.github.io/?https://github.com/avizini/privacytools.io/blob/master/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -1389,7 +1389,7 @@
 								<button type="button" class="btn btn-success">Website: i2pbote.xyz</button>
 							</a>
 						</p>
-						<p>OS: Windows, Mac, Linux, Android.</p>
+						<p>OS: Windows, Mac, Linux, Android, BSD / Solaris.</p>
 					</div>
 				</div>
 			</div>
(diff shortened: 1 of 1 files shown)


c0rdis:
@Shifterovich makes sense to me. I'd add BSD / Solaris instead (as per https://geti2p.net/en/)

Shifterovich:
geti2p.net:

- Windows
- Mac OS X
- Linux / BSD / Solaris
- Debian / Ubuntu
- Android

Debian/Ubuntu is Linux. Therefore:

**OS: Windows, Mac, Linux, Android, BSD / Solaris.**

@Avizini

Avizini:
@Shifterovich @c0rdis Look's good now?

Shifterovich:
Yeah. Also I wouldn't worry about F-Droid, as people can find that on the I2P website.

PR #428, opened 2018-04-04
Reinstate DNSCrypt

NPN:
### Description

Add back DNSCrypt, as per discussion in #384. List of platforms has also been updated.

We may also want to change the logo, as the current one does not match the one on the website.

### HTML Preview

http://htmlpreview.github.io/?https://github.com/NPN/privacytools.io/blob/reinstate-dnscrypt/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -2452,9 +2452,11 @@
 						<p><img src="img/tools/DNSCrypt.png" alt="DNSCrypt" align="right" style="margin-left:5px;">A protocol for securing communications between a client and a DNS resolver. The DNSCrypt protocol uses high-speed high-security elliptic-curve cryptography and
 							is very similar to DNSCurve, but focuses on securing communications between a client and its first-level resolver.</p>
 						<p>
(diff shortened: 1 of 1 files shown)


g-monk:
Question: As you guys know I’m new and learning, I wanted to ask why DNSCrypt was removed in the first place.

Shifterovich:
I think it was not being maintained or something like that.

zero77:
@NPN 
I think DNSCrypt2 should be mentioned as it is fairly comprehensive now and it very stable.

NPN:
@g-monk The issue I linked to above explains it. The basic reason is that v1 was dropped. Then, the website was redirected and eventually squatted. So, we removed it, because it seemed that the project might have been abandoned. Now, there's a new website and DNSCrypt-Proxy 2 is stable, so I think we should add it back.

@zero77 This change links to the DNSCrypt website, which has a big button that links you to "Implementations". DNSCrypt-Proxy 2 is at the top of that list. I think this is a better solution because it allows people learn about the project and then choose what implementation they want to use.

Issue #432, opened 2018-04-09
Add disclaimer to signal


ghost:
As you know, signal requires you to use your phone number in order to use it, but in some countries(like mine), phone numbers are tied to specific people. For example, here, you have to give your id number and use your fingerprint in order to get a phone number, even for prepaid ones! (dunno how it is for other countries).
That is why, at least in my case, using signal pretty much defeats the whole privacy thing.
Please excude my paranoia :P

Shifterovich:
Thanks, and you can use a foreign service https://infosec-handbook.eu/blog/signal-myths/#m2

zero77:
@zeroweb91 
Although, you have to use a phone number to register and for verification, you can use a disposable number.

There are many free disposable number services or it is possible to create your own.

Vincevrp:
@Shifterovich this was implemented in #436, but it seems to be gone?

Shifterovich:
No idea why. Anyway, the warning should be implemented by a label, similar to Riot.

libBletchley:
@zeroweb91
> Please excude my paranoia :P

You're not paranoid in this case:

([PDF](https://github.com/privacytoolsIO/privacytools.io/files/3102795/phone_registration.pdf))
![phone_registration](https://user-images.githubusercontent.com/18015852/56492243-7e442c80-64eb-11e9-90d0-39a919487e44.png)

blacklight447-ptio:
adding a phone number is bad for anonymity, not privacy, signal never claimed to be anonymous.
anonymity is not part of its threat model. closing issue.

PR #436, opened 2018-04-09
Added warning about Signal

sevengali:
### Description

I added a warning about Signal needing a mobile number to register. #432 

First commit about ghostbin isn't present in this branch (is #435)

### HTML Preview

http://htmlpreview.github.io/?https://github.com/sevengali/privacytools.io/blob/sevengali-signal/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -2580,18 +2580,6 @@
         <div class="col-sm-4">
           <div class="panel panel-success">
             <div class="panel-heading">
-              <h3 class="panel-title">Ghostbin</h3>
-            </div>
-            <div class="panel-body">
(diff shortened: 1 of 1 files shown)


Shifterovich:
Thanks. I'd maybe change "mobile number" to "phone number". I think that's more common.

zero77:
@sevengali 
Although, you have to use a phone number to register and for verification, you can use a disposable number.

I think this is important to mention as there are many free disposable number services or it is possible to create your own.

Shifterovich:
Set up a PIN if using a disposable number https://infosec-handbook.eu/blog/signal-myths/#m2

zero77:
@Shifterovich 
>Set up a PIN if using a disposable number

yes, good point this should be mentioned also.

Shifterovich:
@kewde The disposable number thing isn't mentioned.

kewde:
@Shifterovich cd8f1e8

PR #461, opened 2018-05-20
Add Germany to Key Disclosure Subsection

the-coffee-coder:
### Description
Under German law, you are allowed to refuse giving your keys to authorities.


Shifterovich:
Some source of this information please?

kewde:
> However, there is no legal basis that would compel the user to turn over an encryption or network key, in particular with regard to the nemo tenetur principle.
https://www.loc.gov/law/help/encrypted-communications/germany.php#encryption-communications

PR #466, opened 2018-05-25
Overhaul of the Password generator

Strappazzon:
### Description
I have made the following changes:
 - The password generator now has the same style of the homepage
 - Changed how the messages are displayed on the page
 - Added the navigation bar with a link to the homepage
 - Added an option to include ambiguous characters when generating the password
 - Added an option to include brackets when generating the password
 - Changed the glyph for the Advanced Settings button
 - Replaced the dropdown menu with a text box (the range is still 6 - 100 characters)

### HTML Preview
http://htmlpreview.github.io/?https://github.com/Strappazzon/privacytools.io/blob/patch-2/password.html

### DIFF
--- a/password.html
+++ b/password.html
@@ -1,174 +1,144 @@
 <!DOCTYPE html>
+
+<!-- set the default language as English -->
 <html lang="en">
 
 <head>
-  <meta charset="UTF-8">
-  <title>Secure Password Generator | Privacy Tools</title>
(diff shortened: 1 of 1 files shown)


Shifterovich:
#467 I suggest replacing it with a KeePass recommendation.

ohmynameisrico:
Yeah, creating a password within the browser is not a good idea. Why not use offline software(outside the browser)?

PR #469, opened 2018-05-26
Add an arrow to scroll at the top

Strappazzon:
### Description
When the user scrolls the page down an arrow will show up. When clicked the page will scroll to the top.

### HTML Preview
https://htmlpreview.github.io/?https://github.com/Strappazzon/privacytools.io/blob/patch-3/index.html


Shifterovich:
Seems too large. I think there's some standard size that most pages use. @kewde

Vincevrp:
You can scroll to the top by clicking the home icon.

Issue #474, opened 2018-05-29
Add a warning to the Ricochet description


ghost:
https://github.com/ricochet-im/ricochet/issues/579

It's already dead.

ilovelinux:
It's not dead. As I answered to you in the issue you referenced and in https://github.com/ricochet-im/ricochet/issues/578, Ricochet is sicure and development is still going on. There are no reasons to remove it from "Encrypted Instant Messanger".

ohmynameisrico:
@ilovelinux Are you blind? You are not a ricochet developer anyway. Read.

tya99:
I don't think this should be removed as there's nothing to cause reason that it does not work. The security of this mostly uses Tor Hidden Services.

tya99:
Sure I've got this.

tya99:
How does this look:

![warning-ricochet](https://user-images.githubusercontent.com/42108382/48978307-8e252500-f0a1-11e8-8726-c526e4197578.png)

Shifterovich:
Looks good! The only change I'd make is: <code><a href="#">[icon] See below</a></code> -> <code>See below: <a href="#">[icon] Updating the Tor binary included with Ricochet</a></code>.

So: `[Danger]` <strong>Always keep Tor up to date. See below: <a href="#">[icon] Updating the Tor binary included with Ricochet</a></strong>

And maybe add a command to copy it on Linux and OS X? Hm, maybe a symlink would be better, because updating Tor browser would update the binary for Ricochet as well? And can't the tor binary location be changed in Ricochet's settings? It feels kinda hacky to copy the binary.

The symlink command would look like this:

```sh
$ mv ~/Downloads/ricochet/tor ~/Downloads/ricochet/tor.old
$ ln -s ~/Downloads/tor-browser_en-US/Browser/TorBrowser/Tor/tor ~/Downloads/ricochet/tor
```

tya99:
> Looks good! The only change I'd make is: `[[icon] See below](#)` -> `See below: [[icon] Updating the Tor binary included with Ricochet](#)`.
> 
> So: `[Danger]` **Always keep Tor up to date. See below: [[icon] Updating the Tor binary included with Ricochet](#)**

Easy done.

> And maybe add a command to copy it on Linux and OS X? Hm, maybe a symlink would be better, because updating Tor browser would update the binary for Ricochet as well? And can't the tor binary location be changed in Ricochet's settings? It feels kinda hacky to copy the binary.

[...]

PR #476, opened 2018-05-29
Remove Ricochet as it is already discontinued in 2016.

ohmynameisrico:
## The Tor Project said Ricochet is a "Discontinued software" thus it shall be removed.

Software listed here has become depreciated and development has ceased. They are therefore considered unsafe and it is highly recommended to avoid using them. This section is here purely for historical value.
Ricochet IM is an open-source, decentralised instant messenger project that officially ended development in November 2016. Ricochet starts a Tor onion service on the users local system and facilitates communication with other Ricochet users whom are also running their own Ricochet-created Tor onion service, providing End-to-End encryption by never allowing the connection to leave the Tor network.

### HTML Preview

*Replace [GITHUB_USERNAME] with your GitHub username and [BRANCH] with the branch name.*

http://htmlpreview.github.io/?https://github.com/ohmynameisrico/privacytools.io/blob/patch-1/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -1596,26 +1596,6 @@
 				</div>
 			</div>
 
-
-			<div class="col-sm-4">
-				<div class="panel panel-warning">
-					<div class="panel-heading">
-						<h3 class="panel-title">Desktop: Ricochet</h3>
-					</div>
-					<div class="panel-body">
-						<p><img src="img/tools/Ricochet.png" style="margin-left:5px;" align="right">Ricochet uses the <a href="#browser"><span class="glyphicon glyphicon-link"></span> Tor network</a> to reach your contacts without relying on messaging servers. It creates a hidden service, which is used to rendezvous with
-							your contacts without revealing your location or IP address. Instead of a username, you get a unique address that looks like <em>ricochet:rs7ce36jsj24ogfw</em>. Other Ricochet users can use this address to send a contact request - asking to be
-							added to your contacts list.</p>
-						<p>
-							<a href="https://ricochet.im/">
-								<button type="button" class="btn btn-warning">Download: ricochet.im</button>
-							</a>
-						</p>
-						<p>OS: Windows, macOS, Linux.</p>
-					</div>
-				</div>
-			</div>
-
 		</div>
 		<h3>Worth Mentioning</h3>
 		<ul>


ohmynameisrico:
https://github.com/ricochet-im/ricochet/issues/579
https://github.com/privacytoolsIO/privacytools.io/issues/474

hugoncosta:
With the availability of tens of services like this, I guess removing one that doesn't have proper maintenance is a must. @Shifterovich?

pahakalle:
what is going to take its place?

beardog108:
Its not actually [fully] discontinued. I don't think it should be removed unless someone can show these supposed "10 known vulnerabilities", the Tor wiki page doesn't even mark it as discontinued anymore.

Ricochet is relatively simple, and if it had *known* unpatched vulnerabilities, the authors would make it clear that its not safe to use. 

In addition Ricochet (the protocol anyway) is being extended to support group messaging via the [Cwtch project](https://cwtch.im/), so development around it is not completely stopped.

Quote From dev on June 4 2018:

>I reached out to the author of the Ricochet section on that Tor wiki page and cleared things up. You're free to make judgments about development being stalled, and Ricochet's disclaimers are still in place (experimental, etc), but the "officially ended development" claim was incorrect and is factually false.

IMO Ricochet is just as safe/unsafe as the day of its last release, unless someone can link the vulnerabilities that are claimed to be known. Is it the safest messenger available? No, but it is pretty good, and remember, it did pass the security audit in 2016, which was for the then (and still latest) release.

I wouldn't be upset about moving it to "worth mentioning", since it is clearly not well supported. Tox is alright, but has its own sets of problems and is not as private as Ricochet, although since it is supported, it would be better for the featured spot.

# Tl;dr

I suggest moving it to worth mention like suggested, and move Tox to the recommended. 

Hopefully one day Ricochet is better supported, as it is simple yet secure (in concept) and anonymous, which is rare.

tya99:
As mentioned here https://github.com/privacytoolsIO/privacytools.io/issues/474#issuecomment-438347467 I don't think this should be removed.

I also don't think that "Tox" should be argued as a "replacement" they aren't really the same thing and fundamentally work quite differently.

See:

https://github.com/ricochet-im/ricochet/pull/580#issuecomment-394176462
https://github.com/ricochet-im/ricochet/issues/579#issuecomment-392788611

asddsaz:
Without any strong evidence of major security flaws, why should it be deleted?

If you want security with decent ease of use, you use Ricochet. I am unware of any alternative to take its place.

I2p-bote and Bitmessage are the closest other "options".

tya99:
This should be closed along with the duplicate https://github.com/privacytoolsIO/privacytools.io/issues/474

Issue #465, opened 2018-05-24
Should Seafile really be the first choice for a self-hosted cloud solution? You be the judge.


RichardRMatthews:
https://github.com/haiwen/seafile/issues/350

stevesbrain:
I respect that this issue needs attention drawn to it, though I'm not so sure this is really a useful issue to open.

RichardRMatthews:
More people should be aware of this before making a choice which self hosting cloud they want to use.

stevesbrain:
@RichardRMatthews My apologies - I mistakenly thought this was an open issue on the actual haiwen/seafile repo; was a derp on my end. I can totally see the relevance here :) Ignore me!

RichardRMatthews:
No problem , everyone mixes things up. Have a nice day!

stevesbrain:
You too :)

Vincevrp:
Seafile was removed in #491.

PR #491, opened 2018-07-03
Removed Seafile

sevengali:
### Description

Removed Seafile from both cloud storage providers and self hosted cloud storage. This is due to issues #465 and #490 - the encryption has large flaws and after 5 years, the developers seem to have no intent to fix this.

Under providers, bumped Muonium up to last space.

Under self hosted, bumped NextCloud up to last space.

Not having used anything other than NextCloud previously, I'm unsure of ordering - I'll leave this up to somebody else if there's a better order.

I also fixed the color of the button for Magic Wormhole.

### HTML Preview

http://htmlpreview.github.io/?https://github.com/sevengali/privacytools.io/blob/master/index.html


dakira:
@sevengali I really don't see the reason for this. The other proposed solutions don't offer client side encryption at all (Nextclouds solution is still in beta). As for the password hashing. That's just the matter of submitting a simple PR or issue. For the users where this matters (installations with many users) it usually shouldn't be a problem as they use other means of authentication in the first place (kerboros, shibboleth).

Looks to me like a hater on a crusade.

dakira:
@Shifterovich see above

Shifterovich:
@kewde thoughts?

PR #495, opened 2018-07-19
Add Pi-hole as worth mentioning under DNS

sjlehn:
### Description
Add Pi-hole (https://pi-hole.net/) to the suggestions for DNS.  Works quite well as a network DNS server, blocking ad and tracker domains (and anything else you feel like blocking).

### HTML Preview

http://htmlpreview.github.io/?https://github.com/sjlehn/privacytools.io/blob/master/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -2510,6 +2510,7 @@
 		<ul>
 			<li><a href="https://github.com/quidsup/notrack">NoTrack</a> - A network-wide DNS server which blocks Tracking sites. Currently works in Debian and Ubuntu.</li>
 			<li><a href="https://namecoin.info/">Namecoin</a> - A decentralized DNS open source information registration and transfer system based on the Bitcoin cryptocurrency.</li>
+			<li><a href="https://pi-hole.net/">Pi-hole</a> - A network-wide DNS server for the Raspberry Pi.  Blocks advertising and tracking domains for all devices on your network.</li>
 		</ul>
 
 		<a class="anchor" name="notebook"></a>


Hillside502:
Pi-hole - Wikipedia
https://en.wikipedia.org/wiki/Pi-hole

Source code
https://github.com/pi-hole/pi-hole

Open Source Pi-hole Alternatives - AlternativeTo.net
https://alternativeto.net/software/pi-hole/?license=opensource

hugoncosta:
Isn't this a bit "niche" since it's towards the RPi? I believe the tools here are supposed to be for every machine (or at most, divided by Software and/or Mobile/Computer)

Shifterovich:
Since it says `A network-wide DNS server for the Raspberry Pi`, I think it's fine. If you have a Raspberry Pi you can host it for all computers on your network. So on one hand you need the rpi but on the other hand it works as a server for all devices on your network, just like using a third party DNS. The description specifically says that it's for Raspberry Pi, so it's okay.

sjlehn:
@hugoncosta per their site, you can install it on most Debian based distros:

> Software
> 
> Pi-hole will run on most Debian-based distro’s and is the preferred platform for it.
> 
> We officially support the following:
> 
>     Raspbian: Jessie/Stretch
>     Ubuntu: 16.04 / 16.10
>     Debian: 8 / 9
>     Fedora 28
>     CentOS: 7 (not ARM)

https://discourse.pi-hole.net/t/hardware-software-requirements/273

Shifterovich:
Yes, but you need a server. Though I'm not sure how performant Raspberry Pi is for hosting a DNS server, does anyone here have experience with Pi-hole?

sjlehn:
I've been running it on my network on a Pi 3 model B for a while, as DNS and DHCP (less than a dozen clients) without any issues.  Memory usage seems to be sitting around 12%.

Shifterovich:
Does domain resolution seem slower than using a regular DNS server or are the speeds comparable?

PR #538, opened 2018-10-21
Create panel-item class to replace inline CSS

Vincevrp:
### Description

Replaced inline CSS and obsolete align attribute with a class for all images in `.panel-body`

### HTML Preview

https://htmlpreview.github.io/?https://github.com/Vincevrp/privacytools.io/blob/master/index.html

### DIFF
--- a/css/custom.css
+++ b/css/custom.css
@@ -24,6 +24,11 @@ a.titleanchor {
   font-size:0.7em;
 }
 
+.panel-item {
+  float: right;
+  margin-left: 5px;
+}
+
 .jumbotron {
     padding: 10px 0px;
 }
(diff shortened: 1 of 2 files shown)


Shifterovich:
Thanks. After your PR yesterday, I looked at the W3 validator and saw that there are indeed many, many issues.

Vincevrp:
I'm trying to fix most of them as we speak. Maybe we should implement some sort of CI/CD that validates the code. I'm willing to help implement this.

Shifterovich:
I thought about that a few days ago. We probably don't need something like StyleCI as it's not very relevant for HTML, but maybe some HTML validation CI would be useful. Many of the issues the W3 validator found would be ignored, though. For example \<button\> not being a child of \<a\>.

I'll have to check if I have enough permissions to set up a CI for PTIO, though, as I'm only a collaborator and @privacytoolsIO is not an organization.

PR #550, opened 2018-10-27
Replace obsolete anchor tags with .anchor class

Vincevrp:
### Description

Replace obsolete anchor tags with .anchor class.

### HTML Preview

No preview because of Jekyll.


Shifterovich:
Great. Even the margin is perfect. I added FA icons for the anchors.

Issue #549, opened 2018-10-26
Comma missing in "If you are currently using * you should pick an alternative here"


Shifterovich:
I'm not a native speaker, but it looks weird without the comma before `you`.

Aquakor:
I'm also not a native English speaker, but I agree with you. I think this is a first conditional sentence. If a conditional clause comes before a main clause, we separate them with a comma, [source](https://dictionary.cambridge.org/de/grammatik/britisch-grammatik/conditionals-and-wishes/conditionals). 

I think we should also consider changing the tense of a conditional clause from present continuous to present simple.

Shifterovich:
I don't think the continuous tense necessarily means "using right now". I think it highlights using it actively. `If you currently use Telegram` sounds like "if you use it at all". So in my opinion the continuous tense highlights that you use it actively - there's nothing wrong with using Messenger for some things and Signal for important things.

Though again I'm not a native speaker. We can fix the commas now and the tense possibly in future if we can get more opinions on it.

@Aquakor could you create a PR fixing the commas?

Aquakor:
@Shifterovich yes, I will create it.

PR #564, opened 2018-11-05
Fix grammar

Aquakor:
### Description

Some grammar fixes :pencil:.


Shifterovich:
Thanks. Could you take a look at #549 please? I'd like to know if the commas should be there (I'm not a native speaker).

PR #569, opened 2018-11-07
Optimize images

Vincevrp:
### Description

Applied lossless compression to all images. Gifs are converted to png to reduce file size.


Shifterovich:
Thanks!

Issue #579, opened 2018-11-10
Replace Youtube links for PeerTube


Booteille:
Hi!

Since PeerTube released I think this could be a good idea to re-upload all videos (2 atm) used on privacytools.io on PeerTube and use those links instead of YouTube.

What do you think about that?

Links:
PeerTube: https://joinpeertube.org
Edward Snowden video on PeerTube: https://peertube.mastodon.host/videos/watch/4cdedd90-a5b4-4022-b93d-828e85ed58cd
DEFCON 20 video on PeerTube: https://peertube.mastodon.host/videos/watch/e09915eb-5962-4830-a02f-8da5c2b59e71

beerisgood:
Maybe https://invidio.us / https://github.com/omarroth/invidious too

Booteille:
Yeah. I thought about it but I would prefer PeerTube over invidio.us because invidio.us is a fron-end for YouTube. It still has to call Google servers to obtain the video and I would prefer a solution who gets totally ride of Google.

Shifterovich:
:+1:

beerisgood:
> 
> 
> Yeah. I thought about it but I would prefer PeerTube over invidio.us because invidio.us is a front-end for YouTube. It still has to call Google servers to obtain the video and I would prefer a solution who gets totally ride of Google.

Sure. But the problem with PeerTube is that not that big content from youtube exist their. So i would set Invidio as fallback (which btw not even need JavaScript to display videos!)

Booteille:
Sure. We could use Invidio.us as a fallback.
But the point is since we only have two videos from Youtube displayed on the website, we could just upload the missing one and use peertube links instead of YouTube. We don't need there is any other content on PeerTube than the videos we want to be displayed on the website.

Shifterovich:
> (which btw not even need JavaScript to display videos!)

JS is not inherently bad. It's bad on evil sites which use it to track you. I highly doubt PeerTube is one of those sites.

pdjpdjpdj:
I see those 2 vids on GoogleTube links in the entire repository.
A simple replace should do the job s/youtube.com\/oldlink/thoselinks.

PR #590, opened 2018-11-15
Replace youtube links with PeerTube counterpart

Booteille:
## Description
Replaced YouTube links in sources by their PeerTube counterpart.

Issue: #579

### DIFF
--- a/index.html
+++ b/index.html
@@ -1500,7 +1500,7 @@ layout: default
 
 <h3>Related Information</h3>
 <ul>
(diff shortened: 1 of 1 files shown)


Booteille:
I don't see it merged. Any reason for that?

Shifterovich:
I messed up the conflict resolution, so we removed the commits from the log. @Vincevrp Can you apply the changes manually?

There was a conflict with something GH displayed as a binary change to the index.html file.

PR #605, opened 2018-11-19
Update Bitwarden logo and small modifications to description

kspearrin:
## Description

Update to use official icon for Bitwarden. See https://github.com/bitwarden/brand#product-icons

Also made some small changes to the description so that it reads better (IMO).

image


Vincevrp:
I used the black logo in #587 to match the yellow a bit better, but I agree we'd be better off using the official logo.

Issue #604, opened 2018-11-19
Wrong order in "Key disclosure laws don't apply"


Shifterovich:
https://www.privacytools.io/#kdl

Germany should come before Poland - alphabetic order.

PR #606, opened 2018-11-20
Fix list and order in KDL

Vincevrp:
## Description

Fix lists and order in key disclosure law section.

Resolves: #604

### DIFF
--- a/assets/css/style.scss
+++ b/assets/css/style.scss
@@ -21,6 +21,10 @@ a,
   }
 }
 
+.card-ol {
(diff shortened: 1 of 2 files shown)


Shifterovich:
Thanks, though this broke:



Feel free to push simple fixes directly to master.

BTW, the KDL section doesn't have a navbar link. Since Avoid all US services is in provider, this should be there as well. Though I suggest creating a new section in the navbar for the "theory".

PR #613, opened 2018-11-23
Add Riot Chat Client

asddsaz:
Fixes #597


Shifterovich:
- Add Wire to worth mentioning
- Signal should be before Riot. Not to mention that the colors are wrong.

Shifterovich:
Protocole should be protocol

asddsaz:
@Shifterovich fixed!

Shifterovich:
You removed a link to Wire's website and added a link to Wire's Google Play page.

asddsaz:
I'm so sorry, it should be fixed now :)

asddsaz:
@Shifterovich I'm so sorry, it should be fixed now :)

Shifterovich:
Maybe add a longer description?

asddsaz:
This branch has conflicts that must be resolved

Mikaela:
I think Riot should have experimental label like Brave, because it warns about E2EE being experimental and https://github.com/matrix-org/synapse/issues/1287 .

Shifterovich:
https://github.com/privacytoolsIO/privacytools.io/blob/master/index.html#L1280

Add this warning to the link to Wire in the worth mentioning section.

And I'll resolve the conflicts.

Shifterovich:
Make the warning similar to uTox's warning, but with a tooltip. Here's some sample code you can use for that:

```html
<span class="badge badge-warning" data-toggle="tooltip" title="Brave is a good choice if you want to use a Chromium-based browser. But at this point in Brave's development&comma; it's not as good as Firefox with privacy addons.">experimental <i class="far fa-question-circle"></i>
```

Mikaela:
Did you see my comment above requesting the experimental/warning label to Riot?

Shifterovich:
Sorry, didn't notice it when I was fixing the conflicts. I'll take a look at the issue you sent.

Kcchouette:
Can you remove the "successor of XMPP' part too please @Shifterovich @asddsaz

Shifterovich:
Right, we should remove that and think of a longer description.

Shifterovich:
@Mikaela Also the E2EE isn't enabled by default, right?

Kcchouette:
An example of description: "Riot.im is a client based on Matrix, a recent open protocol for real-time communication. A distributed chat client that offers E2E Encryption. It can bridge other communications via others protocols such as IRC too."

Mikaela:
> @Mikaela Also the E2EE isn't enabled by default, right?

That is the case, it's opt-in and only very few clients support it, mainly Riot. They say that there is no point in E2EEing public rooms, which I agree, and it would also prevent the bridges from working.

> Can you remove the "successor of XMPP' part too please

In case this wasn't changed yet, I don't think anyone working with IRC(v3?) has called them as successor of IRC either.

Mikaela:
Two other issues with Matrix or Riot coming to my mind:

* everyone can see your Matrix ID, I don't know if that is a major issue though, but [Disroot.org cites it as one reason for not liking them.](https://disroot.org/en/blog/matrix-closure)
* you cannot set the public device name before you login, so if your device is "John Doe's iPhone" (or ONEPLUS A3003), everyone can see that in your profile until you change it. Some other clients than Riot ask how to name the device.

Shifterovich:
Now I'm thinking that if the E2EE is experimental, what's the basis for recommending Matrix?

Mikaela:
In case anyone wishes, here are some sources on the experimentality:

* > All files and data transferred over Riot can be encrypted end-to-end *(currently in beta)*, meaning no one can eavesdrop on conversations, including the service provider.
    * emphasis mine, from http://about.riot.im/what-is-riot/#item8

* > As of May 2017 *Riot’s end-to-end encryption is technically in beta*, but this is due to some residual stability bugs and missing usability features. Once these are resolved we plan to get the full implementation security assessed and out of beta. End-to-end encryption will then be turned on by default for private conversations.
    * emphasis mine, from http://about.riot.im/security/

The strings from Android, I imagine iOS has similar (I don't think I need to dig them?):

* https://github.com/vector-im/riot-android/blob/cd0612cd1557b6d6f077dc9fbee18184aa536975/vector/src/main/res/values/strings.xml#L383
* https://github.com/vector-im/riot-android/blob/cd0612cd1557b6d6f077dc9fbee18184aa536975/vector/src/main/res/values/strings.xml#L888

Shifterovich:
From the first string file:

> You should not yet trust it to secure data.

I think this necessarily implies "You should not yet recommend it anyone to trust it to secure data."

Maybe we should add a `labels='warning:beta:The software is currently in beta and the website states "End-to-end encryption is in beta and may not be reliable. You should not yet trust it to secure data."'`

Mikaela:
> Maybe we should add a `labels='warning:beta:The software is currently in beta and the website states "End-to-end encryption is in beta and may not be reliable. You should not yet trust it to secure data."'`

I agree otherwise, but as the string is from Android client and not findable from the website I would say:

``labels='warning:beta:The software is currently in beta and the mobile client states "End-to-end encryption is in beta and may not be reliable. You should not yet trust it to secure data."'``

I only changed *website* to *mobile client* from your suggestion and I ended up digging for the iOS strings just in case it would disagree, but there is at least one of the same:

* > `"room_warning_about_encryption" = "End-to-end encryption is in beta and may not be reliable.\n\nYou should not yet trust it to secure data.\n\nDevices will not yet be able to decrypt history from before they joined the room.\n\nEncrypted messages will not be visible on clients that do not yet implement encryption.";`
* https://github.com/vector-im/riot-ios/blob/167bf8c12495f73f4ad636f3840af70589cb7adb/Riot/Assets/en.lproj/Vector.strings#L274

Kcchouette:
You can add too that it's not e2e by default

asddsaz:
> Right, we should remove that and think of a longer description.

Done, I also added an experimental flag.

Shifterovich:
Look at Brave in the source code, you can add the label using the code @Mikaela used in her comment.

Shifterovich:
Not sure about chatrooms but Signal can do group chats.

johnstonesnow:
I just posted on a thread before realising it was merged with this one and closed. I will post my message here in case Infosec is reading:

"Infosec-Handbook said: "Signal requires an arbitrary phone number which you must control during the registration process. However, you don't have to use your own private cellphone number. You can use the Tor Browser to get a disposable phone number for registration, set a PIN and never need access to the phone number during normal operation."

Sorry for chipping into an old thread, but I am currently torn between Wire and Signal. Two things put me off SIgnal:

1. Phone number needed (I read that you have to have permanent access to the number to keep Signal Desktop working (I only use desktop). 
2. Based in US

If I could remove downside "1", I might see SIgnal as a better option than Wire. Can you confirm if this is still true, that you don't need to use a permanently accessible phone number? If so, do you have a guide or link on how to sign up for Signal without using my own phone number? 

Thanks for any help, oh and PS - Now it's 2019, if you have any other recommendations please say. I tried Tox chat and it's private, but buggy as hell and I can't get people to use it. Signal and Wire are possibles because they have decent features, run properly, and I can get people to use them. But which one? :)"

PR #621, opened 2018-11-25
Add translations to the main site (+markdown transition) [DON'T MERGE]

Shifterovich:
We will slowly transition from HTML to Markdown, wherever practical. This will result in much more readable and newbie-friendly files.

The first commit in this branch shows how we will implement localization. The main site will be multilingual -- this doesn't mean that the forks can't have their own repos and domains. It will be much easier for them to keep up, due to the utilization of <strike>JSON</strike> YAML language files.


Shifterovich:
@privacytools-it @Strappazzon @c0rdis @jxtsai @jxtsai @Anon215 What do you think about using language files like this one: https://github.com/privacytoolsIO/privacytools.io/pull/621/files#diff-baf248d272932ca097c96e89f1d12bf8

If you pulled changes from the main repo, you'd have English content that you could easily localize one day just by creating a json file.

Strappazzon:
I'm fine with it, I like these changes.

Shifterovich:
- We'll use yaml instead of JSON to avoid painful escaping of `"`.
    - This is all you need to know:
        - yaml files start and end with a line of  `---`
        - single-line string: `key: value`
        - multi-line string: 
          ```yaml
          key: >
               value
          ```
- For some complex parts, we'll use HTML. Such as the nothing to hide quote.
- The files with content are stored in `_include/content`
    - They start with
      ```liquid
      {% assign lang = site.data.lang.[include.lang].<<filename>> %}
      {% assign en = site.data.lang.en.<<filename>> %}
      ```
        - `<<filename>>` being the name of the file without the extension
- Inside includes, you use localized strings like this:
  ```liquid
  {{ lang.key | default: en.key }}
  ```
- Inside `index.html`, you use localized strings like this:
  ```liquid
  {{ language.key | default: english.key }}
  ```
    - The variable names are different from includes, because including something in `index.html` would override the variables.
- `index.html` is now in `_includes`
- Files in `languages/` include `index.html` and set the language for it.

Shifterovich:
Instead of having index full of includes, we could keep the content there and simply replace the strings with `{{ lang.key }}`.

Then, index.html would look like this:

```html
<!-- Nothing to hide section -->
{% assign local = site.data.lang.[include.lang].nothing_to_hide %}
{% assign en = site.data.lang.en.nothing_to_hide %}
{% assign lang = local | default: en %} <!-- maybe this wouldn't work and we'd have to use capture; not important -->

<h1>{{ lang.title }}</h1>
```

**However, this would mean that the language files have to contain *all the strings*.**

I'll think about this.

Shifterovich:
The 3 assigns can be turned into one include. Any variable defined within any include within index.html will be in index.html's scope, since index.html is also an include and it seems like includes share a scope.

Shifterovich:
- All strings don't have to be translated. We'll use `{{ lang.title | default: en.title }}`.
    ```
        {% include lang for="nothing_to_hide" lang=selected %}
        {{ lang.title | default: en.title }}
    ```
- Content will remain in index.html

Shifterovich:
I haven't yet decided what all will be converted into markdown.

It would be cool if we could convert

`# headings`

into headings with an anchor link and a FA link icon.

I guess moving the VPN comparison table to a separate file will be great, as index.html will be far less bloated.

screenshot from 2018-12-15 17-28-35

The navigation thing on the right shows how big the VPN comparison table is.

---

Things like

```html
<h3>Related Information</h3>
<ul>
    <li><a href="https://www.bestvpn.com/the-ultimate-privacy-guide/#avoidus">Avoid all US and UK based services</a></li>
    <li><a href="https://en.wikipedia.org/wiki/Surespot#History">Proof that warrant canaries work based on the surespot example.</a></li>
    <li><a href="http://en.wikipedia.org/wiki/UKUSA_Agreement">http://en.wikipedia.org/wiki/UKUSA_Agreement</a></li>
    <li><a href="http://en.wikipedia.org/wiki/Lavabit#Suspension_and_gag_order">http://en.wikipedia.org/wiki/Lavabit#Suspension_and_gag_order</a></li>
    <li><a href="https://en.wikipedia.org/wiki/Key_disclosure_law">https://en.wikipedia.org/wiki/Key_disclosure_law</a></li>
    <li><a href="http://en.wikipedia.org/wiki/Portal:Mass_surveillance">http://en.wikipedia.org/wiki/Portal:Mass_surveillance</a></li>
</ul>
```

Can be converted into markdown easily and it will be a great thing.

I guess this could be converted nicely into markdown, though I'm not sure about `img-fluid float-right`.

screenshot from 2018-12-15 17-30-30

It's important to keep in mind that we can freely use HTML in MD. At least on GitHub. Seems to be that way for most MD compilers. So this will likely be true in Jekyll/Liquid as well.

---

Will we move these sections to separate files too?

screenshot from 2018-12-15 17-34-42

c0rdis:
@Shifterovich what do you think of adding icons to the header template as they are now present at https://privacytools.ru? Also, I've made some re-ordering of the sections (and actually added a couple, which are only relevant to RU - so probably we should consider "floating sections"), so if you believe it makes sense, I can help with both things on the main template.

Shifterovich:
Good idea, that might for now solve the issue with the messy navbar. It will be easier to navigate in it.

BTW, after discussing this with @Vincevrp we came across two things:
- Both YAML and translated markdown files are a trade-off. Simply splitting the website into .md files per sections isn't.
- Maintaining the translations would mean that most pull requests would be related to translations. It's a better idea to split index into markdown files, making it easier for forks to stay up to date (with less conflicts etc).

So we will simply split index.html into markdown files by sections.

Vincevrp:
Closing in favour of #730.

PR #627, opened 2018-11-26
There are NO active key disclosure laws in Sweden.

gjhklfdsa:
## Description
There are no active key disclose laws in Sweden (only proposed).
Source: https://www.wikipedia.org/wiki/Key_disclosure_law#Sweden

Resolves: #none


Shifterovich:
Thanks. Could you just move it after Poland to keep the order alphabetic?

gjhklfdsa:
@Shifterovich Sorry, I didn't notice there was any order to the list. It should be fixed now!

PR #634, opened 2018-11-27
Add Silence

gjhklfdsa:
## Description
Add Silence.im to Encrypted Instant Messenger Worth Mentioning list.

**What is Silence**? Silence is a application to encrypt SMS/MMS similar to PGP/GPG. It is based on the Signal encryption protocol and currently supports Android and most Android forks like CopperheadOS and Replicant.

Resolves: #none


Shifterovich:
I agree.

Zig-03:
May I add that the idea is NOT to promote SMS/MMS over anything, but to say that **In case you really have no other alternatives, but to use SMS/MMS, then use these apps (Silence and perhaps there are more)**. After all encrypted SMS/MMS is better than non-encrypted SMS/MMS, no?

Atavic:
SMS are security-wise a disaster, like emails that leak metadata. If someone has to use SMS, then signal seems a less evil choice (into an SMS disastrous situaton).

Zig-03:
>  then signal seems a less evil choice (into an SMS disastrous situaton).

Why? As far as I know, Signal sends regular SMS as regular SMS, no encryption, nothing.

Atavic:
Obviously, I meant Silence.

PR #636, opened 2018-11-27
remove the old KeePassX completely

quiddity-wp:
After 2.5+ years without change, I suggest removing the old KeePassX completely and slightly expanding the explanation for KeePassXC. Thanks!

### DIFF
--- a/index.html
+++ b/index.html
@@ -1522,13 +1522,13 @@ layout: default
 
 
   {% include card.html color="primary"
-  title="KeePass / KeePassX - Local"
+  title="KeePass / KeePassXC - Local"
   image="assets/img/tools/KeePass.png"
   url="http://keepass.info/download.html"
   website="keepass.info"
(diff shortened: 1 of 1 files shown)


Shifterovich:
Maybe we could go as far as remove KeePass totally and keep only KeePassXC? I've heard from some people that they use KeePassXC because they don't want to use mono to run KeePassX.

beerisgood:
@Shifterovich under Windows the original is still the best

Shifterovich:
In what sense is it better than KeePassXC on Windows? Recommending multiple forks is somewhat confusing. I think we should pick one and recommend only it.

beerisgood:
@Shifterovich There are security differences....

Keepass has Secure Desktop & Two-Channel Auto-Type Obfuscation while KeepassXC doesn't

with Keepass I can create multiple password templates and a default password template for newly generated entries.

Overall I think you have more options and control in Keepass.

Shifterovich:
Okay.

PR #643, opened 2018-12-04
Five Eyes countries: add missing numbers in list

tuxayo:
Australia had a number but not the others. And the Nine Eyes and Fourteen Eyes had numbers too.

## Description

Fixes an inconsistency:
Current:
screenshot_2018-12-04_12-02-34

But this is still an inconsistency between the 3 lists

Do we put bullets on the 3 lists? Or no bullets at all?


Shifterovich:
Use an *ordered list* instead of hardcoding the numbers.

PR #644, opened 2018-12-06
Add Iceland to "Key disclosure laws don't apply"

asddsaz:
## Description

Iceland doesn't have any key disclosure laws.
Source: https://archive.fo/jIBUA

Resolves: #none


BurungHantu1605:
thanks

PR #646, opened 2018-12-06
Fix numbering of Five Eyes countries

torqueyy:
## Description

Australia was the only country in the column to have a number. This change makes the other 4 countries match the formatting of all Fourteen Eyes columns.


Shifterovich:
Use an ordered list instead of hardcoding the numbers.

PR #659, opened 2018-12-11
Replace Laverna (deprecated) with Joplin

matsest:
## Description

As Laverna is [dead](https://github.com/Laverna/laverna/wiki/DEAD-PROJECT-ALERT) and [several users](https://www.reddit.com/r/privacytoolsIO/search?q=joplin&restrict_sr=1) have suggested adding Joplin to the digital notebook section, Joplin indeed seems like a valid replacement. It respects the criterias: open source, cross-platform, easy to use and privacy respecting. It's not a browser-only app like Laverna, but it offers applications for all major platforms.

It currently has 5.8k stars on GitHub and offers a fully featured, self hosted client-based application with sync through Nextcloud or similar. Their development is open and engages with users actively through GitHub and their [support forum](https://discourse.joplin.cozic.net/). Their apps are 100% free, so it's not like StandardNotes which is more of a freemium service. It also offers a much simpler import from Evernote notes than SN. 

Details on Joplin can be found on their [web site](https://joplin.cozic.net/) or in their [repo](https://github.com/laurent22/joplin). Their encryption spec is also described [here](https://github.com/laurent22/joplin/blob/master/readme/spec.md) or possible to validate in their source code.

This PR kan be kept open for discussion, since it's not thoroughly discussed through any issues here - only on Reddit.


matsest:
Thanks for the feedback!

* I added a compressed version with the command you provided (045943f)

* I added a warning about the missing password protection (62137d4)
  * Note: As both the other alternatives offer this I agree that a warning is nice to have. For users using it on a shared device, this app is likely not the best alternative for now, but in general lots of apps have data accessible if a user is already logged in. But I can see that a pin protection like SN has would be nice to have.

PR #663, opened 2018-12-15
Flag fix on the site

Ud4ba:
You should fix the flag , at the VPN section at EarthVPN ... 
This is not  Northern Cypru's flag ...
And lastly you should check Earth's VPN laws there .


Shifterovich:
Hi. To create a PR, use https://github.com/privacytoolsIO/privacytools.io/compare and click the "Compare across forks" button. Then select your fork and the correct branch. Thanks.

PR #667, opened 2018-12-18
Move http → https when it exists

Kcchouette:
## Description

Move link from http to https when it's possible


Shifterovich:
Thanks!

PR #669, opened 2018-12-19
Enable sitemap.xml generation & reintroduce robots.txt

Mikaela:
I disagree with the decision to remove robots.txt in #668. I think it may at least remove 404 errors from logs and sitemap.xml can only help with SEO or won't be harmful.

However I am not sure if privacytools.io has more than one page and I was unable to get Jekyll working locally on this PC.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -3,4 +3,7 @@ description: You are being watched! Knowledge, encryption and privacy tools to p
 url: "https://privacytools.io"
 sass:
   style: compressed
-
+plugins:
+  - jekyll-sitemap
+sitemap:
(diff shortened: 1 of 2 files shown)


Shifterovich:
What's the point of having robots.txt when it simply says "all pages allowed"?

I see the point with a sitemap, though.

Shifterovich:
Though as I expected, the sitemap is pretty useless since we have only two pages:

screenshot from 2018-12-19 13-15-39

Vincevrp:
Can you verify that the `jekyll-sitemap` plugin is supported by Github Pages?

PR #671, opened 2018-12-20
Add Video Streaming Section

asddsaz:
**Description**: Add section for Video Streaming providers.

Added Providers:
To main:

- LBRY

- PeerTube

- FreeTube/Invidio

To Worth Mentioning:

- KopyKate (**Warning: Uncensored**)

-  DTube


Shifterovich:
@Vincevrp

hugoncosta:
I'd just change the warning, I believe it should be inside LBRY's card as it does not pertain to all of them. KopyKate I don't believe it has that much "worth" as it's a project still in alpha without much activity in zeronet.

ADepic:
I think you should remove invidious and freetube since they fetch youtube videos and are not streaming services. I think its better to have a seperate tab for streaming services like peertube and LBRY and a different one for youtube frontends that fetch youtube videos. Check out my PR #694 since now it only have youtube frontends.

I think Bitchute should also be in the worth mentioning, because although it isn't federated, it is a lot more popular than peertube.

index.html
```diff
@@ -1727,6 +1727,53 @@ <h3>Worth Mentioning</h3>
...
+    url="https://lbry.io/"
+    footer="OS: Windows, macOS, Linux, iOS, Android."
+    description="LBRY is a decentralized protocol for sharing data, similar to BitTorrent or ZeroNet."
+    %}
+
+    {% include card.html color="primary"
+    title="PeerTube"
+    image="assets/img/tools/peertube.png"
+    url="https://joinpeertube.org/"
+    footer="OS: Windows, macOS, Linux, Web."
```
ADepic: Peertube has an android app. Does it have desktop applications?

index.html
```diff
@@ -1727,6 +1727,53 @@ <h3>Worth Mentioning</h3>
...
+    description="PeerTube is a federated video streaming alternative."
+    %}
+    
+    {% include card.html color="warning"
```
ADepic: IMHO this should be remove due to it not being a streaming service and added to a different section.

Also freetube and invidious are two different softwares

index.html
```diff
@@ -1727,6 +1727,53 @@ <h3>Worth Mentioning</h3>
...
+    title="Invidio / FreeTube"
+    image="assets/img/tools/freetube.png"
+    url="https://github.com/FreeTubeApp/FreeTube"
+    footer="OS: Windows, macOS, Linux, Web."
+    description="Invidio is a privacy respecting API for YouTube."
+    %} </div>
+    
+
```
ADepic: Maybe put dtube here instead of worth mentioning to replace invidious?

index.html
```diff
@@ -1727,6 +1727,53 @@ <h3>Worth Mentioning</h3>
...
+  <ul>
+    <li>Note: As of Dec 2018, LBRY does not have official support for TOR or I2p.</li>
+  </ul>
+    
+    
+    
+  <h3>Worth Mentioning</h3>
+  <ul>
+    <li><a href="https://github.com/misses-robot/KopyKate-Big/blob/master/README.md">KopyKate</a> - A decentralized uncensored video streaming service on ZeroNet.</li>
+    <li><a href="https://d.tube/">DTube</a> - A distributed video streaming service based on IPFS.</li>
```
ADepic: Maybe add bitchute here since it is more popular p2p streaming service than the others, and move dtube up

gjhklfdsa: I would, but cannot find Bitchute's source code. This leads me to believe it is non-free, is this correct? If you can find the source code (under a free software license) I would certainly consider adding it.

@ADepic

index.html
gjhklfdsa: > Also freetube and invidious are two different softwares

I understand, but Freetube is based on Invidio[1](https://github.com/omarroth/invidious#made-with-invidious). This is meant to be similar to how KeePass and KeePassXC are together[2](https://www.privacytools.io/#pw).

> IMHO this should be remove due to it not being a streaming service and added to a different section.

Remove PeerTube? I thought PeerTube was doing a really good job, I would be open to considering changing it to a PeerTube provider. This might be a better choice.

What do you think should replace it?

index.html
gjhklfdsa: Yes, https://lbry.io/get?showall=1
I'll fix this.

asddsaz: Yes, https://lbry.io/get?showall=1
I'll fix this.

index.html
asddsaz: > Also freetube and invidious are two different softwares

I understand, but Freetube is based on Invidio1. This is meant to be similar to how KeePass and KeePassXC are together2.

> IMHO this should be remove due to it not being a streaming service and added to a different section.

Remove PeerTube? I thought PeerTube was doing a really good job, I would be open to considering changing it to a PeerTube provider. This might be a better choice.

What do you think should replace it?

index.html
asddsaz: I would, but cannot find Bitchute's source code. This leads me to believe it is non-free, is this correct? If you can find the source code (under a free software license) I would certainly consider adding it.@ADepic

index.html
asddsaz: @ADepic It is mainly accessed via the web. Can you send Android client URL?

ADepic:
Yes Bitchute is closed source, however it is a hosted service so it doesn't matter if the source code is open, since they could host any code they wanted on their servers. The main reason online software is open source is so that it can be federated and self hosted. Bitchute is neither, and thats why it should only be in worth mentioning.

I think it should be there so that people can find a platform with some actual content creators. Lots of people harmed by youtube flock to bitchute, so it actually has content on it.

There should be a warning about the fact that it is not federated, but say the reason it is in worth mentioning is so that you can find a youtube alternative with some content on it.

asddsaz:
@ADepic Non-free Software is an issue directly related to privacy.
This is not something I would recommend adding. I would recommend limiting its usage.

More info on free software: https://www.fsf.org/about/what-is-free-software

ADepic:
Bitchute isn't software in the sense that it is free/non-free.
Read this:

https://www.gnu.org/philosophy/network-services-arent-free-or-nonfree.html

gjhklfdsa:
@ADepic If I understand correctly the Bitchute website is non-free

ADepic:
Services cannot be non-free. Its a different issue to software...

Federated services like peertube are better, but bitchute is a good alternative suggestion because it actually has some content.

JonahAragon:
I think it's "private" as opposed to say, using YouTube. And I don't think it's true peer-to-peer, at least if it works the same way Mastodon works. The PeerTube instance is what downloads/caches the video, not your client, so the world would merely see that one specific instance downloaded a video, versus one specific user.

kewde:
The specific instance still sees that you've requested the video though, so instead of trusting YouTube you trust another (potentially unknown) individual?

Don't get me wrong, I like these type of alternatives, but I'm having a hard time seeing the tangible privacy benefits over using YouTube without an account and a VPN.

Mikaela:
> The PeerTube instance is what downloads/caches the video, not your client, so the world would merely see that one specific instance downloaded a video, versus one specific user.

I think you are wrong here and it is your browser which caches the video. Quoting from [PeerTube introduction](https://github.com/Chocobozzz/PeerTube#introduction):

> Video load is reduced thanks to P2P (BitTorrent) in the web browser via [WebTorrent](https://github.com/webtorrent/webtorrent/blob/master/README.md).

Webtorrent again says that Brave supports it directly.

* * * * *

I feel a bit conflicted between picking privacy and P2P and I think I lean a bit to the P2P side. Does decision have to be picked as if P2P is a problem, then Brave should be removed as they support WebTorrent and are working in integrating IPFS which we are also recommending?

JonahAragon:
Thanks @Mikaela, looks like you were all right originally.

---

> then Brave should be removed as they support WebTorrent

I disagree that this is an issue with Brave. Supporting a web protocol is different than implementing it as a key component of a service. WebTorrent *support* is okay, but if Brave were to say, implement some sort of shared cache between all their browsers with WebTorrent for performance then we'd have to reconsider that recommendation.

---

> and are working in integrating IPFS which we are also recommending?

Maybe. I don't know how I feel about this, because IPFS is largely a [public platform](https://medium.com/pinata/ipfs-privacy-711f4b72b2ea), but then the same could be said about PeerTube, or similar technologies like torrents in general. That almost seems like it should be a separate discussion.

---

Now we've reached the root of the problem with decentralized networks: they're great for taking control of your personal data, and they'll keep that data private *from large centralized companies* like Google/Facebook/etc, but they won't keep your *activity* private in the same way that Tor does for example, so they aren't good tools against mass surveillance. At this point we're running into issues like #880 and #848: Should we be recommending tools that will increase user's privacy *above the status quo* but won't necessarily do anything about government surveillance, or should we stick by our "protect your privacy against global mass surveillance" mantra and only recommend tools that will perfectly preserve your privacy online?

blacklight447-ptio:
@JonahAragon Btw, i recall that there was something with firefox esr that it only got critical and high security updates, leaving the medium and low. Don't pin me on this yet though, ill yry and see if i can find it again tonight, but its something to keep in mind.

asddsaz:
> I don't see how these tools are enhancing privacy?
> Decentralized torrenting like tools are pretty much leaking to the whole world that I'm watching a particular video.

**In response to [BitTorrent](https://www.wikipedia.org/wiki/BitTorrent) comment:**
The idea is that nobody know who you are. Especially if you follow the recommended security steps and use I2p/TOR. Sure, you must ask somebody for files but, you don't need to give any more personal info.

As to the others:
Peertube is mostly web based and am I unaware of any security bug or privacy threat in their software.
Fairly certain they don't follow you when you close the tab.

FreeTube stores data locally and supports TOR.

Invidio has an onion address: [http://axqzx4s6s54s32yentfqojs3x5i7faxza6xo3ehd4bzzsg2ii4fv2iid.onion/](http://axqzx4s6s54s32yentfqojs3x5i7faxza6xo3ehd4bzzsg2ii4fv2iid.onion/) ([source](https://github.com/omarroth/invidious/wiki/Invidious-Instances))
No login is required.

Mikaela:
> Especially if you follow the recommended security steps and use I2p/TOR.

I2P welcomes Bittorrent traffic, but by torrenting through Tor, you compromise your anonymity and slow down the network.

* [ Bittorrent over Tor isn't a good idea ](https://blog.torproject.org/bittorrent-over-tor-isnt-good-idea) on Tor blog.

asddsaz:
@Mikaela The reason I say this is because networks like ZeroNet use TOR.
I2p support has been WIP for years: https://github.com/HelloZeroNet/ZeroNet/pull/602

Web-based apps should work perfectly fine over TOR :)

asddsaz:
I'm gonna to rebase the PR and make LBRY "warning".

ggg27:
Just wanted to request the source code being added:

> LBRY: https://github.com/lbryio
> 
> PeerTube: https://github.com/Chocobozzz/PeerTube
> 
> FreeTube: https://github.com/FreeTubeApp/FreeTube
> 
> Invidious: https://github.com/omarroth/invidious
> 
> KopyKate (proprietary) : https://github.com/misses-robot/KopyKate-Big
> 
> DTube: https://github.com/dtube/dtube

PR #674, opened 2018-12-21
Add More Privacy-respecting Browsers

asddsaz:
**Description**: Add More Privacy-respecting Browsers
**Why**? Currently there are only 2 major open-source web browsers, Chromium and Firefox. Giving to much market share to a few companies can be harmful especially when there are plenty of alternatives.
**What software does this add**? The following have been added to "Worth Mentioning":

- [Cliqz](https://cliqz.com/)
- [Basilisk](https://basilisk-browser.org/)
- [NetSurf](http://www.netsurf-browser.org/)
- [Dooble](https://textbrowser.github.io/dooble//)
- [Otter Browser](https://otter-browser.org/)

All of these meet the [PrivacyTools.io Criteria](https://github.com/privacytoolsIO/privacytools.io/blob/master/.github/CONTRIBUTING.md#software-criteria).

### DIFF
--- a/index.html
+++ b/index.html
@@ -532,6 +532,16 @@ layout: default
     -->
 
   </div>
+  
+  
+  <h3>Worth Mentioning</h3>
+  <ul>
+    <li><a href="https://cliqz.com/">Cliqz</a> - Cliqz is a security-focused web browser.</li>
+    <li><a href="https://textbrowser.github.io/dooble//">Dooble</a> - Dooble is a next generation web browser.</li>
+    <li><a href="http://www.netsurf-browser.org/">NetSurf</a> - NetSurf is a web browser focusing on speed.</li>
+    <li><a href="https://basilisk-browser.org/">Basilisk</a> - Basilisk is an XUL-based web browser by the developers of <a href="https://www.palemoon.org/">Pale Moon</a>.</li>
+  </ul>
+
 
   <!-- Browser Fingerprint -->
   <h1 id="fingerprint" class="anchor"><a href="#fingerprint"><i class="fas fa-link anchor-icon"></i></a> Browser Fingerprint - Is your browser configuration unique?</h1>
--- a/index.html
+++ b/index.html
@@ -540,6 +540,7 @@ layout: default
     <li><a href="https://textbrowser.github.io/dooble//">Dooble</a> - Dooble is a next generation web browser.</li>
     <li><a href="http://www.netsurf-browser.org/">NetSurf</a> - NetSurf is a web browser focusing on speed.</li>
     <li><a href="https://basilisk-browser.org/">Basilisk</a> - Basilisk is an XUL-based web browser by the developers of <a href="https://www.palemoon.org/">Pale Moon</a>.</li>
+    <li><a href="https://otter-browser.org/">Otter Browser</a> - Otter Browser is a fully open source browser using the Qt Framework.</li>
   </ul>
 
 


Shifterovich:
Can you test how they perform on panopticlick?

beerisgood:
Privacy respecting isn't the same as Fingerprint resistent.
And non of them have a big userbase so the fingerprint is nearly unique

asddsaz:
> Can you test how they perform on panopticlick?
Fingerprinting is based on the number of users and how hard it would be to guess. It is generally a very hard to use metric. That being said, I will try this and come back and edit this post to include them.

> And "Cliqz". Remember the story with Ghostery add-on? Same company so this isn't privacy friendly
I actually don't, could you please refresh my memory? If there is a genuine security/privacy flaw then I would be willing to remove it from my pull request. 

@beerisgood @Shifterovich

kewde:
> Giving to much market share to a few companies can be harmful especially when there are plenty of alternatives.

I don't see how market share translates itself to decreased security.
I also don't see how alternatives that use the rendering engines _created_ by those companies that you claim are "too big" are a valid solution.

In fact I believe we should replace Brave with Chromium.

0ndrey:
> 
> 
> > Giving to much market share to a few companies can be harmful especially when there are plenty of alternatives.
> 
> I don't see how market share translates itself to decreased security.
> I also don't see how alternatives that use the rendering engines _created_ by those companies that you claim are "too big" are a valid solution.
> 
> In fact I believe we should replace Brave with Chromium.
Shifterovich refuses to remove Brave. They probably paid advertising

Shifterovich:
> In fact I believe we should replace Brave with Chromium.
> Shifterovich refuses to remove Brave. They probably paid advertising

You must be as stupid as @ciampolo.

https://github.com/privacytoolsIO/privacytools.io/issues/649#issuecomment-446340793

If I refused to remove Brave, I obviously wouldn't have had opened an issue to remove Brave and I wouldn't have even participated in that discussion (#649), I would just close it. It's really annoying when you spend a fuckton of time participating in meaningless discussions, rather than closing them after first sign of retardedness (such as the original post by @ciampolo) because @0ndrey and @ciampolo's brain capacity is limited to accusing people of nonsense. And after you put in so much time people still say retarded shit like that.

@kewde Yeah it's possible ungoogled chromium would score better on fingerprinting tests. Can you try that out? It seems like it's possible to detect Brave right now. If ungoogled chromium was (almost) indistinguishable from Chrome/ium, it might be a better choice.

kewde:
@Shifterovich 

I think there's a credible source for Brave detection available.
https://www.ctrl.blog/entry/brave-user-agent-detection

I had a lot of hopes for `muon`. I had a slight hope that it would beat some common sense into developers using electron but the project has been deprecated.

Shifterovich:
@diracdeltas are you aware of what the blog post @kewde sent mentions?

kewde:
Hmm, upon further inspection it seems like the problems might have been mitigated by the move to the pure chromium based browser.

I believe they deployed that on the 7th of December whilst the detection method was published a lot earlier. So this might have been behavior that was inherited from the electron codebase.

Shifterovich:
Still worth checking with Brave devs.

diracdeltas:
@Shifterovich @kewde https://www.ctrl.blog/entry/brave-user-agent-detection indeed refers to the old muon-based Brave.

> Brave may not have its own HTTP User-Agent request header, but even the order of Brave’s HTTP request headers are in themselves unique to Brave. 

I would be surprised if this worked on new Brave which is more similar to Chromium than Electon was.

> you can use the following JavaScript function to detect Brave

This or similar functions will work to distinguish Brave from Chromium. For instance we spoof 3rd party referer (by setting the 3rd party referer to the origin of the first party), which is something no other browser does (except maybe Tor browser when visiting .onions). If a site wanted to, it could abuse this feature to distinguish Brave.

Shifterovich:
> This or similar functions will work to distinguish Brave from Chromium. For instance we spoof 3rd party referer (by setting the 3rd party referer to the origin of the first party), which is something no other browser does (except maybe Tor browser when visiting .onions). If a site wanted to, it could abuse this feature to distinguish Brave.

Anti-fingerprinting features are a pain. By trying to mask one value ("3rd party referer"), you often end up creating another value ("spoofs 3rd party referer") that can be used for fingerprinting. Seems like it's a lot about finding the right balance, since masking some values does more harm than good.

0ndrey:
@Shifterovich 
Insulting other users does not make you taller or smarter than them. @ciampolo gave very well-reasoned comments, unlike you. You for no apparent reason refuse to delete this browser. The links that you provide are only an attempt to remove suspicions of your interest in Brave being present on the site.

Shifterovich:
> @ciampolo gave very well-reasoned comments, unlike you

@ciampolo:

> The only logical explanation I have is there is money involved.

Clearly mentally deficient.

> if I pay you however much @bbondy and/or Eich are paying you and you actually add my malware/adware to your site privacytools.io
> Admittingly if it is enough money I can understand OP's position
> And yes I didn't think Brave devs would be that dumb but if you look at the issues I opened months ago on their issue trackers you will see yes they actually are that retarded
> Mr. Eich, butthurt @bbondy 's boss that got fired off of Mozilla, uses Blink/Webkit. But Webkit can't do shit against zombie cookies

> You for no apparent reason refuse to delete this browser

Why did I open an issue to remove Brave then? Why was I against adding it in the first place? Please do your research for at least 5 minutes before accusing people.

asddsaz:
> > Giving to much market share to a few companies can be harmful especially when there are plenty of alternatives.
> 
> I don't see how market share translates itself to decreased security.
> I also don't see how alternatives that use the rendering engines _created_ by those companies that you claim are "too big" are a valid solution.
> 
> In fact I believe we should replace Brave with Chromium.

Way too many issue with Chromium.

Besides, the only decent open-source layout engine besides is [NetSurf](https://www.wikipedia.org/wiki/NetSurf) and my pull request will add [NetSurf](https://www.wikipedia.org/wiki/NetSurf). 
The layout engine market has become an [Oligarchy](https://www.wikipedia.org/wiki/Oligarchy).

kewde:
@ asddsaz

What security measures does the NetSurf redndering engine provide that would compel me to add them?

David-Beetle:
> @asddsaz
> 
> What security measures does the NetSurf rendering engine provide that would compel me to add them? I don't care about oligarchies in the rendering engines - in fact for fingerprinting privacy it is beneficial to have an oligarchy or monopoly.

I think you are missing the big picture. 
Besides, @asddsaz already did a fingerprinting test:
 
>     Within our dataset of several million visitors tested in the past 45 days, only one in 320170.44 browsers have the same fingerprint as yours.
> 
>     Currently, we estimate that your browser has a fingerprint that conveys 18.29 bits of identifying information.
> 

This is significantly better than what you would see if you tested this in Firefox without making any changes. Besides, it isn't like this PR is for adding it to main, it is just to add it to worth mentioning.



> we don't need more browsers which are just the same thing with a new name

To be fair, most of these browsers like Brave are made to make private browser easier and by default. This is something, the privacy community has historically supported. :/

kewde:
I've grown to become cautious of fingerprinting tests.
The data sets are often skewed, I don't know what data source most of them use but people doing fingerprinting test are often people who are already privacy conscious.

Also, I highly doubt the fingerprinting test above includes the user agent?
I don't know anyone who uses NetSurf and their [custom user agent](https://developers.whatismybrowser.com/useragents/explore/software_name/netsurf/) is rather uncommon. Is that reflected in the fingerprint tests? Are the fingerprinting techniques even optimized for the NetSurf browser and rendering engine?

I don't trust new browsers that roll their own rendering engines. I've seen so many jailbreaks based on exploits in the rendering engines that I've become very distrustful of any product that relies on it. Perhaps NetSurf is different, but I don't want to advocate for the next WebKit fiascos. 

You have a clear stance on monopolies and oligarchies. You don't like them.
But I don't see how more homogeneity caused by such market characteristics decrease privacy.
From the privacy perspective, everyone using the same browser is the best solution. A bigger set of people to mix in with. You could make the case that diversity of browsers is more resilient against a very specific security vulnerability, but that's a very thin argument as it only applies in on specific instance of a discovered exploit. More code, generally means more bugs. The only thing I currently like about NetSurf is that it doesn't even ship with JavaScript at all. But I haven't been able to find any documentation about the security mechanisms in the browser. Nonetheless, I love the project and I'm keeping tabs on it, glad to know that it exists and is being actively developed. With a big experimental warning, it might be worth mentioning but I wouldn't consider it more private or secure in any manner.

beerisgood:
@kewde fingerprint tests of course test your User-Agent. It even exist useragent-only tests

Shifterovich:
> I don't trust new browsers that roll their own rendering engines. I've seen so many jailbreaks based on exploits in the rendering engines that I've become very distrustful of any product that relies on it.

I agree. The argument against monopolies doesn't take into account that browser engines are probably the second most difficult thing to make securely, right after operating systems. Per se -- browsers literally execute arbitrary code on the client's device and have to manage privacy, sandboxing, etc properly. If Chrome & FF have security issues while being developed by a huge team, a browser with a small development team will be full of vulnerabilities assuming basic features like JS.

kewde:
@beerisgood 

It is very counter intuitive that the NetSurf produces a more private fingerprint than rendering engines based on Blink. Even based, merely on the user agents detection, intuitively it should stick out a lot more than the results of other browsers.
It raises the question whether fingerprinting tests are suitable means to make rank-ordering decisions. It's practically speaking the best solution we have, but is it therefore theoretically a "good" solution? 

I can only explain it with the following observations:
* The browsers analyzed by the fingerprinting tool is highly biased for privacy focused users. The amount of users who tested the NetSurf browser is likely higher than the real life usage (assumption though). Is the dataset used by the fingerprinting tools representative of the real world landscape?
* JavaScript is not supported, which obviously renders JS based fingerprinting techniques useless.

BurungHantu1605:
Be nice to each other, guys.

PR #675, opened 2018-12-21
Add Linux Privacy Tools Section

asddsaz:
**Description**: Add Linux Privacy Tool Section (similar to Android Privacy Addons).
**What software does this add**?
To Main:
1. [Firejail](https://firejail.wordpress.com/)
2. [Gufw](http://gufw.org/)

To Worth Mentioning:

- [Flatpak](https://www.flatpak.org/)


angela-d:
Why Flatpak?  What makes Flatpak privacy oriented?

hugoncosta:
Where would they go as "worth mentioning" though? We don't have any category dedicated to network security, maybe we should? For example, revamp the router category into something like "Network Security", which would cover both router and the device?

Shifterovich:
Network security -- routers & open source firewalls? Sounds good.

hugoncosta:
I think we should create a discussion topic on this, especially to discuss network security software for Windows and/or multiplatform, I have no idea if these exist.

Shifterovich:
True, Windows netsec will be difficult. Will you create a Discusion issue?

asddsaz:
> Why Flatpak? What makes Flatpak privacy oriented?

@angela-d Flatpak would be added to worth mentioning. It has very good built-in sand-boxing tools and is significantly easier to use that Firejail.

> Not sure if they need their own "major" section, something like worth mentioning would be enough imo.

@Shifterovich Wdym? Add to which worth mentioning?

> I think we should create a discussion topic on this, especially to discuss network security software for Windows and/or multiplatform, I have no idea if these exist.

@hugoncosta  Windows, overall is bad for privacy. I believe that this should be added to the "Windows 10" section worth mentioning. Linux has a lot of privacy tools that work with its kernel and therefore don't support Windows. Similar to the Android addon section.

> Will you create a Discusion issue?

If you do/have created one plz come back and reply with the URL. :)

Shifterovich:
> @Shifterovich Wdym? Add to which worth mentioning?

Right now let's stick with @hugoncosta's idea and make a network security section.

Though that still leaves us with sandboxing tools. Where should we put them @hugoncosta?

hugoncosta:
> > @Shifterovich Wdym? Add to which worth mentioning?
> 
> Right now let's stick with @hugoncosta's idea and make a network security section.
> 
> Though that still leaves us with sandboxing tools. Where should we put them @hugoncosta?

Was thinking about that, sandboxing is pretty important. What other software is there that helps with privacy besides sandboxes? If we find several, I guess it'd be best to create a catch-all category along the lines of Firefox Add-ons for the three major OSes.

asddsaz:
> What other software is there that helps with privacy besides sandboxes?

@hugoncosta Sandboxing goes a long way. Some software I did consider adding to the PR and can still is [TorSocks](https://github.com/dgoulet/torsocks/) and System wide add blockers. These again, seem to be mainly for Linux and generally are hard to use[1](https://github.com/privacytoolsIO/privacytools.io/blob/master/.github/CONTRIBUTING.md#software-criteria).

Shifterovich:
> System wide ad blockers. However, these seem to be mainly for Linux and generally are hard to use1.

Do we recommend pi hole?

asddsaz:
@Shifterovich Pi hole is generally hard to use/set-up and is more for creating a self-hosted DNS for blocking advertisements.

zero77:
@asddsaz 
There's also OpenSnitch a Linux application firewall.
https://github.com/evilsocket/opensnitch

BurungHantu1605:
hi guys, what's the status here?

PR #689, opened 2018-12-24
specify that riot use e2e not enable by default

Kcchouette:
## Description

specify that riot use e2e not enable by default


Vincevrp:
We currently have a similar warning for this.

PR #690, opened 2018-12-24
Reformulate Routers and add Linux Tools

hugoncosta:
## Description
Regarding #680 
Changed the name of routers to "Network Security" in order to be more encompassing.
Created a new section for Linux tools. Right now just has Gufw, but looking for suggestions on what to add besides it (sandboxing software, you name it).

## Screenshots
https://i.imgur.com/Sj9LjcB.png


asddsaz:
[Gufw](https://en.wikipedia.org/wiki/Uncomplicated_Firewall) is Linux only. Although, it is a great piece of software and would highly recommend it.

Mikaela:
I think gufw being a GUI for UFW should be mentioned as currently it seems to be presented as a standalone firewall.

I would recommend ufw directly whenever possible though, the syntax isn't very complicated and it ships with at least Ubuntu even if it's disabled by default.

hugoncosta:
Was unsure about adding the reference, but makes sense, I'll add it

Mikaela:
I think it only ships with Ubuntu, or at least not with Arch, Debian or Fedora.

Shifterovich:
Isn't UFW based on iptables? I think there should be some cross-distribution iptables GUI tool.

Mikaela:
Yes, UFW is an iptables frontend and does ip6tables simultaneously.

PR #691, opened 2018-12-24
Add Minds to Decentralized Social Networks

asddsaz:
**Description**: Add Minds to Decentralized Social Networks "Worth Mentioning" section.
**Minds URL**: https://www.minds.com/
**Wikipedia**: https://www.wikipedia.org/wiki/Minds
**Doesn't Minds support hate speech**? Although, Minds doesn't censor content it does enable users to flag content that may be found offensive like nsfw. You may then opt-out/filter these.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1681,6 +1681,7 @@ layout: default
   <h3>Worth Mentioning</h3>
   <ul>
     <li><a href="https://gnu.io/social/try/">GNU Social</a> - An open sourced, free and decentralized social network compatible with Mastodon and Friendica.</li>
+    <li><a href="https://www.minds.com/">Minds</a> - An open source usre-friendly distributed social network.</li>
   </ul>
 
   <h3>Related Information</h3>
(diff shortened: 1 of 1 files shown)


Shifterovich:
How exactly is it decentralized?

hugoncosta:
I believe the "decentralized" part they refer to is the use of an ethereum token for payments. Nonetheless, this looks like a STEEM copy to me, I'm unsure if it should be featured.

asddsaz:
> How exactly is it decentralized?

If I understand correctly they utilize a decentralized video content storage and have a self-host option (I haven't used it). They also have a decentralized reward system.

> I believe the "decentralized" part they refer to is the use of an ethereum token for payments. Nonetheless, this looks like a STEEM copy to me, I'm unsure if it should be featured.

STEEM is more of a blogging platform, I know this is a privacy community but have you ever used Facebook? It is very similar to that.

gjhklfdsa:
@danarel They have a user-led censorship system. You can feel free to mark posts as NSFW.

Shifterovich:
> Most Free Software social networks appear to get a lot of attention from censored groups like Neo-Nazis.

This will always happen, since they're the most censored group on most social media. I think Steemit is a bit better though? It's very good tech-wise and also community-wise afaik.

PR #696, opened 2018-12-27
Add files for Travis CI

Shifterovich:
## Description

Add .travis.yml and .github/ci_build.

Squash before merge.


.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: ruby
+rvm:
+  - 2.4.1
+
+before_script:
+ - chmod +x ./.github/ci_build # or do this locally and commit
+
+# Assume bundler is being used, therefore
+# the `install` step will run `bundle install` by default.
+script: ./.github/ci_build
```
Vincevrp: I suggest we move the script to the root of the repository. Also please add a `.sh` extension.

.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: ruby
+rvm:
+  - 2.4.1
+
+before_script:
+ - chmod +x ./.github/ci_build # or do this locally and commit
```
Vincevrp: Is this required? I think github saves file permissions so you can just make it executable and commit it like that.

PR #704, opened 2018-12-29
Actualized VPN table

c0rdis:
## Description
The VPN table was quite outdated - reviewed and fixed number of servers, price and in case of Perfect Privacy - country (from Panama to Switzerland, see https://www.perfect-privacy.com/imprint). Also smaller fixes for better sorting of the table.

### DIFF
--- a/index.html
+++ b/index.html
@@ -211,7 +211,7 @@ layout: default
         <tr>
           <td data-value="AirVPN">
             <a href="https://airvpn.org/"><img alt="AirVPN" src="assets/img/provider/AirVPN.png" width="200" height="70"></a></td>
-          <td data-value="70">54 €</td>
+          <td data-value="54">54 €</td>
           <td><span class="label label-success">Yes</span></td>
           <td>162</td>
           <td><span class="flag-icon flag-icon-it"></span> Italy</td>
@@ -222,9 +222,9 @@ layout: default
           <td data-value="AzireVPN">
             <a href="https://www.azirevpn.com/"><img alt="AzireVPN" src="assets/img/provider/AzireVPN.png" width="200" height="70"></a>
           </td>
-          <td data-value="58">45 €</td>
+          <td data-value="45">45 €</td>
           <td><span class="label label-success">Yes</span></td>
-          <td>5</td>
+          <td>22</td>
           <td><span class="flag-icon flag-icon-se"></span> Sweden</td>
           <td><a href="https://www.azirevpn.com/">AzireVPN.com</a></td>
         </tr>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -211,7 +211,7 @@ <h3 id="usa" class="anchor">Why is it not recommended to choose a US-based servi
         <tr>
           <td data-value="AirVPN">
             <a href="https://airvpn.org/"><img alt="AirVPN" src="assets/img/provider/AirVPN.png" width="200" height="70"></a></td>
-          <td data-value="70">54 €</td>
+          <td data-value="54">54 €</td>
```
Shifterovich: Here you use EUR in data-value

index.html
```diff
@@ -255,9 +255,9 @@ <h3 id="usa" class="anchor">Why is it not recommended to choose a US-based servi
           <td data-value="ExpressVPN">
             <a href="https://www.expressvpn.com/"><img alt="ExpressVPN" src="assets/img/provider/ExpressVPN.png" width="200" height="70"></a>
           </td>
-          <td data-value="100">$ 99.95</td> <!-- USD on March 21, 2017 -->
+          <td data-value="99">$ 99.95</td> <!-- USD on December 28, 2018 -->
```
Shifterovich: And here you use USD in data-value.

index.html
```diff
@@ -266,7 +266,7 @@ <h3 id="usa" class="anchor">Why is it not recommended to choose a US-based servi
           <td data-value="FrootVPN">
             <a href="https://www.frootvpn.com/"><img alt="FrootVPN" src="assets/img/provider/FrootVPN.png" width="200" height="70"></a>
           </td>
-          <td data-value="36">$ 35.88</td>
+          <td data-value="35">$ 35.88</td>
```
Shifterovich: This should be rounded to 36.

index.html
```diff
@@ -277,9 +277,9 @@ <h3 id="usa" class="anchor">Why is it not recommended to choose a US-based servi
           <td data-value="hide.me">
             <a href="https://hide.me/"><img alt="hide.me" src="assets/img/provider/hide.me.png" width="200" height="70"></a>
           </td>
-          <td data-value="65">$ 65.04</td>
+          <td data-value="0">Free</td>
```
Shifterovich: Maybe this should mention that there is a 2 GB limit?

c0rdis: Not sure - if we do that, we should probably also mention limitations of other cheapest/free options - e.g. that of ProtonVPN

index.html
c0rdis: True - although it's visually much easier to get the data sorted according to numbers. Also, we wouldn't be dependent on exchange rates (e.g. now 54 € ~ $62)

index.html
c0rdis: See comment above

index.html
Shifterovich: If it's possible to do that inline while keeping the table nice, sure. Otherwise, maybe add a \* footnote to these two?

c0rdis:
Looks like I've got another commit as part of this PR - related to https://github.com/privacytoolsIO/privacytools.io/pull/621#issuecomment-450481258. Please review it as well.

Shifterovich:
I'll extract it into a separate branch. One change per PR.

index.html
c0rdis: For now, manually updated to reflect USD values, so resolving

index.html
c0rdis: For now, manually updated to reflect USD values, so resolving

c0rdis:
Sure, sorry

PR #706, opened 2018-12-30
Proofing

noplanman:
- Typo and grammar fixes.
- Opt for American English (which seems to be the chosen English)
- Fix some broken links


Shifterovich:
Thanks!

@Vincevrp Yeah we can mention that. We should also write a contributing _guide_ so that newbies can have an easy time contributing, with the recent changes.

PR #714, opened 2019-01-03
Change Etherpad URL to Etherpad.net

gjhklfdsa:
**Description**: Change Etherpad URL to [Etherpad.net](https://etherpad.net)
**Why**? Ease of use, Etherpad.net is plug-n-play while Etherpad is hard to set up.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1811,7 +1811,7 @@ layout: default
     {% include card.html color="success"
     title="Etherpad"
     image="assets/img/tools/Etherpad.png"
-    url="http://etherpad.org/"
+    url="http://etherpad.net/"
     footer="OS: Windows, macOS, Linux."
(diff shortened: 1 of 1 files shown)


Shifterovich:
> Ease of use, Etherpad.net is plug-n-play while Etherpad is hard to set up.

We should recommend both then. Etherpad as the main recommendation and a mention to Etherpad.net in the description perhaps?

gjhklfdsa:
> > Ease of use, Etherpad.net is plug-n-play while Etherpad is hard to set up.
> 
> We should recommend both then. Etherpad as the main recommendation and a mention to Etherpad.net in the description perhaps?

Ok, done.

gjhklfdsa:
> > Ease of use, Etherpad.net is plug-n-play while Etherpad is hard to set up.
> 
> We should recommend both then. Etherpad as the main recommendation and a mention to Etherpad.net in the description perhaps?

@Shifterovich Ok, done.

Kcchouette:
be careful: it's not "change" but "add"
And moreover, we should instead propose https://github.com/ether/etherpad-lite/wiki/Sites-that-run-Etherpad as if you want to find a third-party instance instead of force-proposed etherpad.net who seems to not be an official instance of developpers

Issue #375, opened 2017-12-17
Proposal: Waterfox, IceCat, Pale Moon & Seamonkey / Firefox forks


angela-d:
For those of us displeased with the direction Mozilla is headed, [Waterfox](https://github.com/MrAlex94/Waterfox) and [IceCat](https://www.fsf.org/working-together/gang/icecat) are suitable alternatives.
Are either or both of these browsers contenders for inclusion or mentions on Privacy Tools?

Pros of both:
- Firefox forks
- Telemetry removed
- Privacy tweaks pre-configured (in comparison to vanilla Firefox)
- Open source

Cons:
- Smaller userbase
- Less developers involved in both projects

Although Waterfox is currently just a "fork," it appears it's going to branch off & become a stand-alone project after the demise of ESR; from 56 release notes:

> Waterfox will now remain at 56 for the time being, following the security releases of 59 ESR until it becomes End of Line (Q1 2019). In the meanwhile, a “new” browser will be developed to follow the ethos of Waterfox of customisation and choice, while staying up to do date with the rapidly evolving browser landscape.

It has been mentioned that compatibility with "legacy" extensions is also in work to be retained beyond ESR, as well.

Edit: 
I forgot about Pale Moon, per @beerisgood's suggestion: https://www.palemoon.org/

Also worth mentioning: https://www.seamonkey-project.org/ - Though I'm not sure how intertwined Seamonkey is with Mozilla:
> Under the hood, SeaMonkey uses much of the same Mozilla source code which powers such successful siblings as Firefox and Thunderbird. Legal backing is provided by the Mozilla Foundation. 

Which is concerning, given Mozilla's recent activities; definitely not a company that should be handing out advice.

davidtabernerom:
Hi, could you share what do you mean by "the direction Mozilla is headed"? Is there anything new we don't know?
Thank you.

Atavic:
@davidtabernerom Shields project is meant to look at different user's settings and how they work on different sites. [Here](https://addons.mozilla.org/en-US/firefox/shield_study_16) is the last one.

The only useful part was [this one](https://github.com/mozilla/shield-study-privacy/) imho

Shifterovich:
> the direction Mozilla is headed

[...]

PR #715, opened 2019-01-04
Add Pale Moon

gjhklfdsa:
## Description
Add Pale Moon to browser "Worth Mentioning"

## Software Criteria
**Free Software**: Yes, source: https://github.com/MoonchildProductions/UXP/issues/923
**Cross-platform**: Yes, official supports Windows and Linux with community builds for various other OSes.
**Ease of use**: From personal experiences, yes. Very easy to use.
**Privacy respecting**: Yep, I would argue better than vanilla Firefox but, this is a low bar.
**Different**: Technically not in the criteria, but many seem to care about. Pale Moon isn't like Ice Cat or Waterfox which are basically just old-Firefox rebranded. Pale Moon utilizes its own layout engine [Goanna](https://www.wikipedia.org/wiki/Goanna_(software)) with many differences to gecko like single-process mode for security. If you look around you will find a thriving community supporting privacy.

### DIFF
--- a/index.html
+++ b/index.html
@@ -516,6 +516,12 @@ layout: default
     footer="OS: Windows, macOS, Linux, Android, iOS."
     description="The new open source browser \"Brave\" automatically blocks ads and trackers, making it faster and safer than your current browser. Brave is based on Chromium."
     %}
+    
+<h3>Worth Mentioning</h3>
+  <ul>
+    <li><a href="https://www.palemoon.org/">Pale Moon</a> - Pale Moon is a free and open source web browser utilizing the Goanna Browser Engine.</li>
+  </ul>
+
 
     <!-- IceCat removed
       <div class="col-sm-4">


beerisgood:
No. Pale Moon make you more unique then Firefox.
Also you don't have any improvement against Firefox but opposite like unsigned add-ons, no Webextension add-ons and more.

And yeah, a "own" engine from 1? 2? Guys. Not very secure

gjhklfdsa:
> No. Pale Moon make you more unique then Firefox.

So does Brave and TOR. If I understand correctly, it is still similar enough to evade much fingerprinting.

>  Also you don't have any improvement against Firefox but opposite like unsigned add-ons, no Webextension add-ons and more.

What are you even trying to say? This is all incorrect. Pale Moon  has many extensions[1](https://addons.palemoon.org/extensions/) and you can sign them.

beerisgood:
> So does Brave and TOR. If I understand correctly, Pale Moon is still similar enough to evade much fingerprinting.

Brave yes, Tor no. All Tor browser use (if not modified any stuff) the same signature which provide the only good solution from all browsers. Firefox have some Tor patches but not all features


> What are you even trying to say? This is all incorrect. Pale Moon has many extensions[1](https://addons.palemoon.org/extensions/) and you can sign them. Even for the extensions that might not be signed, you still utilize HTTPS.

I try to say that Pale Moon isn't better choice for privacy AND for security.
Also i don't care that Pale Moon have addons as they are based on old and unsecure Not-WebExtension-base.
And HTTPS have nothing to do with addons.

gjhklfdsa:
> Brave yes, Tor no. All Tor browser use (if not modified any stuff) the same signature which provide the only good solution from all browsers. Firefox have some Tor patches but not all features

TOR, technically is easier to distinguish because of IP (tor exit nodes are public) and extensions + blocking stuff like WebRTC. Not that this doesn't make these improvements.
The TOR team does and continues to evaluate every change with the implications to fingerprinting.

> I try to say that Pale Moon isn't better choice for privacy AND for security.

I'm not saying it is, this is the reason we have a "Worth Mentioning" section.
Generally speaking, these are very different browsers.

> Also i don't care that Pale Moon have addons as they are based on old and unsecure Not-WebExtension-base.

This isn't true. You can find many of the recommended addons like [Decentraleyes](https://addons.palemoon.org/addon/decentraleyes/) and [uBlock Origin](https://github.com/gorhill/uBlock#firefox--firefox-for-android). 

You have to remember, much of Pale Moon isn't from Firefox. They are there separate organization.

beerisgood:
@gjhklfdsa yeah, but old versions of add-ons.
I leave this here. You better read more about crap like Pale Moon and then come back.

gjhklfdsa:
> @gjhklfdsa yeah, but old versions of add-ons.

Not true, these should be actively maintained addons. 

> I leave this here. You better read more about crap like Pale Moon and then come back.

I would love to know the source of your information, perhaps you could point me in the direction?

I would also like to recommend reading this forum post: https://forum.palemoon.org/viewtopic.php?f=4&t=7818

I would also like to point out that to make a good browser, "tons" of extensions is not a requirement. Brave has few with direct support, and TOR recommend you only install a few, as many will harm your privacy. Pale Moon focuses where it counts. ;)

Kcchouette:
Related to #375 
I don't really like that people create a (somewhat forced) PR when discussion is open @Shifterovich :/

PR #717, opened 2019-01-08
Added Vivaldi Browser in place of Brave

danarel:
## Description

I added Vivaldi Browser in place of Brave. If the sites going to feature a Chromium based browser, Vivaldi checks far more boxes than brave, and without trying to sell you Cryptocurrency along the way. Much more control over privacy settings, extensions, and the browser itself is really customizable.


Hillside502:
Vivaldi browser is Proprietary Freeware:-  
https://vivaldi.com/privacy/vivaldi-end-user-license-agreement/

danarel:
I am totally okay with this being denied, but I would say even though not open source, it's still more secure than Brave. I think having a Chromium based browser could be a good thing for the site as it seems a lot of people do prefer it. Giving them a privacy focused version would be key here.

beerisgood:
> 
> 
> I think having a Chromium based browser could be a good thing for the site as it seems a lot of people do prefer it. Giving them a privacy focused version would be key here.

All Chromium and Chrome based browser are bad for privacy.
As for security, Firefox is good too and best for privacy

KosmoEskudo:
Please reopen https://github.com/privacytools/privacytools.io/issues/2206 (more recent) or https://github.com/privacytools/privacytools.io/pull/717 , since no one has said a *why* not to include it. I have researched Vivaldi, and have written up my findings in [this article](https://write.privacytools.io/kolivro/vivaldi-and-privacy) and I don't see a *why* not to include it.

The only thing that has been said is that it is closed source, and this issue is explained by Vivaldi [here](https://vivaldi.com/blog/vivaldi-browser-open-source/).

Translated with www.DeepL.com/Translator

Issue #719, opened 2019-01-14
💬 Discussion | Bitwarden has Google Analytics enabled by default. 


AshTex:
Bitwarden is currently first choice in password management software.

Generally I think Bitwarden is a good piece of software, however they have Google Analytics baked in and turned on by default. You can turn it off [1] but some users may not realise it's there to begin with as it is not well advertised. I think the recommendation for Bitwarden on the site should have a clear disclaimer about disabling Google Analytics, or it should be withdrawn completely. 

1 - https://github.com/bitwarden/browser/issues/27#issue-182945551

beerisgood:
For privacy this is a no-go.
Remove that crap. KeePass ftw

kspearrin:
@AshTex, as an update, I've removed analytics from all Bitwarden client applications for their next release:

Ref: https://github.com/bitwarden/browser/issues/27#issuecomment-454453339

AshTex:
@kspearrin Oh that's great news! 

As an aside - thanks for making such a brilliant product. I'm a paying user :)

PR #720, opened 2019-01-14
Removed Bitwarden

AshTex:
As talked about in issue #719 Bitwarden has Google Analytics baked in and enabled by default. This commit removes Bitwarden from the recommended programs and relegates it to last place in the "worth mentioning" list, with the warning about disabling Google Analytics if it is used.

## Description

Resolves: #719


Shifterovich:
Thanks.

Issue #737, opened 2019-01-23
🌐 Website Issue | Old code about Laverna in netbooks section


victorhck:
## Description
Remove old code about laverna in netbooks section.

I wanted to fix the cod by myself, but I get an error when I pushed my fix. 
this error:

> remote: Resolving deltas: 100% (4/4), completed with 4 local objects.
> remote: error: GH006: Protected branch update failed for refs/heads/master.
> remote: error: Required status check "Travis CI - Pull Request" is expected. At least 1 approving review is required by reviewers with write access.
> To https://github.com/privacytoolsIO/privacytools.io.git
>  ! [remote rejected] master -> master (protected branch hook declined)

Greetings

PS: Seems that I don't have access to master branch. I can't either fix it via GitHub editor.

PR #738, opened 2019-01-23
Remove old code in notebooks section

victorhck:
removed old text about laverna

## Description

Resolves: #737


victorhck:
done. :)

PR #744, opened 2019-01-26
Replacing Youtube (GoogleTube) with Peertube.

pdjpdjpdj:
## Description

Resolves: #579


Kcchouette:
Why do you choose mastodon.host as an instance to peertube? It's an instance not even listed on the peertube website list.

I think we should use the partner instance https://framatube.org/

PR #755, opened 2019-02-09
Moved i2p-bote to worth mentioning due to lack of development

beardog108:
We should move i2p-bote to worth mentioning, due to the website not having a working build available and due to the fact that its development is effectively stalled.

It is also somewhat hard to use due to it only working with I2P.

We could replace it with ZeroMail: https://github.com/HelloZeroNet/ZeroMail although ZeroMail doesn't get much development either anymore.

### DIFF
--- a/_includes/sections/email-alternatives.html
+++ b/_includes/sections/email-alternatives.html
@@ -3,22 +3,13 @@
 <div class="row mb-2">
 
   {% include card.html color="success"
-  title="I2P-Bote"
-  image="assets/img/tools/I2P.png"
-  url="https://i2pbote.xyz/"
-  footer="OS: Windows, macOS, Linux, Android, BSD / Solaris."
-  description="I2P-Bote is a fully decentralized and distributed email system. It supports different identities and does not expose email headers. Currently (2015), it is still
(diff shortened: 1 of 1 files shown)


_includes/sections/email-alternatives.html
```diff
@@ -3,22 +3,13 @@ <h1 id="messaging" class="anchor"><a href="#messaging"><i class="fas fa-link anc
...
-
-  {% include card.html color="primary"
   title="Bitmessage"
   image="assets/img/tools/Bitmessage.png"
   url="https://bitmessage.org/"
   footer="OS: Windows, macOS, Linux."
   description="Bitmessage is a P2P communications protocol used to send encrypted messages to another person or to many subscribers. It is decentralized and trustless,
   meaning that you need-not inherently trust any entities like root certificate authorities. It uses strong authentication which means that the sender of a message cannot be spoofed, and it aims to hide \"non-content\" data."
-  %}    {% include card.html color="warning"
+  %}    {% include card.html color="primary"
```
Vincevrp: Could you start the next include on a new line, please?

beardog108: done

Issue #784, opened 2019-03-20
Add extensions.pocket.enabled false to the firefox about:config tweak list


FreddyMarsden:
Pocket is a part of PRISM

yssmcl:
Source?

Mikaela:
I cannot find a source for it being part of PRISM, but I am happy to get it removed from my browser experience regardless of not remembering why I originally didn't like it. I can only find a source for this option.

* [Firefox Help: Disable or re-enable Pocket for Firefox](https://support.mozilla.org/en-US/kb/disable-or-re-enable-pocket-for-firefox)

beerisgood:
Also gHacks user.js disable it too

PR #786, opened 2019-03-25
Adds extensions.pocket.enabled false

cyaniventer:
## Description

Resolves: #784

### DIFF
--- a/_includes/sections/browser-tweaks.html
+++ b/_includes/sections/browser-tweaks.html
@@ -125,6 +125,9 @@
   </ul>
   </dd>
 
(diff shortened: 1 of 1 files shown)


_includes/sections/browser-tweaks.html
```diff
@@ -125,6 +125,9 @@ <h3>Getting started:</h3>
   </ul>
   </dd>
 
+  <dt>extensions.pocket.enabled = false</dt>
+  <dd>Disables pocket completely.</dd>
```
JonahAragon:
```suggestion
  <dd>Disables Pocket completely.</dd>
```
And maybe add why it should be disabled?

Issue #374, opened 2017-12-17
What do you think about Cloudflare?


ghost:
This is a xreference from prism-break, the similar website endorsing privacy-focused software.

https://github.com/mozilla-mobile/focus-android/issues/1743

I think your website need to mention Cloudflare under "Recommended Privacy Resources" - "Information".

Atavic:
A big company that works with millions of sites, potentially tracking all their users.

davidtabernerom:
Didn't it had a security breach a few weeks ago?
I remember that, of course, they didn't say a word.

Hillside502:
https://en.wikipedia.org/wiki/Cloudflare#Criticism_and_controversies

privacyisdeadinusa:
Fun fact: www.privacytools.io is using Cloudflare.

Hillside502:
https://iplookup.flagfox.net/?ip=104.31.90.13&host=www.privacytools.io
https://www.shodan.io/host/104.31.90.13

Shifterovich:
#96

ohmynameisrico:
Now the company websites are forced to write GDPR compatible privacy policy, what makes me laugh is they - who use Cloudflare to serve websites - are forgetting about Cloudflare MITM thing.

CHEF-KOCH:
Another problem with GDPR is the fact that every website which notifies you (the little EU cookie policy warning on the top or bottom of the page) places now a cookie (first-party) no matter if you accept the terms or not. It not only affects cloudflare hosted sites but it's really annoying because these cookies also have special 'GDPR' names or tags which are then even more traceable in case someone likes to scan your website/user explicitly for GDPR related things ... for eg. research, targeting etc.

ohmynameisrico:
@CHEF-KOCH

"places now a cookie"? Really? I didn't noticed it... Oh ok, I always browse website without cookies anyway. (deny all)

libBletchley:
I have a few issues with CloudFlare:

# Problem with CloudFlare
CloudFlare is a vigilante extremist organization who takes the decentralized web and centralizes it under one corporate power that controls the worlds largest walled-garden. A very large portion of the web (10%+) that was once freely open to all is now controlled and monitored by one central authority who decides for everyone who can see what web content. This does serious damage to net neutrality, privacy, and has immediate serious consequences:

[...]

Issue #792, opened 2019-03-30
🌐 Website Issue | HSTS Headers


JonahAragon:
## Description:

We should enable HSTS headers on `https://www.privacytools.io/`, as we already redirect to HTTPS anyways. Is that something all our subdomains are capable of?

Mikaela:
Related https://github.com/privacytoolsIO/privacytools.io/issues/731 (DNSSEC) which is also option in Cloudflare config.

JonahAragon:
Implemented.

Issue #151, opened 2017-01-04
privacytools.io site fails on mozilla obervatory


p43b1:
The site privacytools.io gets an awful score on mozilla oberservatory

https://observatory.mozilla.org/analyze.html?host=www.privacytools.io

Shifterovich:
Thanks. All the points we lost are because of headers. Not sure whether this can be improved, as we use CloudFlare. @privacytoolsIO can you add headers with CF?

Vincevrp:
The website also has a bad score on [gtmetrix](https://gtmetrix.com/). Prehaps @privacytoolsIO can optimize some of these issues?

Shifterovich:
Doesn't GH support HTTPS without CF now?

Vincevrp:
> Doesn't GH support HTTPS without CF now?

It does but Github cache settings aren't optimal, I think you could better optimize it with CF. Though if CF is seen as a privacy issue, we'd be better of without it.

nikhiljha:
GitHub's server also scores an F on observatory for similar reasons.

JonahAragon:
>The website also has a bad score on gtmetrix.

Here you go.

![image](https://user-images.githubusercontent.com/3637842/55269515-518d5400-5262-11e9-9c36-038ec2030b42.png)

We also [score fine](https://globalsign.ssllabs.com/analyze.html?d=www.privacytools.io&s=104.28.22.220&latest) on SSL Labs. I don't *think* Mozilla Observatory is a big deal since we redirect to HTTPS, nor something that we can fix while on GitHub Pages.

We could probably consider implementing HSTS though.

JonahAragon:
Fixed, we now score an A on Mozilla Observatory and SSL Labs.

PR #794, opened 2019-03-30
Prep for move off CloudFlare

JonahAragon:
We are going to move off of CloudFlare to an Nginx server we control that will reverse proxy to the GitHub Pages host. Essentially we will be acting in the exact same way CloudFlare currently does for us, but under our control.

We need to start by serving this site on `privacytoolsio.github.io/privacytools.io` instead of `www.privacytools.io` so we can actually implement this server side (which is what this commit accomplishes). Once we move to the new DNS infrastructure we will point the main domain to our Nginx server and get Let's Encrypt installed. Hopefully this will be accomplished with minimal downtime.

Obviously, don't merge this until the backend is ready.

---

Resolves: #374

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,1 +1,0 @@
-www.privacytools.io


Shifterovich:
Why not just let GH take care of everything?

JonahAragon:
@Shifterovich that’s something I thought about, but this is my thinking...

1. If we just let GH handle it, GitHub will have access to all their access logs of course and will be able to see Visitor IPs. If we proxy it, GH will only be able to see our IP. And
2. If we proxy it, we can enable all the custom headers we want, which will help with #792, #151, etc. We can also do better browser caching for attachments. 

Thoughts?

Shifterovich:
If that's something @BurungHantu1605 can set up and maintain, sure, I see the benefits.

Using just GH is the easiest solution. For me personally, the convenience outweighs the few drawbacks (like custom headers and GH having log access).

JonahAragon:
@BurungHantu1605 and I are setting up servers for stuff like [Searx](https://search.privacytools.io/) and [Mastodon](https://social.privacytools.io/) and DNS so this is just included in all of that work.

JonahAragon:
This won't be necessary for the new plan :)

PR #797, opened 2019-03-31
Added Psono

chickahoona:
Added Psono as password manager


BurungHantu1605:
Thanks. And you're funny.

PR #820, opened 2019-04-03
Adds a classic version of the site.

JonahAragon:


BurungHantu1605:
hi jonah, great. some people were complaining on reddit.

can we use 
https://www.privacytools.io/classic/
instead of?
https://www.privacytools.io/old/

Issue #842, opened 2019-04-09
🌐 Website Issue | Remove legacy anchors


subsign:
An example can be found under the [Contacts section](https://www.privacytools.io/software/calendar-contacts/) when clicking on both Email Provider links.

See: https://github.com/privacytoolsIO/privacytools.io/blob/48fd518cb781944bb1472bdc13daf14f88d94423/_includes/sections/calendar-contacts-sync.html#L17-L24

PR #855, opened 2019-04-12
Fixed legacy anchors

LordShedy:
## Description
I attempted to fix the broken legacy anchors. 
Resolves: #842


BurungHantu1605:
thx

PR #861, opened 2019-04-13
Add Chameleon to browser addons section

harryadel:
## Description

A user-agent spoofing addon is essential to evade browser finger-printing.


beerisgood:
> 
> ## Description
> 
> A user-agent spoofing addon is essential to evade browser finger-printing.

Totally wrong. Faking useranget makes you even more unique cause you can't fully fake it and the real one can be read with JavaScript.
Better use RFP (RESIST FINGERPRINTING)

harryadel:
@beerisgood I didn't know that. I guess it's time for a bit of research. Thanks you!

PR #869, opened 2019-04-16
"Avoid US & UK services" in the header now links to the proper section

CommenterOfComments:
really small change but it is absolutely required and necessary the world will end if this isn't included trust me


BurungHantu1605:
thanks for saving the world, guys.

PR #825, opened 2019-04-03
Remove national flags from the language select.(Solution 2)

troutperch:
## Description

Resolves: https://github.com/privacytoolsIO/privacytools.io/issues/185

PR #873, opened 2019-04-18
Enable flags again

Strappazzon:
#825 removed national flags representing languages, but this also removed flags representing **countries** (see screenshot below). I added the flags again but without representing languages.


assets/img/flags/.move svg files in the 4x3 folder if you want to use them
```diff
@@ -0,0 +1 @@
+
```
Mikaela: I think this is a bit of nasty filename, could the information be in README.md of that directory instead?

JonahAragon: Agree, this file doesn't appear necessary.

Strappazzon:
@JonahAragon The `4x3` folder was there before the removal.
I removed `4x3`, `.move svg files` and edited `flag-icon.scss` accordingly in https://github.com/privacytoolsIO/privacytools.io/pull/873/commits/8be2c18416d29bb7796151dd8ee52436c059bfa1.

Mikaela:
> This seems fine to me

I am taking this as OK to merge as I don't see anything strange anymore either.

Thank you for your contribution @Strappazzon :)

PR #858, opened 2019-04-12
Add French to Community Translations

Strappazzon:
I added the link to privacytools-fr and its source code.

PR #881, opened 2019-04-24
browser-recommendation: suggest Firefox ESR instead of Firefox

Mikaela:
I wonder if this could be a compromise suggestion to #856 as with ESR the potentially privacy invasive updates also come more slowly and may in some cases be cancelled before they reach the ESR branch.

Problem: I don't think ESR exists on mobile Android and iOS.

Solution: Having different browser recommendations for those platforms such as Tor Browser (Alpha), Firefox Klar or DuckDuckGo (give me issue number please) and Brave?

### DIFF
--- a/_includes/sections/browser-recommendation.html
+++ b/_includes/sections/browser-recommendation.html
@@ -12,12 +12,13 @@
   %}
 
   {% include card.html color="primary"
-  title="Mozilla Firefox"
+  title="Mozilla Firefox ESR"
   image="/assets/img/tools/Firefox.png"
-  url="https://www.firefox.com/"
+  website="mozilla.org"
+  url="https://www.mozilla.org/en-US/firefox/organizations/all/"
   footer="OS: Windows, macOS, Linux, Android, iOS, BSD."
   description='Firefox is fast, reliable, open source and respects your privacy. Don\'t forget to adjust the settings according to our
-  recommendations: <a href="#webrtc"><i class="fas fa-link"></i> WebRTC</a> and <a href="#about_config"><i class="fas fa-link"></i> about:config</a> and get the <a href="#addons"><i class="fas fa-link"></i> privacy add-ons</a>.'
+  recommendations: <a href="#webrtc"><i class="fas fa-link"></i> WebRTC</a> and <a href="#about_config"><i class="fas fa-link"></i> about:config</a> and get the <a href="#addons"><i class="fas fa-link"></i> privacy add-ons</a>. The extended support release (ESR) gets new features more slowly, while still receiving backported security updates promptly.'
   %}
 
   {% include card.html color="warning"


_includes/sections/browser-recommendation.html
```diff
@@ -12,12 +12,12 @@ <h1 id="browser" class="anchor"><a href="#browser"><i class="fas fa-link anchor-
   %}
 
   {% include card.html color="primary"
-  title="Mozilla Firefox"
+  title="Mozilla Firefox ESR"
   image="/assets/img/tools/Firefox.png"
-  url="https://www.firefox.com/"
+  url="https://www.mozilla.org/en-US/firefox/organizations/all/"
```
Mikaela: I will need help on how to format URL if this PR is judged acceptable.

Screenshot from 2019-04-24 22-48-44

Shifterovich: Add a `website="mozilla.org` parameter. I set it up so that when `website` is missing, it's created from `url` by removing things like slashes, `https:` etc.

Atavic:
Mozilla info:

> Is Mozilla Firefox ESR available for Android and iOS? No. Firefox ESR will only be offered for Windows, macOS and Linux for desktop computers.

five-c-d:
2 grammar suggestions, and one clarity-suggestion:  

*  **Current:**  ...gets new features slower while still receiving security updates timely.
*  **Changes:**  ...gets new features <s>slower</s> <ins>more slowly,</ins> while still receiving <ins>backported</ins> security updates <s>timely</s> <ins>promptly</ins>.
*  **Suggest:**  ...gets new features more slowly, while still receiving backported security updates promptly.

Could also say "in a timely fashion" but I think "promptly" is clear enough.  

p.s.  I am not sure I agree with the **idea** motivating this pull-request, see my complaints in the https://github.com/privacytoolsIO/privacytools.io/issues/856#issuecomment-486396539 thread-comments.  Grammar suggestions should not be interpreted as, me thinking this is ready to push live :-)  They are just grammar suggestions  ;-)

five-c-d:
> kinda strange

The wording makes sense, just, our American ears require more of a midwestern-diaspora dialect, or something :-)  Grammar is not that important in the blurbs, people care about the recommendations themselves more than the verbiage, although, might as well tweak it I guess.  

<details><summary>real reason to comment was the 'backport' thing, I think it matters</summary><p>

The reason I even brought up grammar-fixes at all, is I think the 'backporting' word is the only thing that needs clarification:  the security-fixes in firefoxESR are promptly **backported** from the stable-rolling-release (or from the alpha-channel-rolling-release more likely), is a very important distinction.  FirefoxESR does not add new features (like Waterfox and to some extent TorBrowser and very much so PaleMoon), so the backports are extra-clean extra-well-vetted.  Lots of eyeballs have looked over each security-fix-backport, because generally speaking if there is a security-hole in firefoxESR there is **also** going to be a security-hole in firefox-rolling-release-stable, and for the most part vice versa, the exception being super-new features only found in rolling-release.  

The same cannot be said of the soft-forks which are more than very-light-ESR-type-soft-forks, because they *add features*.  Prompt security-patches to the Waterfox-specific portions of waterfox might occur, but there are a lot fewer eyeballs on those Waterfox-specific portions.  There ARE not any firefoxESR-specific portions, that I'm aware of, because it is just a frozen-in-time v60 flavour of firefox-rolling-release, plus security-fix backports.  

</p></details> 

>  Problem: ...Android and iOS.  Solution: ...[recommend instead] Tor Browser (Alpha), Firefox Klar or DuckDuckGo (give me issue number please) and Brave?

This I am not thinking is a good idea.  Well, at least, the part about Klar.  

<details><summary>keep recommending TorBrowser&maybeOrbot APK as now 1A, keep recommending braveBrowser as now 3A, but suggest recommdending firefoxFocus as 2A rather than firefoxKlar which is less-field-tested and FDroid-wedged at present </summary><p>

Currently there are three browsers:  1) TorBrowser the soft-fork-of-ESR-which-is-very-lightly-forked-stable, 2) Firefox stable-channel, and 3) BraveBrowser soft-fork of chromium these days i.e. not muon/electron anymore.  So privacyToolsIO is already recommending TorBrowser and BraveBrowser, on whatever platforms they support, including mobile flavors.  

After this commit is pushed, there will be the difference that 2) will become FirefoxESR the very-light-soft-fork-of-stable.  There is no android APK for that, and no iOS either.  

TorBrowser on android **cannot** be a slight soft-fork of FirefoxESR (like TorBrowser on desktop-platforms which benefits from IBM and Canonical and Oracle and so on cooperation on backporting), and moreover, is in alpha.  Orbot is still around, but sunset'd.  

BraveBrowser on android probably *functions* fine, but I don't like using a chromium-based browser on an AOSP-based LineageOS because that is putting all the security in the google-basket, effectively (even if sans-gApps sans-playStore the OS and the browser are both still soft-forks of google-controlled upstreams).  

We could keep recommending firefox-rolling-release-stable version (with addons and tweaks and such) for android and ios, where it is fully supported, and recommend firefox-esr on linuxDesktop/osx/win7 systems where it is fully supported.  The downside there is that org.mozilla.firefox v66 <a href="https://reports.exodus-privacy.eu.org/en/reports/67787/">contains</a> LeanPlum analytics and AdjustSDK, although <a href="https://support.mozilla.org/en-US/kb/how-do-you-use-leanplum-firefox">fairly benign</a> utilization arguably.  Rather than firefoxKlar (see notes below on stuck-ness), my recommendation would probably be <a href="https://en.wikipedia.org/wiki/Firefox_Focus">firefoxFocus</a>, which drops LeanPlum and drops a <a href="https://reports.exodus-privacy.eu.org/en/reports/39500/">couple dozen permissions</a> as well (many of them *excellent* things to drop like 'CAMERA' perm).  The main downside to firefoxFocus is it still includes AdjustSDK tracker in the APK -- the code for the SDK is libre-licensed however -- and it also uploads all URLs visited to GoogleSafeBrowsing, to help prevent endpoint-pwn'age (privacyToolsIO website recommends an about:config tweak to disable that though).  You can get the firefoxFocus code on github, the firefoxKlar variant is on F-Droid proper, firefoxFocus is in the @IzzySoft repo (still?) methinks.  

*  https://android.izzysoft.de/applists/category/named/network_webbrowser#group_879
*  **1A**.  <s>&nbsp;Orfox APK&nbsp;</s> <ins>TorBrowser_alpha APK</ins>:  **10,000k dl**, 50k reviews, orfox is sunset'd as of "early 2019" (and contains googleAds?) but replacement is still in alpha-state.  
*  **2A_stable**.  firefoxStable APK: 100,000k dl, 3210k reviews, 4.4 rating, <a href="https://support.mozilla.org/en-US/kb/how-do-you-use-leanplum-firefox">2 trackers</a>, 32 perms
*  **2A_focus**.  firefoxFocus APK:  **5,000k dl**, 46k reviews, 4.4 rating, <a href="https://github.com/mozilla-mobile/focus-ios/wiki/Install-Tracking-in-Firefox-Focus-with-the-Adjust-SDK">1 tracker</a>, 9-to-11 perms, github 
*  **2A_klar**.  firefoxKlar APK:  100k dl, 2k reviews, 4.3 rating, ~0 trackers(?), 10 perms, f-droid
*  **3A**.  braveBrowser APK:  **10,000k dl**, 89k reviews, 4.3 rating, <a href="https://reports.exodus-privacy.eu.org/en/reports/59965/">0 trackers</a>, 23 perms, soft-google-glued (still?).  Believe it requires android7+ whereas firefoxFocusAndKlar work on android5+ so another reason not to ditch firefox-on-mobile methinks 

This is <a href="https://github.com/mozilla-mobile/focus-ios/wiki/Install-Tracking-in-Firefox-Focus-with-the-Adjust-SDK">the helpdoc from 2017</a> explaining that firefoxFocus uses adjustSDK for "install-tracking" to determine whether the enduser installed "due to an advertising campaign of mozilla foundation" or not, and firefoxKlar does not have adjustSDK tracker.  2017 complaints about AdjustSDK == https://www.ghacks.net/2017/02/12/firefox-focus-privacy-scandal , early 2018 complaints about AdjustSDK, https://chefkochblog.wordpress.com/2018/05/25/firefox-focus-the-privacy-browser-which-collects-and-sells-your-data/ , note that the 'event tracking' stuff was removed from both firefoxFocus and firefoxKlar in ~v3.3 but the 'install tracking' stuff does still exist.  

There was a shift in the firefoxFocus-slash-firefoxKlar implementation a few months ago, it now uses GeckoView, and the F-Droid folks are <a href="https://forum.f-droid.org/t/firefox-klar-update/5138">stuck on firefoxKlar 6.x</a> of August 2018 due to build-related difficulties, <a href="https://github.com/mozilla-mobile/focus-android/releases">upstream</a> is at 8.0.9 nowadays.  IzzySoft is hosting firefoxFocus in their android-repo I believe, but it is not in F-Droid "itself" (though because of the build-problem neither is firefoxKlar *properly* in F-Droid nowadays either).  

There is also a <a href="https://itunes.apple.com/us/app/firefox-focus-privacy-browser/id1055677337?mt=8">firefoxFocus4ios</a>, which has good review-ratings, 18k reviews and 4.8 rating.  Like on android, firefoxKlar4ios has lower review-count and lower rating, 4k and 4.7 today.  Not sure if the tracker-situation is the same over on the ios flavour, I'm only familiar with android APK tracker-detecting-websites.  In particular, it looks like the firefoxFocus for iOS is available on the english-language itunes store, but I'm only finding firefoxKlar for iOS in the german-language portion?  https://itunes.apple.com/de/app/klar-by-firefox/id1073435754?l=en&mt=8  

</p></details> 

So there are some headaches associated with firefoxKlar that don't seem to impact the upstream firefoxFocus (of which firefoxKlar is a soft-fork).  My understanding is that firefoxFocus is *not* really a soft-fork of firefox-stable-rolling-release, it is a project off by itself in most ways (similar to the early days of firefox when it was "a project off by itself mostly-distinct from the huge hulking mozilla-browser").  

But I think either firefoxFocus, if the audience is everyday folks, or firefoxKlar if the audience is hardcore privacy-folks willing to go the extra mile, is important.  Orbot is sunset'd and TorBrowserAPK is alpha-experimental still, BraveBrowser is android7+ and possibly google-glued still, and the regular firefoxStable APK has a few too many compromises for my liking.  

>  give me issue number please

I searched a little and did not find any issue-number for DuckDuckGo browser (nor for FirefoxKlar-versus-FirefoxFocus either).  I think the only other option is the older DolphinZero browser, maybe?  IzzySoft lists it as zero-tracker, but I don't know if it is any good, I've never heard of anybody using it.  Not sure it is libre-licensed either.

IzzySoft:
@five-c-d "FirefoxKlar-versus-FirefoxFocus": Klar has no trackers, Focus has something (I don't remember what). Klar is really quite barebones – OK for the "quick link" you open from some mail/chat/whatever, but not a "daily driver" with bookmarks etc (this matches Focus as well; the two are mostly identical, just Focus has some telemetry and something other, IIRC).

> but I don't know if it is any good

I haven't tried either. Both were in my repo initially, until Klar was ready and moved to the official F-Droid repo. Later I had to remove Focus as well, as it became too big: it reached 50M+ already due to including Gecko now (initially it was using webview), while my repo allows for ~20M per-app.

Oh, Focus also comes with one tracker, Adjust, as you can see [here at Appbrain](https://www.appbrain.com/app/org.mozilla.focus#app-libraries) – which makes it implausible to me: you cannot say "my product is focused on privacy", and at the very same time integrate trackers into it.

five-c-d:
> you cannot say "my product is focused on privacy", and at the very same time integrate trackers into it

Orfox has four trackers :-)  because it is/was a fennec-soft-fork.  Orbot has zero, and the alpha TorBrowserAndroid has zero ... but it is not well-vetted code yet.  

BraveBrowser for android has zero trackers, but is soft-glued-to-google in your IzzySoft listings -- and needs a lot of permissions.  And is chromium-based so "more risky" in some sense than something not forked from google-upstream.  But most importantly I believe it only works with android7+ so older devices cannot use it if they are still running android 4/5/6.  

FirefoxKlar has no trackers, but FDroid is two major versions behind (the Gecko engine thing that caused you to evict Focus from your custom IzzyRepo), and playStore seems to only have German-language FirefoxKlar??  It is considerably less-vetted than FirefoxFocus in terms of android&ios review-counts, but as Klar is a very light soft-fork of Focus that might not be crucial.  Does anybody know whether FirefoxKlar disables SafeBrowsing by default, unlike FirefoxFocus? Supports back to android5, like upstream Focus does.  

FirefoxFocus has one tracker, AdjustSDK, and they claim it is used just once at install-time (back in 2017 timeframe both Klar and Focus had the tracker and used it at install-time AND also at runtime for 'event tracking').  I believe if you installed Orbot first, or TorBrowserAlpha first, and then used either of those to install FirefoxFocus, it would blind their install-tracker :-)  But for everyday endusers that a clicking on playStore and using the playStore **app** this might not help.  

Like braveBrowser for android, FirefoxStable APK *is* a daily-driver type browser, which has two "fairly benign" trackers:  AdjustSDK like Focus, and also LeanPub -- see my comments <a href="https://github.com/privacytoolsIO/privacytools.io/pull/881#issuecomment-486527559">inside the collapse box</a> "suggest recommdending firefoxFocus as 2A rather than firefoxKlar" up above.

IzzySoft:
I wouldn't recommend BraveBrowser for privacy reasons (it looks a bit weird to me, as they just replace the ads) unless you are sure it does no tracking etc. Might be good if you want to get money for browsing (and don't care about probably paying for that with your data; not sure about this part, though) – and the "nagging part" might be minimized. But as long as you cannot validate the code, you cannot say what tracking might be in. As for the (minimum) Android version: when at a desktop browser, simply hover your mouse over the app's name and it will tell you.

> FirefoxKlar has no trackers, but FDroid is two major versions behind (the Gecko engine thing that caused you to evict Focus from your custom IzzyRepo), and playStore seems to only have German-language FirefoxKlar??

FirefoxKlar is the German variant of FirefoxFocus (the two share the very same Github repo) – so basically it is a "special build variant" shipping Focus without trackers. So yes, it absolutely makes sense that the German variant is available in German only 🤣 As for being behind in F-Droid: most likely waiting for Fennec, so the Gecko stuff is built. That's a rather complex task I've been told.

five-c-d:
> As for the (minimum) Android version: when at a desktop browser, simply hover your mouse over the app's name and it will tell you

IzzySoft has a comparison of a few dozen browsers... so yes, over there you can hover to see the minimum android-version.  But privacyToolsIO has just three browsers:  torBrowser which is alpha on android, firefox-or-maybe-firefoxESR if this pull-request is committed, and braveBrowser which is minimum-android7+.  I think that android7+ problem, and that alpha-or-4trackers problem with TBB/Orfox, is reason enough to explicitly recommend either Klar/Focus/Stable flavours of firefox-on-android&ios here in the listings.  

@Mikaela was originally arguing for FirefoxKlar, and I think that is fine as long as the enduser wants a non-daily-driver browser and is savvy enough to figure out how to get it installed (german itunes repo and outdated fdroid repo and other headaches).  FirefoxFocus is a bit less headache to use, and very similar to Klar except for install-tracker via AdjustSDK and (maybe) SafeBrowsing on-by-default, which I think is a good tradeoff.  

If they need a **daily-driver** browser on smartphone though, only FirefoxStable with two trackers, and BraveBrowser with zero-detected-trackers-according-to-exodus (not the same as NO tracking of course!) are the only "one-wizard-of-hassle" options.  BraveBrowser has the advantage that they proxy SafeBrowsing via their own braveServer to keep google from getting every URL visited, but *without* leaving the everyday enduser insecure.  But yeah, there are a lot of controversial aspects to BraveBrowser -- the advert-swap-thing is off-by-default, which helps, I guess. 

> as long as you cannot validate the code

I think <a href="https://www.reddit.com/r/brave_browser/comments/a2y9iv/is_brave_fully_opensource_or_just_some_parts/">the client-side portion</a> of braveBrowser is 100% libre, is that *NOT* true?

beerisgood:
I wonder why nobody say one word about Fennec: Fennec F-Droid (Browse the web) - https://f-droid.org/app/org.mozilla.fennec_fdroid

It's F-DROID build of Firefox but removed a lot telemetry and other shit.
And they update it ASAP after official Firefox releases

Mikaela:
Wording changed.

> I wonder why nobody say one word about Fennec: Fennec F-Droid (Browse the web) - https://f-droid.org/app/org.mozilla.fennec_fdroid

I didn't remember it exists, the only Firefox with updates outside of F-Droid used to be Nightly. But there probably should be a different issue for mobile recommendations anyway.

five-c-d:
>  But there probably should be a different issue for mobile recommendations anyway

You want to split out the browser-section into a top3 recommendations for desktop, and a separate area which has top3 recommendations for mobile?  That is how the OS area already works... and actually, there could be a separate top3 for portable-browsers, I guess, to mirror the LiveCD stuff?  I would agree that is logical, but I don't think there are enough web-browsers to fill the listings.  

At the moment, there is only the one listings-page related to browsers, and it does not have any worthMentioning options.  FirefoxStable is on all major platforms, BraveBrowser is on all major platforms, and although TorBrowser is only officially available on desktop-platforms (since it is a soft-fork from firefoxESR which is only on desktops), the TorBrowser listing explicitly links to alpha/unofficial versions of TorBrowser for android and iOS with direct links.  

>  not convinced moving our recommendation to ESR is necessarily the best

If we change from firefoxStable in the second slot to firefoxESR in the second slot, my expectation is that we need to link to firefoxFocus / firefoxKlar / firefoxStable-for-mobile / or maybe firefoxFennec (there is an fdroid-flavor but not a playStore-flavor except for Nightly-aka-Aurora which is unsuitable for normal endusers) from the ...Android, iOS... portion of the firefoxESR slot.  

Per the arguments made above, I think firefoxFocus is probably the correct pick since the firefox-in-slot-two is primarily intended for people that could not stomach the somewhat-higher-difficulty of TorBrowser in slot-one.  Klar is a little too much hassle, and fennec way too much hassle, to get installed methinks.  I'm not against linking to them, but think they should be in WorthMentioning, rather than highlighted in the top3 portion.

Mikaela:
#858 is closed so I guess there is no point keeping this open either.

Issue #882, opened 2019-04-24
*please* revert to old privacytools.io website - with all elements as one singe wall 


qubecub:
Some time ago privacytools.io change the layout of the site introducing "categorized" view with drill-down sections to look up specific group of security applications, technologies or topics. This is *very* bad idea. 
The way privacytools site was structured initially was great as it allowed me to just scroll one page and I could identify differences, new additions at the first glance. Now, with the new design it takes ages to navigate, tons of clicks, back and forward navigation. It is just so impractical. 

I *urge* you, *PLEASE* go back to the old single-wall design of your site for the privacy applications you're presenting. 

BTW. I love what you do, I love the content, I just can't help but hating the new design. 

I would be immensely grateful (as would many other users as I saw on multiple forums) if you do revert to the old privactytools site design.

Thank you!

Mikaela:
You can still use https://www.privacytools.io/classic/ .

Mikaela:
Reopening, because I cannot see where to find /classic/ without knowing it's there, maybe it should be an arrow next to the home button to find it?

qubecub:
Hi Mikaela, 
You're the bearer of great news. I didn't know the 'classic' tag supports the old view. In this case my world is whole again :)
Many thanks! 
qubecub

qubecub:
Just tested it and it works ! Perfect.

five-c-d:
> I cannot see where to find /classic/ without knowing it's there

It is linked to from the homepage, but very "subtle" and easy to miss.  I found it but only after like a month.  Right under the six big super-categories (provider / browser / SW / OS / svcs / donate) there is a line of text with a hyperlink:  

*  :information_source: Prefer the classic site? It's <a href="https://www.privacytools.io/classic/">still around</a>!

Right under that is a huge "Privacy? I don't have anything to hide" section header.  So the small line of normal-size text gets a little lost-in-the-visual-sea and I suspect many people miss it.  I also don't think "the classic site" is a good description... recommend saying something like this: 

##  &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;:information_source: Want *ALL* the recommendations 
##  &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;filling <a href="https://www.privacytools.io/classic/">**One *GIANT* Page**</a> 
[...]

PR #883, opened 2019-04-24
WIP: resources: move the classic site on top for more visibility

Mikaela:
The classic button wasn't within the first screenful of text, so I missed it while reopening and closing #882.

Resolves: #882


Mikaela:
I will take some from #882 at least the one page part. I cannot see a button for turning this into a draft though.

JonahAragon:
I'm fine with merging this simple change for now, and then any additional changes requested in #882 can be discussed and merged in another PR.

PR #900, opened 2019-05-04
.travis.yml & build.sh: add html5validator & profiling

Mikaela:
Enable html5validator and jekyll profiling.

We aren't valid HTML5 currently. I blacklisted classic, because I think it's taking errors that aren't errors on the new site and I am in belief that it's not cared about that much.


JonahAragon:
I'm not sure how travis works.

Mikaela:
@Shifterovich ping? :)

nitrohorse:
I think we can close and fold this into https://github.com/privacytoolsIO/privacytools.io/issues/1093.

PR #908, opened 2019-05-07
[Discussion] Add Osada to Social Networks

asddsaz:
**Description**: Add popular social network Osada to social networks.


Mikaela:
I cannot remember hearing of Osada before. Is it related to Hubzilla? Has it been discussed anywhere on PTIO previously?

Mikaela:
It doesn't look that popular on [the-federation.info](https://the-federation.info/osada), but they don't handle multiple nodes on the same domain https://github.com/thefederationinfo/the-federation.info/issues/168 so their stats may be a bit off.

blacklight447-ptio:
Besidesthe fact that osada seems rather dead, there has not been any discussion about it on the issue tracker, kindly start and issue tracker there and wait for a community consensus before opening a pull request.

PR #915, opened 2019-05-08
[Discussion] Add Instances for Federated Services

ghbjklhv:
**Discussion**: Add Instances For Federated/Self-Hosting Services like Mastadon or Nextcloud
**Why**? In my experience, many users won't go through the effort of choosing an instance. 
This way users will have less steps to actually start using a service.

**Some instances I recommend**:
Mastadon: [Mstdn.io](https://www.mstdn.io/)
Diaspora: [Diasp.org](https://www.diasp.org/)
Frendica: [Quitter](https://quitter.pl/)

For Nextcloud it is very easy to find and set up an instance so I just provided the provider list.

**What needs to be discussed?**
First of all, are there any major flaws with these instances?
Secondly, are there any other services that need an instance listed.

And the usual, should this PR be merged :)

### DIFF
--- a/_includes/sections/calendar-contacts-sync.html
+++ b/_includes/sections/calendar-contacts-sync.html
@@ -12,6 +12,7 @@
   url="https://nextcloud.com/"
   footer="Client OS: Windows, macOS, Linux, BSD, Unix, iOS, Android, Fire OS. Server: Linux."
   description="NextCloud is a suite of client-server software for creating and using file hosting services. This includes calendar sync via CalDAV and contacts sync via CardDAV. Nextcloud is free and open-source, thereby allowing anyone to install and operate it without charge on a private server."
+  extra_button='href="https://nextcloud.com/providers/">Providers</a>'
   %}
 
   {% include card.html color="primary"
--- a/_includes/sections/cloud-storage.html
+++ b/_includes/sections/cloud-storage.html
@@ -9,6 +9,7 @@ title="Nextcloud - Choose your hoster"
 image="/assets/img/provider/Nextcloud.png"
 description="Similar functionally to the widely used Dropbox, with the difference being that Nextcloud is free and open-source, and thereby allowing anyone to install and operate it without charge on a private server, with no limits on storage space or the number of connected clients."
 website="https://nextcloud.com/"
+extra_button='<a class="btn btn-success mb-1" href="https://nextcloud.com/providers/">Providers</a>'
 forum="https://forum.privacytools.io/t/discussion-nextcloud/287"
 github="https://github.com/nextcloud"
 windows=""
--- a/_includes/sections/social-networks.html
+++ b/_includes/sections/social-networks.html
@@ -14,6 +14,7 @@ github="https://github.com/tootsuite/mastodon"
 android=""
 ios=""
 web=""
+extra_button='<a class="btn btn-success mb-1" href="https://mstdn.io/">Open Instance: Mstdn.io</a>'
 %}
 
 {% include cardv2.html
@@ -25,6 +26,7 @@ forum="https://forum.privacytools.io/t/discussion-diaspora/290"
 github="https://github.com/diaspora/diaspora"
 android=""
 web=""
+extra_button='<a class="btn btn-success mb-1" href="https://diasp.org/">Open Instance: Diasp.org</a>'
 %}
 
 {% include cardv2.html
@@ -38,6 +40,7 @@ windows=""
 linux=""
 android=""
 web=""
+extra_button='<a class="btn btn-success mb-1" href="https://quitter.pl/">Open Instance: Quitter</a>'
 %}
 
 {% include cardv2.html
@@ -62,6 +65,7 @@ ios=""
 linux=""
 windows=""
 web=""
+extra_button='<a class="btn btn-success mb-1" href="https://status.fsf.org/">Open Instance: status.fsf.org</a>'
 %}
 
 <h3>Worth Mentioning</h3>
--- a/_includes/sections/social-networks.html
+++ b/_includes/sections/social-networks.html
@@ -14,7 +14,7 @@ github="https://github.com/tootsuite/mastodon"
 android=""
 ios=""
 web=""
-extra_button='<a class="btn btn-success mb-1" href="https://social.privacytools.io/">PrivacyTools.io Social</a>'
+extra_button='<a class="btn btn-success mb-1" href="https://social.privacytools.io/">Open Instance: social.privacytools.io</a>'
 %}
 
 {% include cardv2.html
--- a/_includes/sections/social-networks.html
+++ b/_includes/sections/social-networks.html
@@ -14,7 +14,7 @@ github="https://github.com/tootsuite/mastodon"
 android=""
 ios=""
 web=""
-extra_button='<a class="btn btn-success mb-1" href="https://mstdn.io/">Open Instance: Mstdn.io</a>'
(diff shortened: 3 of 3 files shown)


BurungHantu1605:
I'd rather recommend our own instance: https://social.privacytools.io/

_includes/sections/social-networks.html
```diff
@@ -14,6 +14,7 @@ <h1 id="social" class="anchor"><a href="#social"><i class="fas fa-link anchor-ic
 android=""
 ios=""
 web=""
+extra_button='<a class="btn btn-success mb-1" href="https://mstdn.io/">Open Instance: Mstdn.io</a>'
```
Mikaela: How do we know we can trust Mstdn.io? We also have our own instance at https://social.privacytools.io/ .

I don't like having to trust us either though, maybe P2P solutions will improve in usability.

_includes/sections/social-networks.html
```diff
@@ -25,6 +26,7 @@ <h1 id="social" class="anchor"><a href="#social"><i class="fas fa-link anchor-ic
 github="https://github.com/diaspora/diaspora"
 android=""
 web=""
+extra_button='<a class="btn btn-success mb-1" href="https://diasp.org/">Open Instance: Diasp.org</a>'
```
Mikaela: Why Diasp.org? Trust?

_includes/sections/social-networks.html
```diff
@@ -38,6 +40,7 @@ <h1 id="social" class="anchor"><a href="#social"><i class="fas fa-link anchor-ic
 linux=""
 android=""
 web=""
+extra_button='<a class="btn btn-success mb-1" href="https://quitter.pl/">Open Instance: Quitter</a>'
```
Mikaela: Why?

_includes/sections/social-networks.html
```diff
@@ -62,6 +65,7 @@ <h1 id="social" class="anchor"><a href="#social"><i class="fas fa-link anchor-ic
 linux=""
 windows=""
 web=""
+extra_button='<a class="btn btn-success mb-1" href="https://status.fsf.org/">Open Instance: status.fsf.org</a>'
```
Mikaela: I might trust FSF, but why centralize on them?

_includes/sections/calendar-contacts-sync.html
```diff
@@ -12,6 +12,7 @@ <h1 id="calendar_contacts" class="anchor"><a href="#calendar_contacts"><i class=
   url="https://nextcloud.com/"
   footer="Client OS: Windows, macOS, Linux, BSD, Unix, iOS, Android, Fire OS. Server: Linux."
   description="NextCloud is a suite of client-server software for creating and using file hosting services. This includes calendar sync via CalDAV and contacts sync via CardDAV. Nextcloud is free and open-source, thereby allowing anyone to install and operate it without charge on a private server."
+  extra_button='href="https://nextcloud.com/providers/">Providers</a>'
```
Mikaela: I think there is inconsistency on how you link to Nextcloud provider page here and below, but to specific instances on others. Why not something like https://the-federation.info/ ? There is another service that I couldn't find right now.

_includes/sections/cloud-storage.html
```diff
@@ -9,6 +9,7 @@ <h1 id="cloud" class="anchor"><a href="#cloud"><i class="fas fa-link anchor-icon
 image="/assets/img/provider/Nextcloud.png"
 description="Similar functionally to the widely used Dropbox, with the difference being that Nextcloud is free and open-source, and thereby allowing anyone to install and operate it without charge on a private server, with no limits on storage space or the number of connected clients."
 website="https://nextcloud.com/"
+extra_button='<a class="btn btn-success mb-1" href="https://nextcloud.com/providers/">Providers</a>'
```
Mikaela: See above.

_includes/sections/social-networks.html
ghbjklhv: The idea was to list the most popular and respected instances among the privacy community.

But you and @BurungHantu1605 make a good point. I will change this.

_includes/sections/social-networks.html
ghbjklhv: @Mikaela Diasp.org is the un-official flagship instance of Diaspora.

_includes/sections/social-networks.html
ghbjklhv: @Mikaela Again, well known.

The 2 most popularly know Freindica instances that I've seen through the forums are Quitter and Friendica.me. I'm unaware of any major differences between the 2.

_includes/sections/social-networks.html
ghbjklhv: @Mikaela They kinda created GNU Social.
If you have a different instance you would recommend, I'd love to hear it!

_includes/sections/calendar-contacts-sync.html
ghbjklhv: The federation is more hubs in general. 
Nextcloud lists only Nextcloud providers.

_includes/sections/social-networks.html
Mikaela: I would recommend sending people to https://the-federation.info/gnusocial and letting them make their own decision on who to trust, however there is https://github.com/thefederationinfo/the-federation.info/issues/168 which biases their statistics a lot.

_includes/sections/calendar-contacts-sync.html
Mikaela: You can open a protocol page and it only shows instances for that protocol e.g. Mastodon https://the-federation.info/mastodon

_includes/sections/social-networks.html
```diff
@@ -14,6 +14,7 @@ <h1 id="social" class="anchor"><a href="#social"><i class="fas fa-link anchor-ic
 android=""
 ios=""
 web=""
+extra_button='<a class="btn btn-success mb-1" href="https://social.privacytools.io/">PrivacyTools.io Social</a>'
```
Mikaela: This is minor issue, but I don't like saying "PrivacyTools.io Social", because `.social` appears to be a valid TLD so there could theoretically be `privacytools.io.social` and I see a bit room for confusion and in my opinion it seems inconsistent from the other ones that are referred to by domain name.

_includes/sections/calendar-contacts-sync.html
asddsaz: @Mikaela Does "https://the-federation.info/" list Nextcloud providers?

Or is this more about the other hubs like Gnu Social?

_includes/sections/social-networks.html
ghbjklhv: @Mikaela Many like Quitter I did put the name versus the URL.
Others like Diasp.org also call themselves by their URL. 
Others call themselves by the software they run like status.fsf.org is also occasionally just called "GNU Social".

My intentions were to try to make the button smaller. But, seeing as there could be confusion I will change it.

Thanks.

_includes/sections/calendar-contacts-sync.html
Mikaela: That was more about other social hubs, but they do have an issue about Nextcloud at https://github.com/thefederationinfo/the-federation.info/issues/138.

blacklight447-ptio:
I feel the same as jonah, listing these services can feel like an endorsement to some people. if their choice happens to turn out in a bad way, it could backfire to us.

Mikaela:
Closing due to at least three team members agreeing that recommending specific instances is not the best way to go with this.

Issue #914, opened 2019-05-08
💬 Discussion | should VPN-section recommend Tor-use, versus VPN-use? 


five-c-d:
Side-discussion, motivating me to open this issue, found over here == https://github.com/privacytoolsIO/privacytools.io/issues/880#issuecomment-488965108

@blacklight447-ptio has these four recommendations, which I've edited for clarity, let me know if any corrections are needed please, and I will edit:  

1.  we should shove the VPN listings all the way down, to the very bottom of the /classic and /summary pages, on privacyToolsIO, and 
2.  at the top of the VPN-section, we should insert a big 
3.  new sub-section, explaining how Tor works, how Tor helps the reader, and 
4.  why readers should generally prefer Tor-usage, rather than VPN-usage.  

I agree with suggestion#3, and partially agree with suggestion#4. 

<details><summary>discussion of s4</summary><p>

I don't think we should flatly recommend normal everyday endusers MUST only utilize Tor and should NEVER utilize a VPN, because I think VPNs are about 10x more user-friendly than Tor, have about 10x the support-network on normal places like reddit and telephone helpdesks and whatnot, plus if an enduser that is not tech-savvy were to ask their decently-tech-savvy friend 

*  "hey please help me with Tor not working right" versus 
*  "hey please help me with my VPN not working right" 

I suspect there is about 100x chance they will get **better** advice with getting their VPN working properly again, versus the advice they would get to 'help' them make TorNetwork (e.g. Orbot) function ***properly***/anonymously/securely/etc.  

</p></details> 

I disagree with suggestion#1 pretty strongly.  I don't disagree with suggestion#2, except for the bolded part:  "we need a huge subsection promoting Tor **inside** the VPN section" keyword.  We should just *link* from top of VPN-section, to the Tor s3 writeup. 

<details><summary>discussion of s2</summary><p>

This would be, for want of a better word, off-topic clutter.  Better to write a proper TorNetwork_vs_VPNs section (suggestion#3), and then link there from the VPNs section, which is about -- and supposed to be about -- VPNs.  There should be a yellow-bar-warning of a sentence or two, at the top of the VPN listings, just like there is a yellow-bar warning at the top of the LineageOS+UbuntuPhone+GrapheneOS mobile OS listings about GoogleServices.  

Here is what privacyToolsIO says now, about gApps/playStore: 

*  now:  [yellow] ...installing Google Apps may compromise your setup [/yellow]
[...]

PR #918, opened 2019-05-09
Update ProtonVPN

JonahAragon:
Correct number of servers according to https://protonvpn.com/secure-vpn


hugoncosta:
Already at 400, should we just put 400+?

JonahAragon:
Huh, they've been busy lol -- that probably seems fine to me.

JonahAragon:
@Mikaela this was more for ensuring factual accuracy since we already have these services listed, but I would agree that we should probably move forward on #914.

tbnv:
ProtonVNP's paid plans start at $**48**/year, not $96/year. Knowing this may help people in choosing their VPN.

_Sorry to comment in an old thread but I guessed this "Update ProtonVPN" discussion would be good enough._

Issue #884, opened 2019-04-25
💬 Discussion | Copyleft police! come out with your licenses up in the air! :-) 


five-c-d:
There was <a href="https://github.com/privacytoolsIO/privacytools.io/issues/856#issuecomment-485334311">a discussion</a> about an old pull-request that failed to get merged:  the code in question was written in Dec'17 when the privacyToolsIO website was saying the website-contents were CC-BY-SA-4.0, and the github repo against which the pull-request was made was saying the repo-contents were GPLv3.  

The pull-request from @angela-d was about to be merged in Nov'18, but a problem was discovered by an anonymous person:  the license for the pull-request was GPLv3 (as it should have been), but in between the authoring of the pull-request and the attempt to merge the pull-request, the entire website-license and repo-license had both been changed to WTFPL. 

* https://github.com/privacytoolsIO/privacytools.io/commits/master/LICENSE.txt 
* GNU GPL v3.0 .  BurungHantu1605 committed on Aug 10, 2015 
* #418 [wtfpl].  Burung Hantu committed on Apr 5, 2018 

My understanding is that the 2018-04-05 commit was incorrect, because unless I'm missing something, there is no clickwrap on the pull-requests to a contributor-licensing-agreement or something of that nature.  Each contributor to the website, between 2015-08-10 all the way through 2018-04-04, was contributing code (HTML+CSS markup specifically... not programmatic code... but copyright law does not make a distinction) under the GPLv3 terms.  However, each contributor retained copyright-ownership of the material they submitted, unless I'm mistaken.  

Thus, even though @BurungHantu1605 is the founder and project-leader, because they were not the **sole** contributor of material, and because there was no copyright-assignment nor clickwrap implying a contributor-licensing-agreement, then I don't think the strong copyleft nature of GPLv3 (and for that matter the strong copyleft nature of CC-BY-SA-4.0 as well) actually *permit* changing the license to WTFPL.  This is not because the change is a bad idea, and not because the change cannot be properly accomplished -- it is because, specifically, the exact terms of the GPLv3 license-text say explicitly that any derivative work must remain under the same license, unless the copyright-holder(**s**) agree jointly to re-license the combined work.  

[...]

PR #919, opened 2019-05-10
Create WARRANTY

asddsaz:
Add WARRANTY.txt

As listed here: http://www.wtfpl.net/faq/


Shifterovich:
Warranty notice is usually included in the license text.

JonahAragon:
I'm not sure if this is necessary since this is documentation, not source code? Nobody is "running" this on their own machines.

I'd rather see a license change to CC0 if anything.

Mikaela:
> I'd rather see a license change to CC0 if anything.

WTFPL is compatible and would let us perform relicensing to CC0, wouldn't it?

asddsaz:
@Shifterovich It shouldn't matter where you put the warranty.

Shifterovich:
> It shouldn't matter where you put the warranty.

So if it's in the license, why open this PR?

JonahAragon:
@Mikaela it is possible but I don’t know if something like #884 would cause us any issues. 

@Shifterovich I don’t think it is in the WTFPL is it? The license is pretty bare.

Shifterovich:
WTFPL is a shitty license. We should use an actual license like MIT or preferably CC0.

gjhklfdsa:
@Shifterovich As I understand it, warranty can be put anywhere is a source code file and still be valid.

PR #933, opened 2019-05-17
Addd Tranquillity Hosting

gjhklfdsa:
**Description**: I would like to discuss adding Tranquillity Hosting to the hosting section.
They are a zero-log VPS provider based in Sweden.
They also appear to be mostly open source: https://git.frykholm.com/

Their website does not appear to require any non-free JS.
If I understand correctly, new users must apply over phone or email.

However, these are very easy to get anonymously.


gjhklfdsa:
> Where has it been discussed previously?

This is the discussion thread.
I figured since it isn't very popular, I would get many responses in the issues column. :)

blacklight447-ptio:
Kindly first start an issue on the issue tracker where it will be disucssed under normal circumstances, then we can talk about adding it via a pull request :).

PR #940, opened 2019-05-18
Relicense under CC0

JonahAragon:
Releasing our work into the [public domain](https://creativecommons.org/publicdomain/zero/1.0/) seems like the correct path. WTFPL currently grants users an unlimited license to use our work, which is *maybe* fine, but CC0 actually revokes all copyright wherever possible, and falls back to an unlimited license where that's legally not a possibility. CC0 also includes a warranty disclaimer (so this merge Closes #919).

Under the terms of WTFPL we're able to relicense this site however we see fit so there should be no issues converting to CC0.


gjhklfdsa:
Why wasn't this discussed more among the community?

Issue #300, opened 2017-07-31
Why is Tomato not listed in 'Open Source Router Firmware'?


GETandSELECT:
Hey
thanks you guys for fantastic privacy resources collection.

I wonder why [Tomato](https://advancedtomato.com/) is not listed in 'Open Source Router Firmware'?

The same [question popped up on Reddit](https://www.reddit.com/r/privacytoolsIO/comments/38owio/tomato_router_firmware/) two years ago.

Atavic:
It supports a few routers. OpenWrt/LEDE supports many more, see: [lede-project](https://lede-project.org/toh/start) vs [tomato](http://www.polarcloud.com/tomatofaq#what_will_this_run_on)

2E0PGS:
Tomato is a great WIFI AP and router software for the old series of WRT54G and more:
"Linksys' WRT54G/GL/GS, Buffalo WHR-G54S/WHR-HP-G54 and other Broadcom-based routers".

Yeah it maybe limited to the WRT line but there are a tonne of them out there. So I think its coverage is quite decent. I personally own 5 or so WRT54 devices. They are even manufacturing them again due to demand!

I would list it if it was me. Plus people maybe going back to using older routers due to modern backdoor hardware microcode.

Atavic:
I would list it, too. All those projects that fork official router firmware are great. My previous comment was simply an opinion, maybe biased as I see few buffalo products on stores - I live in a EU country.

Vincevrp:
[Quality over quantity](https://github.com/privacytoolsIO/privacytools.io/blob/master/.github/CONTRIBUTING.md#software-criteria), but it might worth mentioning?

Atavic:
As long as the project is active, it's definitely a candidate for one of the three mentions or for a link on *Worth Mentioning* section.

GETandSELECT:
development of Tomato derivates was long time not active. Now FreshTomato (new derivate) release regularly new builds. See http://freshtomato.org

I personally like very much Tomato based Firmware, but I an not an expert.

JonahAragon:
https://github.com/privacytoolsIO/privacytools.io/pull/942#pullrequestreview-242576999

PR #942, opened 2019-05-20
Add FreshTomato to Worth Mentioning in Router section

GETandSELECT:
Description: Add FreshTomato to Worth Mentioning in Router section
Has this been discussed here before: Yes, https://github.com/privacytoolsIO/privacytools.io/issues/300
What do you propose: inform users about Tomato derivates. For more info see https://en.wikipedia.org/wiki/Tomato_(firmware) -> this Wikipedia entry not maintained anymore. But anyway a good starting point.

### DIFF
--- a/_includes/sections/router-firmware.html
+++ b/_includes/sections/router-firmware.html
@@ -29,5 +29,6 @@
 
 <ul>
   <li><a href="https://www.openbsd.org/">OpenBSD</a> - A project that produces a free, multi-platform 4.4BSD-based UNIX-like operating system. Emphasizes portability, standardization, correctness, proactive security and integrated cryptography.</li>
+  <li><a href="https://freshtomato.org">FreshTomato</a> - Alternative open source firmware for Broadcom-based routers.</li>
   <li><a href="https://dd-wrt.com/">DD-WRT</a> - A is Linux-based firmware for wireless routers and wireless access points. It is compatible with several models of routers and access points.</li>
 </ul>


_includes/sections/router-firmware.html
```diff
@@ -29,5 +29,6 @@ <h3>Worth Mentioning</h3>
 
 <ul>
   <li><a href="https://www.openbsd.org/">OpenBSD</a> - A project that produces a free, multi-platform 4.4BSD-based UNIX-like operating system. Emphasizes portability, standardization, correctness, proactive security and integrated cryptography.</li>
+  <li><a href="https://freshtomato.org">FreshTomato</a> - Alternative open source firmware for Broadcom-based routers.</li>
```
Mikaela: Are there any sources for this being official website or related to the original TomatoWRT somehow? I am not very familiar with router firmware.

GETandSELECT: @Mikaela there are many Tomato derivates, but only FreshTomato is in active development. The developers (and users) hang out in https://www.linksysinfo.org/index.php?forums/tomato-firmware.33/.

blacklight447-ptio: their website doesn't seem to support proper HTTPS. it gives me a certificate warning.

GETandSELECT: @blacklight447-ptio yes I saw that. Anyway decided to link the HTTPS version instead of plain HTTP. What you recommend me?

GETandSELECT: let's mention on of the devs @Jackysi - maybe he joins the discussion

ghbjklhv:
As far as I can tell, the projects frontend is non-free software.

I don't think recommending software like this is efficient, as it doesn't highlight the key issues with much of the hardware it supports. 

Proprietary software is inherently a black-box and we have no way of authenticating if it does what it says it does. :)

Issue #948, opened 2019-05-26
💬 Discussion | Instant Messaging page


Perelandra0x309:
Hello all,
I would recommend the IM page be updated with information and recommendations of some apps and I wanted to open a discussion about it before posting a pull request.  There are some great security focused messengers that deserve some mention.

In looking at the 3 websites referenced (full disclosure- I run the securechatguide.org website) There are some apps that all websites agree provide good alternatives.  I will attempt to make a table- 1 or .5 points is for a recommendation from the website, 0 for not recommended and blank is not reviewed.

| App:          |   TP  | SMA | SCG | Total |
| ---------- | ----- | ---- | ---- | ---- |
| Signal        | 1       |      1 |       1 |     3 |
| Wire          | .5      |      1 |       1 |   2.5 |
| Wickr        | 1       |      0 |       1 |     2 |
| Riot           |        |        0 |       1 |     1 |
| Keybase   |        |            |       1 |     1 |
| Threema  |         |       1  |        1 |     2 |

Based on this I think we should add Wire, WickrMe and Threema as recommended apps for instant messengers. 

There are some other apps I think worth mentioning, such as Briar, Keybase and TwinMe.  We should also create a separate section for Experimental/Beta apps since these may be risky to use.

I have made a mockup locally of what I think are reasonable changes.  Please make suggestions, and I can submit a pull request when ready.

![image](https://user-images.githubusercontent.com/22124003/58381115-9a723800-7f87-11e9-8f3c-62a4e662cde3.png)

![image](https://user-images.githubusercontent.com/22124003/58381140-c4c3f580-7f87-11e9-99d6-895147021c59.png)

Thanks for your time and effort on this site, it is a great resource!

Mikaela:
I would prefer seeing a draft pull request so the changes would be more obvious and easier to comment upon than a screenshot.

> TP | SMA | SCG

What are these abbreviations?

> Based on this I think we should add Wire, WickrMe and Threema as recommended apps for instant messengers.

I am moving my family to Wire before removing my WhatsApp account and I agree on it, but I am under impression that Wickr and Threema aren't open source and wouldn't recommend them.

> There are some other apps I think worth mentioning, such as Briar, Keybase and TwinMe. We should also create a separate section for Experimental/Beta apps since these may be risky to use.

[...]

Issue #840, opened 2019-04-09
💬 Discussion | Why is PrivacyTools recommending Riot over Wire when it's less private without a warning on unencrypted metadata?


rawlife56:
Ok! I get it, Riot is federated and their app is a step in the right direction coupled with Matrix protocol but it doesn't come without compromise due to its early beta stage.

privacytools.io mentions 'Wire' stores contact data unencrypted on their servers as their only reason removing it from the top recommendations. Doesn't Riot do the same ? AFAIK, Riot only encrypts the message content as of now if we enable it. Everything from the time stamps to people we have [contacted](https://github.com/vector-im/riot-web/issues/7649), [stickers sent](https://github.com/vector-im/riot-web/issues/6739), few attachments, Call duration-recipients and much more except the message content stays unencrypted somewhere in the matrix server. Isn't this situation much worse than Wire ? Yes, we can host our own server unlike any other client but I really doubt that will be more than a small minority of people. Hiding [metadata](https://github.com/matrix-org/synapse/issues/4565) like Signal does isn't their priority either right now looking at their road map.

So the question arises, Shouldn't we mention the large amount of metadata leaks in Riot on the website warning the users beforehand. I'm really in love with Riot for its stability in such an early stage as a user who used wire since a year as my main IM but people should be aware of caveats because they expect the website to mention it because the same website did warn about unencrypted contacts metadata for another chat application. I wouldn't use Riot over Wire or Signal for anything remotely sensitive in its current state.

I may miss something obvious because I'm still an amateur in this stuff. Would love to know if I'm missing anything.

@muppeth summed it up well [here](https://github.com/privacytoolsIO/privacytools.io/pull/562#issuecomment-457878353). I'm quoting one of his sentence which holds true and explains my concerns in a much easier way 'At this moment I don't see how this(Riot) could be advice as privacy aware service alternative. It's quite possible synapse stores more metadata then whatsapp at this point.'

five-c-d:
Wireapp does do the same, with the difference that if you WANT to secure your metadata and use RiotIM, you actually can:  by running your own Synapse homeserver, or by picking somebody YOU trust that is running one.  Ability to pick the riotIM server you connect with, also gives you ability to pick what jurisdiction that server is hosted in, unlikely with wireapp.  

[...]

Issue #740, opened 2019-01-26
🆕 Software Suggestion | Keybase


pdjpdjpdj:
## Basic Information

**Name:** Keybase
**Category:** e2eE chat, e2eE storage with kbfs, e2eE private git repo, cryptographic account linkability and proof of identity
**URL:** https://keybase.io/

## Description
Imho Keybase should be added. It's early days for a lot of it's features but still already nice filestorage with better guarantees than the competitors. E2E chat and

pdjpdjpdj:
So Adding keybase to?
-  Encrypted Cloud Storage Services
- Secure hosting provider (e2eE git repository hosting)

beerisgood:
Site doesn't have CSP policy, track referer, use Amazon AWS server (from USA)
Not realy recommend for important/ private data.

See https://webbkoll.dataskydd.net/en/results?url=http%3A%2F%2Fkeybase.io

infosec-handbook:
@beerisgood 
>Site doesn't have CSP policy, track referer

You don't have to go to the website to use Keybase.

>use Amazon AWS server (from USA)

DuckDuckGo uses AWS, Signal uses AWS, GitHub uses AWS – all recommended by privacytools.io … So there are more recommendations on privacytools.io that don't match your criteria.

beerisgood:
@infosec-handbook maybe nobody see this problem with Amazon yet?
But it's there and should be noticed

libBletchley:
This criticism of keybase is 1 year old, so I'm not sure how much of it still applies:

# Keybase, we have a problem.
The Keybase software and service are both littered with severe bugs
that create a security and legal nightmare.  Here are some of the
issues:

  * Deception: Their software is a server masquerading as a client
    app.  They simply call it an "app" on this page:
    https://keybase.io/docs/the_app/install_linux but it's actually a
    surreptitious *server* that runs continuously in the background as
    a daemon.
  * Deception: Tor mode serves only to mislead users.  The tool
    actually surreptitiously phones home to the central server of
    Keybase, Inc. without using Tor at all.  This is not the usual DNS
    leak that Tor users are accustomed to, the connection itself takes
    place outside of the #Tor network.  It's not incidental.  This is
    in their _privacy policy_: "When you access or use the Service,
    we automatically collect and store information about your browsing
    habits and your use of the Service (“Usage Information”),
    including: a. Your computer’s IP address.. f. Session times and
    lengths"
[...]

Issue #60, opened 2016-08-02
Add XMPP clients


jubalh:
The section about messengers is sadly very misleading in my opinion.
Have you ever used ChatSecure?
I suppose you recommend it becuase it runs on multiple mobile operating systems.
Are you aware that it is different on each of these, has different featuers?
Can it do http_upload, carbons? Do you tell people about how OTR can also be a pain if you have multiple devices? It doesn't seem so, which will result in users trying the software, seeing that it doesn't work as expected and saying its no good.

In my opinion the best XMPP client for mobile is Conversations, which is mentioned on the page too.

I think one should just mention XMPP in general and then link to a broader explanation of it. Explaining that behaviour of clients can differ depending on which XEPs they support. And listing a good pre selection for people who do not want to read all those details. Which in my opnion is: Conversations for Android, Gajim and Swift for desktop. I can't speak for iOS since I don't use it.
This would also give the user the right impression: it's not just for mobile but for all kinds of things. Currently in my opinion it looks like its a mobile only thing.

jubalh:
Just today: http://www.reuters.com/article/us-iran-cyber-telegram-exclusive-idUSKCN10D1AM?sp=alcms
So much about secure and privacy

PrivacyDefender:
There's a list which compares different servers and their support for different XEPs (https://gultsch.de/compliance.html). However, privacytools.io suggests using OTR/openPGP, while there is a far more sophisticated encryption (OMEMO) available, which is currently supported by Conversations, Gajim and CryptoCat (ChatSecure for iOS already announced to support it with the next app release).

privacytoolsIO:
@jubalh So your idea is to add a "XMPP" recommendation and link to several different clients for Desktop, iOS and Android? 

As far as I know: Conversations for Android and Chatsecure is still good for iOS?

Please help me out here.

jubalh:
@privacytoolsIO :)

Maybe these tips are helfpul:
A list with clients and which XEPs they support. https://www.zash.se/xmpp-clients.html
http://xmpp.iodoru.org/details.html mentions which XEPs are important to have a usable chat experience.
I agree usual users shouldnt have to think about such things thats why currently many clients try to make things easier and implement all the important XEPs.
The best client is Conversations I'd say, I even have the feeling that its the leading example and often drags the others along.
Gajim is a good client for the desktop, but it needs some tweaking (going to settings and download plugins for some of the XEPs). Swift-im is another good client which wants to make things easy.
[...]

PR #951, opened 2019-05-27
Instant Messaging page updates

Perelandra0x309:
## Description

Resolves: #948 

Here is my proposal for changes as discussed in #948 

Overview of changes:

Add criteria to cards with details.  This would be nice to show an overview of how an app adheres to each category, and give a chance to briefly explain any concerns.
Open Source
Cross Platform
Ease of Use
Privacy Respecting
Prevents Mass Surveillance

Add Threema, Wire and WickrMe as cards

Make securechatguide.org link into two links to the EFF and Features Matrix pages

Combine all XMPP clients into one list under the "Worth Mentioning" section

Add Briar, Keybase and TwinMe to the "Worth Mentioning" section

Create a separate "Experimental and Beta" section

### DIFF
--- a/contact.md
+++ b/contact.md
@@ -15,7 +15,7 @@ It's very important to us to stay up-to-date on the latest changes in the privac
 
 [<i class="fab fa-reddit fa-lg fa-fw"></i> Suggest something new on reddit](https://www.reddit.com/r/privacytoolsIO/)
 
-For complete transparency, software will only be considered for this website after discussions take place on either of the above platforms. 
+For complete transparency, software and providers will only be considered for this website after discussions take place on either of the above platforms. 
 
 <span class="text-success"><i class="fas fa-comment"></i> New:</span> We're now on Matrix! Join us at `#general:privacytools.io` to chat with us and other members about this site and privacy in general. If you need a Matrix account, you can sign up with our own homeserver (`https://chat.privacytools.io`) using [riot.privacytools.io](https://riot.privacytools.io/) or any Matrix client! If you prefer XMPP, you're welcome to [use the Matrix.org XMPP bridge to join the discussion](https://conversations.im/j/%23general%23privacytools.io@bridge.xmpp.matrix.org).
 


blacklight447-ptio:
Would add another field for anonymity and one registration information.

_includes/sections/instant-messenger.html
```diff
@@ -1,15 +1,23 @@
 <h1 id="im" class="anchor"><a href="#im"><i class="fas fa-link anchor-icon"></i></a> Encrypted Instant Messenger</h1>
 
 <div class="alert alert-warning" role="alert">
-  <strong>If you are currently using an Instant Messenger like WhatsApp, Viber, LINE, Telegram or Threema, you should pick an alternative here.</strong>
+  <strong>If you are currently using an Instant Messenger like WhatsApp, Viber, LINE, Telegram or Facebook Messenger, you should pick an alternative here.</strong>
```
Mikaela: I disagree with removing Threema from this list. Also could this have alphabetical order or are they in some obscure order of which is the least worst? Personally I have also started calling WhatsApp as `Facebook WhatsApp` (note the nonbreakable white space).

_includes/sections/instant-messenger.html
```diff
@@ -1,15 +1,23 @@
...
 </div>
 
 
+<h3>Mobile Devices</h3>
 {% include cardv2.html
-title="Mobile: Signal"
```
Mikaela: Why remove Mobile?

_includes/sections/instant-messenger.html
```diff
@@ -1,15 +1,23 @@
...
+title="Signal"
 image="/assets/img/tools/Signal.png"
-description="Signal is a mobile app developed by Open Whisper Systems. The app provides instant messaging, as well as voice and video calling.
-All communications are end-to-end encrypted. Signal is free and open source."
+description="Signal is a mobile app developed by Open Whisper Systems. The app provides instant messaging, as well as voice and video calling.<br>
+All communications are end-to-end encrypted. Signal is free and open source.<br>
+<table border=\"0\"><tr><td>
+Open Source: Yes<br>
+Cross Platform: 5 Platforms<br>
```
Mikaela: Isn't it still Android & iOS? I last used Signal desktop today and it wanted linking to my phone by scanning a QR code, so I wouldn't call it as cross-platform, maybe `Android & iOS + X remote clients).

_includes/sections/instant-messenger.html
```diff
@@ -20,11 +28,37 @@ <h1 id="im" class="anchor"><a href="#im"><i class="fas fa-link anchor-icon"></i>
 linux=""
 %}
 
+{% include cardv2.html
+title="Threema"
```
Mikaela: :-1: for not being open source.

_includes/sections/instant-messenger.html
```diff
@@ -20,11 +28,37 @@ <h1 id="im" class="anchor"><a href="#im"><i class="fas fa-link anchor-icon"></i>
...
+description="Threema is a secure messenger that does not require any personal information to sign up.  Threema is partially open source, has been audited and uses the open NaCl encryption library.<br>
+<table border=\"0\"><tr><td>
+Open Source: Partial<br>
+Cross Platform: 2 Platforms<br>
+Ease of Use: Easy<br>
+Privacy Respecting: Yes<br>
+Prevents Mass Surveillance: Yes
+</td></tr></table>"
+website="https://signal.org"
+github="https://github.com/signalapp"
```
Mikaela: I think you forgot to change these two lines while copy-pasting?

_includes/sections/instant-messenger.html
```diff
@@ -20,11 +28,37 @@ <h1 id="im" class="anchor"><a href="#im"><i class="fas fa-link anchor-icon"></i>
...
-description="Riot.im is a decentralized free-software chatting application based on the <a href\"https://matrix.org/\">Matrix</a> protocol, a recent open protocol for real-time communication offering E2E encryption. It can bridge other communications via others protocols such as IRC too. <span class=\"badge badge-warning\" data-toggle=\"tooltip\" title=\"The software is currently in beta and the mobile client states 'End-to-end encryption is in beta and may not be reliable. You should not yet trust it to secure data.'\">beta <i class=\"far fa-question-circle\"></i></span>"
+description="Riot.im is a decentralized free-software chatting application based on the <a href\"https://matrix.org/\">Matrix</a> protocol, a recent open protocol for real-time communication offering E2E encryption. It can bridge other communications via others protocols such as IRC too. <span class=\"badge badge-warning\" data-toggle=\"tooltip\" title=\"The software is currently in beta and the mobile client states 'End-to-end encryption is in beta and may not be reliable. You should not yet trust it to secure data.'\">beta <i class=\"far fa-question-circle\"></i></span><br>
```
Mikaela: Riot 1.0 has been released some time ago, so it's not in beta anymore.

_includes/sections/instant-messenger.html
```diff
@@ -20,11 +28,37 @@ <h1 id="im" class="anchor"><a href="#im"><i class="fas fa-link anchor-icon"></i>
...
+<table border=\"0\"><tr><td>
+Open Source: Yes<br>
+Cross Platform: 6 Platforms<br>
+Ease of Use: Intermediate - Key management and manual device verification<br>
+Privacy Respecting: Yes<br>
+Prevents Mass Surveillance: No, encryption not on by default, syncs with non-encrypted platforms such as IRC
```
Mikaela: (See also https://github.com/privacytoolsIO/privacytools.io/issues/840)

_includes/sections/instant-messenger.html
```diff
@@ -36,10 +70,62 @@ <h1 id="im" class="anchor"><a href="#im"><i class="fas fa-link anchor-icon"></i>
...
 %}
 
+{% include cardv2.html
+title="Wire"
+image="/assets/img/tools/wire.png"
+description='A free software End-to-End Encrypted chatting application that supports instant messaging, voice, and video calls. Full source code is available. 0 trackers. <span class="badge badge-warning" data-toggle="tooltip" title="Wire stores metadata such as your contacts in plaintext (= not encrypted)."><a href="https://www.vice.com/en_us/article/gvzw5x/secure-messaging-app-wire-stores-everyone-youve-ever-contacted-in-plain-text">meta-data concerns</a> <i class="far fa-question-circle"></i></span><br>
+<table border="0"><tr><td>
+Open Source: Yes<br>
+Cross Platform: 6 Platforms<br>
+Ease of Use: Easy<br>
```
Mikaela: It's doing some weird things though like the key verification is a bit hidden.

Mikaela: Where are you taking the number of supported platforms by the way?

_includes/sections/instant-messenger.html
```diff
@@ -36,10 +70,62 @@ <h1 id="im" class="anchor"><a href="#im"><i class="fas fa-link anchor-icon"></i>
...
+android=""
+ios=""
+mac=""
+windows=""
+linux=""
+web=""
+%}
+
+{% include cardv2.html
+title="WickrMe"
```
Mikaela: :-1: for not being open source.

_includes/sections/instant-messenger.html
```diff
@@ -50,7 +136,7 @@ <h1 id="im" class="anchor"><a href="#im"><i class="fas fa-link anchor-icon"></i>
 
 <h3>Complete Comparison</h3>
 <ul>
-	<li><a href="https://securechatguide.org/effguide.html">securechatguide.org</a> - Guide to Choosing a Messenger.</li>
+	<li>securechatguide.org - A <a href="https://securechatguide.org/featuresmatrix.html">Complete Features Matrix</a> and an easier <a href="https://securechatguide.org/effguide.html">Practical Application of EFF's Guide<a/>.</li>
```
Mikaela: I need to read this at a better time to comment.

_includes/sections/instant-messenger.html
```diff
@@ -60,12 +146,24 @@ <h3>Worth Mentioning</h3>
...
-  <li><a href="https://get.wire.com/">Wire</a> <span class="badge badge-warning" data-toggle="tooltip" title="Wire stores metadata such as list of your connections/conversations in plaintext (= not encrypted).">experimental <i class="far fa-question-circle"></i> (<a href="https://www.vice.com/en_us/article/gvzw5x/secure-messaging-app-wire-stores-everyone-youve-ever-contacted-in-plain-text">more info</a>)</span> - A free software End-to-End Encrypted chatting application that supports instant messaging, voice, and video calls.</li>
-  <li><a href="https://status.im/">Status</a> - <span class="badge badge-warning">Experimental</span> A free and open-source, peer-to-peer, encrypted instant messanger with support for DAPPs. </li>    
-  <li><a href="https://en.wikipedia.org/wiki/Off-the-Record_Messaging#Client_support">List of OTR Clients - Wikipedia</a></li>
+  <li>XMPP federated clients:</li>
+  <ul>
+    <li><a href="https://www.chatsecure.org">ChatSecure</a> (iOS) - ChatSecure is a free and open source messaging app that features OTR encryption over XMPP.</li>
```
Mikaela: I hear from iOS users that ChatSecure has issues with notifications and [Monal](https://monal.im/) is more often recommended.

_includes/sections/instant-messenger.html
```diff
@@ -60,12 +146,24 @@ <h3>Worth Mentioning</h3>
...
+    <li><a href="https://play.google.com/store/apps/details?id=eu.siacs.conversations">Conversations</a> (Android) - An open source Jabber/XMPP client for Android 4.4+ smartphones. Supports end-to-end encryption with either OMEMO or openPGP. There is also <a href="https://play.google.com/store/apps/details?id=eu.siacs.conversations.legacy">Conversations Legacy</a> which still supports OTR. 0 trackers.</li>
```
Mikaela: I would prefer linking to https://conversations.im/ instead of Google and mentioning https://f-droid.org/packages/eu.siacs.conversations/ for users who cannot get over the price.

Mikaela: Oh and I think it's OpenPGP with a capital O, but I need t ocheck.

_includes/sections/instant-messenger.html
```diff
@@ -60,12 +146,24 @@ <h3>Worth Mentioning</h3>
...
+    <li><a href="https://zom.im/">Zom</a> (Android, iOS) - Simple to use XMPP client. 0 trackers.</li>
+    <li><a href="https://en.wikipedia.org/wiki/Off-the-Record_Messaging#Client_support">List of OTR Clients - Wikipedia</a></li>
```
Mikaela: I don't think OTR should be recommended anymore as there are platforms with better E2EE by default.

_includes/sections/instant-messenger.html
```diff
@@ -60,12 +146,24 @@ <h3>Worth Mentioning</h3>
...
+  </ul>
+  <li><a href="https://kontalk.org/">Kontalk</a> - A community-driven instant messaging network. Supports end-to-end encryption. Both client-to-server and server-to-server channels are fully encrypted. However media files are <a href="https://github.com/kontalk/androidclient/issues/951">not stored encrypted</a> and your phone number is used as your ID. 1 tracker- Google Crashlytics.</li>
+  <li><a href="https://briarproject.org/">Briar</a> - Peer-to-peer encrypted messaging and forums which communicates over WiFi, Bluetooth and Tor. Verified contacts are made by meeting in person to exchange keys. 0 trackers.</li>
+  <li><a href="https://www.keybase.io/">Keybase</a> - Encrypted chat, groups, file storage, git.  Prove your online identities such as websites, Github, Reddit and cryptocurrency addresses. 0 trackers.</li>
```
Mikaela: I wonder if https://github.com/privacytoolsIO/privacytools.io/issues/740 should reach concensus first?

_includes/sections/instant-messenger.html
Perelandra0x309: Even if Threema doesn't get added as a recommendation, I don't think it deserves to be grouped together with the likes of Viber and Messenger.  Threema is always encrypted using elliptical curve and XSalsa20 protocols.

_includes/sections/instant-messenger.html
five-c-d: You have to have (exactly) one master-device, which is always going to be either signal4android or signal4ios, in order to **install** signal4desktop, but not to actually use signal4desktop (i.e.  once it is installed).  

The reason is because some functionality is only available within the smartphone-apps, in particular, registration requires the ability to receive inbound SMS or inbound robocall which signal4desktop doesn't do -- because laptops tend not to be able to get inbound SMS/PSTN.  You also need signal4smartphone if you want to make and receive cryptocalls, signal4desktop does not have that yet -- because again, laptops tend to be a much worse UX versus desktops.  

If you are like me though, you install signal4android, then link signal4desktop running on your choice of Linux distro ... after which you can pop the battery out of your smartphone and signal4desktop will work just fine.  You cannot do this *permanently* but signal4desktop is definitely not some kind of "remote control of the phone" type implementation, which is how whatsapp functions for example.  

People that are serious about not having a smartphone as their upstream-device, can use the unofficial github.com/AsamK/signal-cli as a workaround:  register that as your master-device, and then link the (official) signal4desktop client to your (unofficial) sig4cli master, and you need never leave your laptop.  If you want cryptocalling from your laptop, you can also install android into a VM, and then install signal4android into the VM.  It is a bit tricksy to link signal4android-in-a-VM on your laptop, with signal4desktop running on the SAME laptop, because it is tough to QR-scan your own screen, but there are ways to skin that cat.  

Signal4desktop has some downsides, the main ones being A) it uses a few hundred megs of RAM because it is an electron-based app for ease of portability, B) you have to switch gears and use signal4smartphone if you want to cryptocall or perform groupchat-management or a few other such platform-parity things, C) there are not yet any real keyboard-shortcuts available.  But it works fine with texting, file-transfer, voiceNotes aka audio-recordings as a kinda-sorta-substitute for cryptocalls, and most other messenger-type-things.  

Whether the signal4desktop quasi-standalone slave-device "really counts" as cross platform or not, depends on your definition.  To me, signalapp is linux-compatible (works on LineageOS+GrapheneOS and works on Debian+Ubuntu officially as well as Fedora+Arch+etc "unofficially"), and that is what really matters, but YMMV.  Wireapp has "officially unofficial" Linux support, but some people prefer that.

_includes/sections/instant-messenger.html
five-c-d: RiotIM folks are still cautious about their crypto, and MegOlm is still "officially" in beta.  But I agree, in practice they are quite mature, self-imposed overly-strict we-still-call-ourselves-beta kind of thing should not be held against them.  https://matrix.org/docs/guides/faq#what-is-the-status-of-e2e%3F  "End-to-End Encryption is currently in late beta. Rooms can have encryption enabled, but it is not by default."  Hosting your own Synapse server tends to mean you CAN have e2e by default, however, is my understanding

_includes/sections/instant-messenger.html
five-c-d: I have heard the same (from the interwebz rather than from actual XMPP-on-iOS endusers however).  But there are sometimes compatibility issues when you want Monal + OMEMO + ability to talk with people on non-Monal platforms, rumor has it?

_includes/sections/instant-messenger.html
five-c-d: Portions of it are open-source, but yeah, most of it is not, and because charging endusers to join is the business-model, pretty much never has a hope of that either.  

To me though, the detail of "does not require any personal information to sign up" seems wrong... don't you have to enter a credit card number, or something, to sign up?  You are issued a threema hash-num, so you don't have to give out personally identifiable details to USE threema, but you do have to pay, right?  I assume they accept Monero or zCash or whatever, but 99% of everyday endusers are not going to do that, instead they will just use Visa or Amex, doxxing themselves in the process.  

To be clear, I'm not saying that Threema folks *retain* such details, just, that the majority of everyday endusers will *give out* the sensitive details during the payment process.  To me this is a downside, over and above the difficulty of convincing folks to pay for something the marketplace has convinced them ought to be free-as-in-beer

_includes/sections/instant-messenger.html
Mikaela: > Threema is always encrypted using elliptical curve and XSalsa20 protocols.

How can this be confirmed if it's not open source?

_includes/sections/instant-messenger.html
Mikaela: > If you are like me though, you install signal4android, then link signal4desktop running on your choice of Linux distro ... after which you can pop the battery out of your smartphone and signal4desktop will work just fine. 

Do you have a link to documentation or something about this? Regardless I would still find calling it as cross-platform misleading as the user still needs Android/iOS for similar setup and if Signal Desktop would work after death of my smartphone, it would still seem like a temporary measure until I got a new one.

_includes/sections/instant-messenger.html
Mikaela: (sorry, the better time is not now)

_includes/sections/instant-messenger.html
Mikaela: I am having no issues OMEMOing from Gajim and Conversations to Monal user (or vice versa), but I will ask my contact to comment here.

_includes/sections/instant-messenger.html
Mikaela: Correct, OpenPGP is written with capital O judging by https://www.openpgp.org/

_includes/sections/instant-messenger.html
Spydar007: Hello. ChatSecure does indeed have issues with notifications: they are not reliable and seem to only come through sporadically or when the app is opened. However, I have also had issues with Monal being able to upload images (and I believe this is a bug in the way it requests access to Photos on iOS), and ChatSecure is also unable to deal with images correctly.

I should mention that I run the betas of both these apps but I don't think these particular issues are specific to the betas, and these issues could also be an issue with my own server installation.

I'm not aware of any "compatibility issues when you want Monal + OMEMO + ability to talk with people on non-Monal platforms".

I think you are probably best to link both ChatSecure and Monal because they are both viable options for iOS users and are both actively developed.

five-c-d: Ah, sorry I was misremembering.  Monal is the one which has full OMEMO support, there are still some things missing in ChatSecure (and Zom-derived-from-ChatSecure-as-opposed-to-NewZom-derived-from-MatrixOrg).  http://omemo.top/ lists Monal at 100% OMEMO support, but ChatSecure is stuck at ~75% support.  (Zom is listed as "100%" but I suspect that refers only to zom4android which formerly was a soft-fork of ConversationsIM, if I understand things correctly.)  

So my question is, should privacyToolsIO list Monal, instead of zom4ios and chatSecure?  I don't think we want to list **every** XMPP client, we just want to list the ones that everyday people might actually need/want as their "best overall option" for the platform in question:  

*  ConversationsIM from FDroid, 
*  Monal for iOS, 
*  Gajim for desktop, and maybe also 
*  Converse.JS for webapp 

would be my "only from reading the interwebz not from actual experience" recommendations.  Dropping chatSecure because it has less-than-full-OMEMO (plus notifications-problems apparently), dropping Zom because they are switching to Matrix-architecture, dropping ConversationsLegacy because (why is it on there still? honest question), dropping ConversationsPlayStore because why encourage paying when we can encourage FDroid, and **only** mentioning four XMPP apps because four is a *lot* of things to need to mention!  :-)

Mikaela: My list of mostly personal experience is from https://github.com/privacytoolsIO/privacytools.io/issues/60#issuecomment-471736220 and I hear Dino recommended for beginners more often than Gajim even if Gajim is my personal choice.

Worth mentioning: Disroot has usage instructions for multiple XMPP clients mostly including screenshots.

* https://howto.disroot.org/en/communication/chat

five-c-d: Dino is Linux-only, and not 100% OMEMO though, right?  Whereas Gajim is cross-platform and 100% OMEMO, so to me the only one WorthMentioning here on privacyToolsIO is Gajim, because it is the more privacy-respecting and more cross-platform choice.  

The disroot helpdoc is about *chatting* and not about *privacy*:  "SASL and TLS has been built into the XMPP core and E2E encryption can be implemented".  I think that privacyToolsIO should only recommend IM clients that ***actually*** implement end2end crypto, preferably on-by-default.  People that are the readership of privacyToolsIO will want 

1) chat-clients that give them solid privacy and 

2) something that runs on their existing platform AND the existing platforms of all their contacts.  

With care, XMPP gives a reasonable amount of privacy, if you run your own ejabberd/prosody to shield metadata, and if you pick clients with full OMEMO support, **and** if you configure everything such that OMEMO is *always used* correctly (with somewhat-well-vetted crypto implementations).  

The vast majority of XMPP clients are legacy options which don't fully support end2end crypto, or unfinished projects that have yet to fully implement OMEMO, and those are ones we should *not be listing* at all.    That's my strong opinion anyways  :-)

BurungHantu1605:
Thanks for your work on this. I'm willing to add the suggested new messengers and adopt to the new sections, but this code cant be published currently. Our descriptions aim to be brief, and we use icons to show if its open source or what platforms are supported.

_includes/sections/instant-messenger.html
Mikaela: > Dino is Linux-only, and not 100% OMEMO though, right? 

https://omemo.top/ links to https://github.com/dino/dino/issues/36 which is closed and I don't know what they owuld be missing. They are Linux-only though.

_includes/sections/instant-messenger.html
five-c-d: Replied over at #967 with links/etc, but can report from personal anecdata, it works just fine  :-)   Definitely does require *some* kind of master-device-running-android though, or an unofficial workaround like github.com/AsamK/signal-cli or android-in-a-VM if you want to PURELY operate from a laptop, not just 99% from the laptop

_includes/sections/instant-messenger.html
Mikaela: https://github.com/vector-im/riot-web/issues/6779 could be a better link for it.

_includes/sections/instant-messenger.html
Mikaela: I have a contact who found it difficult to find where to verify devices of a contact, but now that I am looking at it, there is an arrow next to the contact name on the menu where it works, so I am not sure it applies.

_includes/sections/instant-messenger.html
Mikaela: Ok, this is probably the third time I read my previous two comments so now it's on my actual todo list, I probably won't remember it otherwise.

_includes/sections/instant-messenger.html
ghbjklhv: @Perelandra0x309 Privacytools.io is tasked with listing the best in privacy, not all.
[Software freedom](https://www.wikipedia.org/wiki/Free_software) can be just as important as ease of use.

ghbjklhv: > If you get the app through the Play or Apple Store then those companies process your payment, not Threema. Threema just knows that you paid somehow and you have a valid registration from Google/Apple. You can also purchase it direct on their website with Bitcoin or PayPal. You can also get a registration code as a gift, you don't have to prove you paid for the license you have by verifying an email or anything.

Unless, they have solved the decades long issue of taking payments and staying private then it is still a privacy violation. 

IMHO, Apple and Goggle Play violate privacy and freedom therefore paying through them may be worse. Donation-only services respect privacy.

Other services implement pay-what-you-want and just use the [honor system](https://www.wikipedia.org/wiki/Honor_system) for verifying you actually paid when connecting. Payment is not linked to account in any way.

Perelandra0x309: > Software freedom can be just as important as ease of use.

Yes it can be, but that is not the philosophy of this website.  https://github.com/privacytoolsIO/privacytools.io/blob/master/.github/CONTRIBUTING.md states:

> Software Criteria
Easy to use. Could your mother use that tool or service? Usability is most important.

I'm just asking if being not open source is a disqualification, which it does not seem to be.  Again quoting:

> Open Source / free software is preferred but not required.


> Donation-only services respect privacy and in many cases allow you to create burner accounts. Which can protect your privacy further. :)
Other services implement pay-what-you-want and just use the honor system for verifying you actually paid when connecting. Payment is not linked to account in any way.

What you are talking about is being anonymous.  That is different than privacy.  This site is focused of protecting what you do or say (privacy) not who you are (anonymity).  So discussions about whether a certain service knows that you specifically paid for their product isn't really relevant to the philosophy of this site as I understand it.  If that is not the sase then the criteria as stated needs updating.

However this thread is getting away from the original comment about Threema not being open source.  I can understand not wanting to include it as a card, as I have written before it is not something others want so I am willing to remove Threema as a card.

Issue #912, opened 2019-05-08
💬 Discussion | Remove Non-Free Search Engines


gjhklfdsa:
Most of the recommended search engines on the privacytools.io website are [non-free](https://www.wikipedia.org/wiki/non_free_software).
Including [StartPage](https://www.wikipedia.org/wiki/Startpage.com), [DuckDuckGo](https://www.wikipedia.org/wiki/DuckDuckGo), and [Qwant](https://www.wikipedia.org/wiki/Qwant).

**Despite there being several amazing [free software](https://www.wikipedia.org/wiki/Free_software) search engines:**
_Open issues/PR:_ 
- Sarchy (#907)
https://github.com/agnelvishal/yacy_search_server

_Closed issues/PR:_
- Gigablast (#700)
https://github.com/gigablast/open-source-search-engine

_Accepted Issues/PR:_
- Jive Search (#701)
https://github.com/jivesearch/jivesearch
- Yacy (#702)
https://github.com/yacy/yacy_search_server
- SearX
https://github.com/asciimoo/searx

Related to #699 

Why is this the current status of the privacytools.io search engine section?
Please discuss! :)

blacklight447-ptio:
While I like open source, I doubt how much actule difference this will make. How can we verify that the one hosting the search engine runs this code?

gjhklfdsa:
> While I like open source, I doubt how much actule difference this will make. How can we verify that the one hosting the search engine runs this code?

LibreJS, it currently has limited support among search engines but many have made some progress to support it or have open issues.

Jive Search: https://github.com/jivesearch/jivesearch/issues/89

Yacy and Sarchy: https://github.com/yacy/yacy_search_server/pull/56

Gigablast (no response): https://github.com/gigablast/open-source-search-engine/issues/146

Many are also [self-hostable](https://www.wikipedia.org/wiki/Self-hosting) like [SearX](https://github.com/asciimoo/searx) which has an instance run by [PrivacyTools.io](https://archive.fo/m9keo) and one ran by [Disroot](https://archive.fo/cpY26).

JonahAragon:
It should be noted that while we prefer open source applications in general, it is not a requirement to be listed. We list the best overall tools in each category because if they are privacy-respecting and *people will use them*, then they are better for most users.

[...]

PR #960, opened 2019-05-31
Remove Non-Free Search Engines

gjhklfdsa:
**Description** Remove Non-free Search Engines
**Has this been discussed before?** Yes, many times. Most popular is probably #912 .
**But don't free software search engines provide low quality results?** I don't know who started spreading this, but non-free engines like [DuckDuckGo](https://www.wikipedia.org/wiki/DuckDuckGo) get their search engines from the same sources.
It is called a meta search engine:  https://www.wikipedia.org/wiki/Metasearch_engine
**What free software search engines are there?** Many, the ones listed on privacytools.io are [SearX](https://www.wikipedia.org/wiki/Searx), [JiveSearch](https://jivesearch.com/), [Yacy](https://www.wikipedia.org/wiki/YaCy), and [MetaGer](https://www.wikipedia.org/wiki/MetaGer).


BurungHantu1605:
Full libre is not a requirement. Usability and good search results is what counts.

Mikaela:
Do we have those requirements listed anywhere? I hope they also include not bubbling.

gjhklfdsa:
@BurungHantu1605 @Mikaela Libre makes everything better.
More transparency, can lead to better privacy.

I wasn't stating it as a requirement, but closed software is not transparent.
Therefore, we have no way to verify.

Libre Software fixes these issues.

gjhklfdsa:
Privacytools.io states in its contribution guidelines to only show the best software for privacy.

If free software improves privacy, then shouldn't it count as a method to rate providers/software?

@BurungHantu1605 @Mikaela @blacklight447-ptio @Shifterovich @JonahAragon @Vincevrp

gjhklfdsa:
Privacytools.io states in its contribution guidelines to only show the best software for privacy.

If free software improves privacy, then shouldn't it count as a method to rate providers/software?

@BurungHantu1605 @Mikaela

blacklight447-ptio:
Thing is, software and providers are different things, whether a provider uses free software is pointless as a requirement as you can never verify it anyway.

PR #963, opened 2019-05-31
Add Warnings to Non-free Email Providers

ghbjklhv:
**Feature**: Add warnings to email providers with non-free software
**Why?**: To highlight the issue of black box providers.
**What providers**: The majority of my sources is from the FSF with the default response being set to "no". Other providers like [Posteo](https://www.fsf.org/blogs/community/fsf-javascript-guidelines-picked-up-by-posteo-webmail), [Kolab Now](https://www.wikipedia.org/wiki/Kolab_Now), and Disroot list themselves as being free software.
https://www.fsf.org/resources/webmail-systems

### DIFF
--- a/_includes/sections/email-providers.html
+++ b/_includes/sections/email-providers.html
@@ -70,7 +70,7 @@
         <td data-value="1000">1 GB</td>
         <td data-value="0"><span class="label label-warning">Free</span></td>
         <td data-value="0"><span class="label label-primary">No</span></td>
-        <td data-value="1"><span class="label label-success">Yes</span></td>
+        <td data-value="1"><span class="label label-success">Yes (backend is non-free)</span></td>
         <td data-value="1"><span class="label label-success">Built-in</span></td>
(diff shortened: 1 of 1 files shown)


blacklight447-ptio:
what would this add though, as you have no way of knowing that the software they claim they use is the same as the software the really run on their servers.

ghbjklhv:
> what would this add though, as you have no way of knowing that the software they claim they use is the same as the software the really run on their servers.

**For the frontend**:
Most free software providers are [LibreJS](https://www.wikipedia.org/wiki/GNU_LibreJS) complaint. 

**For the backend**:
Although, it is very difficult to check without 3rd party certifications, we also cannot verify anything they do on the backend including following their privacy policy. Luckily, if the get caught in many countries this would be misleading consumers and there may be a fine.
Plus, there is little reason for them to not run free software.

ghbjklhv:
@BurungHantu1605 @blacklight447-ptio What was the reason for closing this PR? :)

Issue #920, opened 2019-05-11
✨ Feature Suggestion | P2p Filesharing Client Recommendations


ghbjklhv:
**Feature**: Bittorrent/P2p File-sharing Client Recommendations

**What I can recommend**:
There are very few clients directed at privacy.
To get the ball going maybe add:

- [XD Client](https://xd-torrent.github.io/)

- [OFFLoad](http://offload.sourceforge.net/) - Uses [OFFSystem](https://www.wikipedia.org/wiki/OFFSystem) (not Bittorrent)

- [The Owner Free File System](off.system/) - Reboot of [OFFSystem](https://www.wikipedia.org/wiki/OFFSystem) (not Bittorrent)

**Networks I find worth discussing but don't know enough to endorse**:
- [StealthNet](https://www.wikipedia.org/wiki/StealthNet)
- [Calypso](http://calypso.sourceforge.net/)
- [iMule](https://www.wikipedia.org/wiki/IMule)
- [Phex](https://www.wikipedia.org/wiki/Phex#Anonymous_Phexhttps://www.wikipedia.org/wiki/Phex#Anonymous_Phex)
- [Frost](https://www.wikipedia.org/wiki/Freenet#Tools_and_applications)

**But why not ___ protocol?**
Here is a basic list of popular protocols and issues with them:

**IPFS**: No known TOR or I2p Support

**DAT**: Same as IPFS.

blacklight447-ptio:
I warn against using bittorrent based file sharing for privacy purposes, they often leak you ip address by design(not because they are evil, but just because that how they were built to work.)

gjhklfdsa:
@blacklight447-ptio XD uses I2p.

@ghbjklhv I wouldn't add qBittorrent to main, tho. 
Too easy for somebody to leak their IP.

nchristensen:
[BiglyBT](https://github.com/BiglySoftware/BiglyBT/wiki/I2P) (A fork of [Vuze](https://en.wikipedia.org/wiki/Vuze)) support I2P with a plugin. I believe the I2P console also provides a Bittorrent client.

Mikaela:
> IPFS: No known TOR or I2p Support

https://github.com/ipfs/notes/issues/37

JonahAragon:
I haven't heard of the software listed here and it isn't clear to me how they're significantly better than what we currently recommend (namely Syncthing): https://www.privacytools.io/software/file-sync/

PR #976, opened 2019-06-07
Add P2p File Hosting

ghbjklhv:
**Description**: Add Bittorrent alternatives and clients
**How is this different from the file sharing section?**: Filesharing is very diffrent that file hosting.
Especially, long term hosting like Bittorrent. This is about sharing public files.
**Has this been discussed before**: Yes, #920
**What do you recommend adding**:
To Main:
- [XD Client](https://xd-torrent.github.io/)

- Frost

- [The Owner Free File System](off.system/) - Reboot of [OFFSystem](https://www.wikipedia.org/wiki/OFFSystem) (not Bittorrent)

To Worth Mentioning:

ZeroUp - ZeroNet

I2p-Snark


blacklight447-ptio:
+1 for I2p as bittorrent alternative.

ghbjklhv:
> +1 for I2p as BitTorrent alternative.

:+1: 

>  I think we shouldn't recommend frost, as it seems as a largely abandoned project. I say this because the front page says this: Frost version 2011-05-03 released!

Dang, I didn't even realize. Unless, somebody has a some other argument I plan to remove it.
Very sad though.

>  Also the latest update to the OFFsystem seems to be from 2012.

The original OFFSystem is basically dead.
My recommendation is to add the reboot: https://www.off.systems/

Mikaela:
> How is this different from the file sharing section?: Filesharing is very diffrent than file hosting.
Especially, long term hosting like Bittorrent. This is about sharing public files.

I think we currently have them mixed up on the website and I am not sure I understand the difference.

For example https://ipfs.io/ is currently worth mentioning in file sharing IIRC, but it's more of file hosting, while I would consider file sharing as private and I would use OnionShare for sensitive file sharing instead of IPFS.

I pick IPFS as an example as I am somewhat big user of it.

Issue #936, opened 2019-05-17
🆕 Software Suggestion | GNewSense


gjhklfdsa:
**Description**: Add [GNewSense](https://www.wikipedia.org/wiki/GNewSense) to OSes.
**Why?** [GNewSene](https://www.wikipedia.org/wiki/GNewSense) is a fully free'd version of Debian.
**How?**: I recommend changing the Debian description to:

> Debian is a Unix-like computer operating system and a Linux distribution that is composed of free and open-source software, most of which is under the GNU General Public License, and packaged by a group of individuals known as the Debian project. GNewSense is a completely free/libre and open source version of Debian.

**But isn't Debian fully free'd**: Yes and no, this is what the FSF has to say:

> Debian's Social Contract states the goal of making Debian entirely free software, and Debian conscientiously keeps nonfree software out of the official Debian system. However, Debian also provides a repository of nonfree software. According to the project, this software is “not part of the Debian system,” but the repository is hosted on many of the project's main servers, and people can readily find these nonfree packages by browsing Debian's online package database and its wiki.
> 
> There is also a “contrib” repository; its packages are free, but some of them exist to load separately distributed proprietary programs. This too is not thoroughly separated from the main Debian distribution.
> 
> Debian is the only common non-endorsed distribution to keep nonfree blobs out of its main distribution. However, the problem partly remains. The nonfree firmware files live in Debian's nonfree repository, which is referenced in the documentation on debian.org, and the installer in some cases recommends them for the peripherals on the machine.
> 
> Debian's wiki includes pages about installing nonfree firmware.
> 
> Instead of this nonfree distribution, use one of the free distros which are made from it: gNewSense, and PureOS.

This is very similar to the way [Parabola](https://www.wikipedia.org/wiki/Parabola_GNU/Linux-libre) is listed as a free'd version of [Arch](https://www.wikipedia.org/wiki/Arch_Linux).

blacklight447-ptio:
So what it brings to the table is the removal of an optional contrib repo?

Mikaela:
They appear to be on GNU's list.

* https://www.gnu.org/distros/free-distros.html

I was thinking of [Trisquel](https://trisquel.info/) but they are based on Ubuntu instead.

gjhklfdsa:
> So what it brings to the table is the removal of an optional contrib repo?

Not just that, it also gets all the benefits of being an FSF endorsed distro.
Its bounty program is largely backed by the large Free Software Foundation.
[...]

PR #978, opened 2019-06-08
Add gNewSense

gjhklfdsa:
**Description**: Add [GNewSense](https://www.wikipedia.org/wiki/GNewSense) to OSes.
**Why?** [GNewSene](https://www.wikipedia.org/wiki/GNewSense) is a fully free'd version of Debian.
**How?**: I recommend changing the Debian description to:

> Debian is a Unix-like computer operating system and a Linux distribution that is composed of free and open-source software, most of which is under the GNU General Public License, and packaged by a group of individuals known as the Debian project. GNewSense is a completely free/libre and open source version of Debian.

**But isn't Debian fully free'd**: Yes and no, this is what the FSF has to say:

> Debian's Social Contract states the goal of making Debian entirely free software, and Debian conscientiously keeps nonfree software out of the official Debian system. However, Debian also provides a repository of nonfree software. According to the project, this software is “not part of the Debian system,” but the repository is hosted on many of the project's main servers, and people can readily find these nonfree packages by browsing Debian's online package database and its wiki.
> 
> There is also a “contrib” repository; its packages are free, but some of them exist to load separately distributed proprietary programs. This too is not thoroughly separated from the main Debian distribution.
> 
> Debian is the only common non-endorsed distribution to keep nonfree blobs out of its main distribution. However, the problem partly remains. The nonfree firmware files live in Debian's nonfree repository, which is referenced in the documentation on debian.org, and the installer in some cases recommends them for the peripherals on the machine.
> 
> Debian's wiki includes pages about installing nonfree firmware.
> 
> Instead of this nonfree distribution, use one of the free distros which are made from it: gNewSense, and PureOS.

This is very similar to the way [Parabola](https://www.wikipedia.org/wiki/Parabola_GNU/Linux-libre) is listed as a free'd version of [Arch](https://www.wikipedia.org/wiki/Arch_Linux).

**About the comments on non-free vulnerability fixes:**
To address the concerns made in #936 , I edited the banner to educate users on any possible risks on installing non-free OSes.

**Has this been discussed before?**: Yes, #936 

As always, feedback is welcome! :)

### DIFF
--- a/_includes/sections/operating-systems.html
+++ b/_includes/sections/operating-systems.html
@@ -19,7 +19,7 @@
   image="/assets/img/tools/Debian.png"
   url="https://www.debian.org/"
   tor="http://sejnfjrq6szgca7v.onion"
-  description="Debian is a Unix-like computer operating system and a Linux distribution that is composed of free and open-source software, most of which is under the GNU General Public License, and packaged by a group of individuals known as the Debian project. <a href="http://www.gnewsense.org/">gNewSense</a> is a completely free/libre and open source version of Debian."
+  description="Debian is a Unix-like computer operating system and a Linux distribution that is composed of free and open-source software, most of which is under the GNU General Public License, and packaged by a group of individuals known as the Debian project. <a href='http://www.gnewsense.org/'>gNewSense</a> is a completely free/libre and open source version of Debian."
   %}
 
   {% include card.html color="warning"
--- a/_includes/sections/operating-systems.html
+++ b/_includes/sections/operating-systems.html
@@ -1,7 +1,7 @@
 <h1 id="os" class="anchor"><a href="#os"><i class="fas fa-link anchor-icon"></i></a> PC Operating Systems</h1>
 
 <div class="alert alert-warning" role="alert">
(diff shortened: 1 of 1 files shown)


danarel:
According to the Wikipedia article you link to, GNewsSense hasn't been maintained since 2017 and it's latest release support expired in May of 2018. 

Without regular security updates at best, I fail to see how this is a good recommendation for users.

ghbjklhv:
@kewde gNewSense does still recieve security updates from the FSF and partnering developers.

@danarel Debian releases a new stable about every 3+ years. 
If I recall properly, gNewSense uses the LTS version.

Mikaela:
According to Kewde in https://github.com/privacytoolsIO/privacytools.io/issues/936#issuecomment-500139678 it's Debian 7 and Debian 7 LTS went EOL 31st May 2018 [according to Debian wiki](https://wiki.debian.org/LTS).

They have [no news since Tue 10 May 2016 08:25:14 PM UTC](https://savannah.nongnu.org/news/?group=gnewsense).

PR #993, opened 2019-06-16
Cleanup and combine XMPP client list.  

Perelandra0x309:
## Description

Resolves: #948 

Combine all the XMPP clients into one list.  Also change recommendations:
-Replace ChatSecure with Monal
-Add Gajim (Linux)

Also remove link to Wikipedia list of OTR clients and replace with link to omemo.top website.

### DIFF
--- a/_includes/sections/instant-messenger.html
+++ b/_includes/sections/instant-messenger.html
@@ -60,12 +60,16 @@ linux=""
 
 <ul>
   <li><a href="https://retroshare.cc/">RetroShare</a> - An E2E encrypted instant messaging and voice/video call client. RetroShare supports both TOR and I2P. </li>
-  <li><a href="https://www.chatsecure.org">ChatSecure</a> - ChatSecure is a free and open source messaging app that features OTR encryption over XMPP. </li>
+  <li>XMPP federated clients with OMEMO support:</li>
+  <ul>
+    <li><a href="https://monal.im/">Monal</a> (iOS, MacOS) - An XMPP client in active development.</li>
+    <li><a href="https://conversations.im/">Conversations</a> (Android) - An open source Jabber/XMPP client for Android 4.4+ smartphones. Supports end-to-end encryption with either OMEMO or OpenPGP.</li>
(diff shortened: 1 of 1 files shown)


_includes/sections/instant-messenger.html
```diff
@@ -60,12 +60,16 @@ <h3>Worth Mentioning</h3>
 
 <ul>
   <li><a href="https://retroshare.cc/">RetroShare</a> - An E2E encrypted instant messaging and voice/video call client. RetroShare supports both TOR and I2P. </li>
-  <li><a href="https://www.chatsecure.org">ChatSecure</a> - ChatSecure is a free and open source messaging app that features OTR encryption over XMPP. </li>
+  <li>XMPP federated clients with OMEMO support:</li>
+  <ul>
+    <li><a href="https://monal.im/">Monal</a> (iOS, MacOS) - An XMPP client in active development.</li>
```
ghbjklhv: Does not meet the standard of being cross platform.

_includes/sections/instant-messenger.html
```diff
@@ -60,12 +60,16 @@ <h3>Worth Mentioning</h3>
 
 <ul>
   <li><a href="https://retroshare.cc/">RetroShare</a> - An E2E encrypted instant messaging and voice/video call client. RetroShare supports both TOR and I2P. </li>
-  <li><a href="https://www.chatsecure.org">ChatSecure</a> - ChatSecure is a free and open source messaging app that features OTR encryption over XMPP. </li>
+  <li>XMPP federated clients with OMEMO support:</li>
```
Mikaela: I think the "federated" may be unnecessary word here.

danarel: It might be, but I also think it will help people understand how it works. With Signal, Wire, etc, being chat's between specific clients, some people new to XMPP may not realize you can chat with any client you want, with a user on any other client, etc. So while it may not be "necessary" it may still be useful.

Perelandra0x309:
Oops, it looks like the standalone Conversations list item and the OTR link wasn't removed during the merge.

PR #997, opened 2019-06-17
Remove Threema from explicit non-recommendation, alphabetize list, …

Perelandra0x309:
…add WhatsApp EFF article link, add SMS messages mention

## Description

Resolves: #948 

-Remove Threema from explicit non-recommendation
I don't think Threema should be categorized along with WhatsApp, Line, etc as being explicitly not recommended.  The only reasons I have encountered in discussions for others not liking it is that it is not fully open source and costs money (the price of a coffee).  Threema uses the open source NaCl box encryption model using elliptical curve DH 25519 key exchange and XSalsa20 encryption with all messages encrypted end to end.

-Alphabetize the remaining list of not recommended apps.

-Add a link for WhatsApp to an EFF article expressing their concerns with it.

-Add regular SMS messages to the list.  This may not really fit, as it is almost impossible to completely eliminate using SMS.

### DIFF
--- a/_includes/sections/instant-messenger.html
+++ b/_includes/sections/instant-messenger.html
@@ -1,7 +1,7 @@
 <h1 id="im" class="anchor"><a href="#im"><i class="fas fa-link anchor-icon"></i></a> Encrypted Instant Messenger</h1>
 
 <div class="alert alert-warning" role="alert">
-  <strong>If you are currently using an Instant Messenger like WhatsApp, Viber, LINE, Telegram or Threema, you should pick an alternative here.</strong>
+  <strong>If you are currently using an Instant Messenger like LINE, Telegram, Viber, <a href="https://www.eff.org/deeplinks/2016/10/where-whatsapp-went-wrong-effs-four-biggest-security-concerns">WhatsApp</a> or plain SMS messages you should pick an alternative here.</strong>
 </div>
 
 


ghbjklhv:
Without transparency we have no way to verify the encryption standards.
Plus, we cannot verify if it is doing anything else like tracking us.

**Privacy** is impossible without **Free Software**.

Perelandra0x309:
> Without transparency we have no way to verify the encryption standards.

Threema has been very transparent about it's encryption. In fact you can take saved Threema message ciphertext and your private Threema key and use the NaCl library outside of Threema to decrypt your messages and verify them.  See https://threema.ch/validation/

> Plus, we cannot verify if it is doing anything else like tracking us.

Yes you can, it's called network packet inspection.  There's this cool tool called WireShark.  If you can capture any tracking communications back to Threema, let us know.

> **Privacy** is impossible without **Free Software**.

These blanket statements aren't adding anything to the discussion.  PTIO is not an extension of the FSF.  I have told you before to go look at the PTIO contribution guidelines.

Perelandra0x309:
> I disagree with this PR due to Threema not being open source.

There are a lot of messengers that are not completely open source.  Why single out Threema and not mention others such as BBM, Confide, Eleet, FireChat, Hoccer, Keybase, SafeSwiss, Sid, StealthChat, TwinMe, Zangi, Google Hangouts/Chat, Discord, Yahoo, iMessage, Facetime, Slack or WeChat? And these are only small list of the most widely known apps.

blacklight447-ptio:
> > Without transparency we have no way to verify the encryption standards.
> 
> Threema has been very transparent about it's encryption. In fact you can take saved Threema message ciphertext and your private Threema key and use the NaCl library outside of Threema to decrypt your messages and verify them. See https://threema.ch/validation/

How dont you know that they dont simply send your key to their servers ? :)

> 
> Yes you can, it's called network packet inspection. There's this cool tool called WireShark. If you can capture any tracking communications back to Threema, let us know.

Threema will always connect to threemas server to send messages, how do you see the difference between normal threema network traffic and tracking?

> > **Privacy** is impossible without **Free Software**.
> 
> These blanket statements aren't adding anything to the discussion. PTIO is not an extension of the FSF. I have told you before to go look at the PTIO contribution guidelines.

These are not blanket statements, they are valid concerns, and I agree with them.
as for why single out threema, there is not any particular reason why i has to be threema, but threema is a fairly well known messenger that claims to be secure, yet we cannot verify that, by including it in the list we have it as an example that even messengers that claim to be secure, does not mean they are as we have no means of verifying this.

five-c-d:
> as for why single out threema, there is not any particular reason why i has to be threema, but threema is a fairly well known messenger that claims to be secure

Threema has about 55k playStore reviews... making it roughly a hundredfold smaller than iMessages+Facetime, which are more closed than Threema.  

But I think the biggest reason that it is a little jarring to see threema in the yellow area, is because it is lumped in with the likes of facebook apps.  

If threema "must" be kept in the list of avoid-for-sure (but e.g. WickrMe and iMessages *not* listed), we should at least have two sentences.  One for products that we don't have a special reason to believe are privacy-violating such as threema which has partially-libre e2e crypto on by default, and another for projects that there IS a special reason to believe are risky (because they have closed-source crypto that is off by default in particular would be red flag of a completely more severe nature).  

I think the purpose of privacyToolsIO is to list **good** recommended tools, and the listing of tools-to-avoid is far less crucial.  Either a tool is top3, or WorthMentioning, or implicitly NOT worth mentioning ... but being listed in the yellow-warning-bar, should be for things that regular endusers often ignorantly or mistakenly believe are private.  If they are supposed to be just "list of things that are not perfect" then we need to have a MUCH longer list :-)

Mikaela:
> There are a lot of messengers that are not completely open source. Why single out Threema and not mention others such as BBM, Confide, Eleet, FireChat, Hoccer, Keybase, SafeSwiss, Sid, StealthChat, TwinMe, Zangi, Google Hangouts/Chat, Discord, Yahoo, iMessage, Facetime, Slack or WeChat? And these are only small list of the most widely known apps.

I don't know the history of how Threema ended up on that list, but I guess it's the only one of those that advertises being E2EE, but isn't open source. I didn't start researching that list, so it's possible that some of those are fine.

Keybase is open source and E2EE, Discord and Slack aren't E2EE and have their own issues, WeChat is controlled by Chinese goverment and iMessage is iOS-only and I guess most of people using it may not even realize they are using it as it's integrated to the SMS app?

nitrohorse:
I agree with @five-c-d and @JonahAragon. LGTM.

five-c-d:
> Keybase is open source and E2EE

Per commentary on another thread, it looks like keybase *client* is libre, but their server-side code is closed/proprietary.  This is not a dealbreaker to them being WorthMentioning perhaps, but the current listings (signalapp + wireapp + riot&matrix + linphone) are all libre-licensed for 99% of their codebases -- i.e. including server-side.  

Not that that has much to do with whether threema ought be mentioned :-)  But figured I would say that here, before the thread got too stale  ;-)  Thanks for merging, nitrohorse

PR #1001, opened 2019-06-20
Add Security Policy

### LINKED ISSUE #988: 💬 Discussion |  Add Security Policy (SECURITY.MD)
Since privacytools.io has somewhat recently become a service provider, I suggest we have an official bug reporting policy. 

I know we entirely/mostly host existing projects, but it is possible we could have misconfigurations in the server or hosted software. Since we are handling private information, we should have an avenue for people to report issues.

Github supports adding a security policy file to projects: [see docs on this](https://help.github.com/en/articles/adding-a-security-policy-to-your-repository). The policy would be visible in the file and [here](https://github.com/privacytoolsIO/privacytools.io/security/policy)

Secondarily, I suggest we add a [security.txt](https://securitytxt.org/) file on the main website, which is a recent standard. I could see an argument to not add this file because it arguably results in few productive reports, as indicated by [this hacker news discussion](https://news.ycombinator.com/item?id=19151213)

Obviously, a few things would have to be out of scope:

* excessive automated scanning
* denial of service
* user's accounts
* social engineering
* infrastructure PTIO is not in control of

@JonahAragon

JonahAragon:
As recommended in #988:

> Since privacytools.io has somewhat recently become a service provider, I suggest we have an official bug reporting policy.

This sounds good to me. I added some basic instructions to a security policy file, do you think this is enough or well-constructed? Let me know if anything should be added @beardog108.

---

Closes #988

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,32 @@
+# Security Policies and Procedures
+
+This document outlines security procedures and policies for the `privacytools.io` repository/code and all services hosted by privacytools.io, such as Mastodon, Matrix, Riot, et cetera.
+
+## Reporting a Bug
+
+We take all security bugs related to our code and our infrastructure very seriously. Thank you for improving the security of our projects and services. We appreciate your efforts and responsible disclosure, and will make every effort to acknowledge your contributions.
+
+Report any security bugs by emailing the services administrator at [admin@privacytools.io](mailto:admin@privacytools.io).
+
+The administrative team will acknowledge your message within 48 hours, and will provide a detailed response within 72 hours detailing the next steps for handling your report. After our initial reply we will make every effort to keep you informed of the progress towards a fix and announcement, and we may ask for additional information or guidance.
+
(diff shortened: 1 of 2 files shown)


beardog108:
Thanks for adding. Looks pretty good, but i'd clarify that user & admin accounts not owned by the tester are out of scope as well.

SECURITY.md
```diff
@@ -0,0 +1,32 @@
...
+Please report any security bugs in third-party projects to the person or team maintaining that project.
+
+The following are out of scope and should **not** be performed:
+
+* Excessive Automated Scans
+* Denial of Service Attacks
+* Social Engineering Attacks
+* Reports against infrastructure outside our control
+
+## Disclosure Policy
```
ghbjklhv: This does not seem to include if users will be informed in case data is leaked. 

My understanding is that social.privacytools.io and other services collect information like email.
What happens if this information gets stolen? How would users be informed?

JonahAragon: Added info to ad344be

Issue #982, opened 2019-06-10
🆕 Software Suggestion | Mojeek Search Engine


FinnBrownbill:
## Basic Information

**Name:** Mojeek
**Category:** Search Engine
**URL:** https://www.mojeek.com

## Description

Mojeek was the first search engine ever to state in their privacy policy that they don’t track their users, and have been committed to it ever since. Link to original privacy policy from March 2006 - https://web.archive.org/web/20060318104627/http://www.mojeek.com/privacy.html

Mojeek follows a strict no tracking privacy policy. Mojeek does not possess any previous identifying information on you, such as IP addresses, search history or click behaviour. More details can be found in the privacy policy: https://www.mojeek.com/about/privacy.html 

A few privacy metasearch engines are currently listed, but Mojeek is an independent crawler based search engine with its own algorithm and index of web pages. With an index of over 2 billion pages and plans in place to quadruple the size within the next 12-24 months, it is attempting to be a true genuine alternative. This is where those currently listed on privacytools.io get their organic results from: Startpage uses Google results, DDG mainly uses Bing, Qwant uses Bing for their English results, and Searx can use a variety of search engines but mainly Bing.   

About page: https://www.mojeek.com/about

blacklight447-ptio:
While I generally am not a fan of UK based services, do to their creepy government, actully  having a private search engine with an indepent engine is attractive thing to have

danarel:
The independent engine is a big sell. The search results kind of suck (subjective), but I would worry this would draw the same criticism the site gets for listing DDG: It seems to be closed source and bad location for privacy laws.

I could see this in worth mentioning though.

blacklight447-ptio:
@JonahAragon  @Mikaela @BurungHantu1605 some thoughts?

Mikaela:
I am not familiar with Mojeek, but if you checked the links, I can agree with the previous commentors and am fine with it being added as worth mentioning.

JonahAragon:
I'd approve a Worth Mentioning PR, but I don't know enough about it to make it a Top 3 recommendation.

blacklight447-ptio:
After some searching, i haven't been able to find anything especially shady, id say add as worth mentioning.

erciccione:
Closed source and based in a country with very restrictive privacy policies, this 2 should be enough to not list it. Also, I think would be a big mistake to list this search engine with so few research. Basically it's being added as "worth mentioning" only based on what they say on their website.

[...]

PR #1003, opened 2019-06-21
search-engines: add Mojeek to worth mentioning

Mikaela:
Resolves: #982


JonahAragon:
> The only ~~possible~~ **listed** exception is stating that their are no viable alternatives.

Seems like a fair arguement for a worth mentioning inclusion: https://github.com/privacytoolsIO/privacytools.io/issues/982#issuecomment-500601057

Mikaela:
@privacytoolsIO/editorial What do you think about the two NACKs?

Mikaela:
I think I am going to wait for @privacytoolsIO/editorial here.

erciccione:
So, this got merged beside the 2 nacks with no explanation/further discussion? One member jumps in and merge a PR without addressing the concerns or give any sort of explanation?

I hope this is not the normal workflow of privacytools...

PR #562, opened 2018-11-01
[WIP] Utilize Riot Instead of Wire

asdfghjz:
### Description
Some of the many advantages to Riot/Matrix

- Distributed
- Does not require phone number
- Works over TOR
- Does not store unnecessary data.

### HTML Preview

*Replace [GITHUB_USERNAME] with your GitHub username and [BRANCH] with the branch name.*

https://htmlpreview.github.io/?https://github.com/asdfghjz/privacytools.io/blob/Riot/index.html

PR #1004, opened 2019-06-22
Remove Riot beta warning badge

Perelandra0x309:
## Description

Resolves: #none

Riot is no longer considered in beta:
https://matrix.org/blog/2019/06/11/introducing-matrix-1-0-and-the-matrix-org-foundation

> We are very excited to announce the first fully stable release of the Matrix protocol and specification across all APIs - as well as the Synapse 1.0 reference implementation which implements the full Matrix 1.0 API surface.
> 
> This means that after just over 5 years since the initial work on Matrix began, we are proud to have finally exited beta!! This is the conclusion of the work which we announced at FOSDEM 2019 when we cut the first stable release of the Server-Server API and began the Synapse 0.99 release series in anticipation of releasing a 1.0.

### DIFF
--- a/_includes/sections/instant-messenger.html
+++ b/_includes/sections/instant-messenger.html
@@ -24,7 +24,7 @@ linux=""
 {% include cardv2.html
 title="Riot.im"
 image="/assets/img/tools/Riot.png"
-description="Riot.im is a decentralized free-software chatting application based on the <a href\"https://matrix.org/\">Matrix</a> protocol, a recent open protocol for real-time communication offering E2E encryption. It can bridge other communications via others protocols such as IRC too. <span class=\"badge badge-warning\" data-toggle=\"tooltip\" title=\"The software is currently in beta and the mobile client states 'End-to-end encryption is in beta and may not be reliable. You should not yet trust it to secure data.'\">beta <i class=\"far fa-question-circle\"></i></span>"
+description="Riot.im is a decentralized free-software chatting application based on the <a href\"https://matrix.org/\">Matrix</a> protocol, a recent open protocol for real-time communication offering E2E encryption. It can bridge other communications via others protocols such as IRC too."
 website="https://riot.im/"
 forum="https://forum.privacytools.io/t/discussion-riot-im/665"
 github="https://github.com/vector-im"


BearerOfHonesty:
While Riot.im is the reference client implementation it is neither the Matrix specification or the reference server implementation (Synapse).

BearerOfHonesty:
> While Riot.im is the reference client implementation it is neither the Matrix specification or the reference server implementation (Synapse).

Maybe I should have read things a bit more closely. Riot is supposedly out of beta but using it sure doesn't feel that way not that that's an argument against removing the badge.

Perelandra0x309:
Yes it would have been more correct for me to write Riot/Matrix is out of beta, considering the entire system. Riot itself has been out of beta for a while but Matrix not until recently.

BearerOfHonesty:
> I don't think this can be merged until [vector-im/riot-web#6779](https://github.com/vector-im/riot-web/issues/6779) is resolved as the warning refers to it.
Wouldn't that more correctly be needing an experimental badge instead?

Mikaela:
Sure, I would be fine with changing it to experimental even if it would hopefully soon be removed if the Matrix/Riot people aren't being too optimistic.

erciccione:
I would still leave the warning. As @Mikaela pointed out, many privacy related concerns are still unresolved.

Perelandra0x309:
Here's a good update on what the developer considers outstanding issues.
https://github.com/privacytoolsIO/privacytools.io/pull/562#issuecomment-502681069

Perhaps it is best to change the badge text and tooltip to something besides beta, which at this point is not correct.  It could be a "Warning" and a tooltip about E2EE having these issues from the above comment:

> E2E search (which is almost done, as of last week)

This seems like just an ease of use issue, not a privacy concern.

> Cross-signing to simplify verification (demoable as of last week)

Another ease of use issue which will be nice to have.

> Support for non-E2E clients via a shim (working as of a few weeks ago).

This is a new feature which allows some intermediate server to act as the E2E endpoint that a non E2E enable client connects to.

> Metadata is stored on servers. Fixing this is a long term project.

This is worth mentioning as a warning.

> We need to better at helping users understand where their contact data goes, and hash contact data during lookups

Another thing worth mentioning.

PR #1007, opened 2019-06-23
Switch Riot Beta tag to Experimental

JonahAragon:
Closes #1004 (this PR is an alternative modification): Riot is out of beta. Replaces beta tag with experimental tag:

image

More info links to this comment: https://github.com/privacytoolsIO/privacytools.io/pull/562#issuecomment-502681069

### DIFF
--- a/_includes/sections/instant-messenger.html
+++ b/_includes/sections/instant-messenger.html
@@ -24,7 +24,7 @@ linux=""
 {% include cardv2.html
 title="Riot.im"
 image="/assets/img/tools/Riot.png"
-description="Riot.im is a decentralized free-software chatting application based on the <a href\"https://matrix.org/\">Matrix</a> protocol, a recent open protocol for real-time communication offering E2E encryption. It can bridge other communications via others protocols such as IRC too. <span class=\"badge badge-warning\" data-toggle=\"tooltip\" title=\"The software is currently in beta and the mobile client states 'End-to-end encryption is in beta and may not be reliable. You should not yet trust it to secure data.'\">beta <i class=\"far fa-question-circle\"></i></span>"
+description="Riot.im is a decentralized free-software chatting application based on the <a href\"https://matrix.org/\">Matrix</a> protocol, a recent open protocol for real-time communication offering E2E encryption. It can bridge other communications via others protocols such as IRC too. <span class=\"badge badge-warning\" data-toggle=\"tooltip\" title=\"End-to-End encryption is not enabled by default and there are some concerns about metadata storage in Matrix, but it is generally acceptable for encrypting message content.\">experimental <i class=\"far fa-question-circle\"></i> (<a href=\"https://github.com/privacytoolsIO/privacytools.io/pull/562#issuecomment-502681069\">more info</a>)</span>"
 website="https://riot.im/"
 forum="https://forum.privacytools.io/t/discussion-riot-im/665"
 github="https://github.com/vector-im"


Mikaela:
Is it possible to have links in tooltips or could there be footnotes pointing to the Riot issue on finalizing E2EE and either to our issue on Riot being recommended instead of Wire or our official forum thread on Riot where I collected links to different issues and threads about it or even The Grid essay on how information is leaked to Matrix.org/Vector.im even when selfhosting?

*Good night, I may be having a busy day tomorrow*

Perelandra0x309:
The experimental tag isn't really correct either.  Something better might be "Caution" since the purpose is to just point out that care must be taken to enable E2EE and be aware of meta data issues.  Experimental actually seems like a worse tag than Beta, since beta usually means almost ready to make final and just doing final testing to find any remaining major bugs.  Experimental is more like this project can break at any time.  Matrix was officially "Beta" until recently so that was appropriate until a few weeks ago.  I haven't seen anything calling this project experimental.

_includes/sections/instant-messenger.html
```diff
@@ -24,7 +24,7 @@ <h1 id="im" class="anchor"><a href="#im"><i class="fas fa-link anchor-icon"></i>
...
+description="Riot.im is a decentralized free-software chatting application based on the <a href\"https://matrix.org/\">Matrix</a> protocol, a recent open protocol for real-time communication offering E2E encryption. It can bridge other communications via others protocols such as IRC too. <span class=\"badge badge-warning\" data-toggle=\"tooltip\" title=\"End-to-End encryption is not enabled by default and there are some concerns about metadata storage in Matrix, but it is generally acceptable for encrypting message content.\">experimental <i class=\"far fa-question-circle\"></i> (<a href=\"https://github.com/privacytoolsIO/privacytools.io/pull/562#issuecomment-502681069\">more info</a>)</span>"
```
Mikaela: I think https://github.com/vector-im/riot-web/issues/6779 could be a better link for more info.

five-c-d:
Second the opinion, that "experimental" is very much the wrong word to use.  

*  back-of-the-napkin
*  work-in-progress
*  experimental 
*  proof-of-concept
*  alpha-stage
*  beta-stage
*  release-candidate 
*  v1.0 
*  v2.0 
*  mature 

Riot+MegOlm has been in the "late beta" aka release candidate stage for a long time now, and Riot (but not yet MegOlm which is still off-by-default) has moved to the v1.0 stage officially.  Calling it experimental would give me the wrong idea, that it was just barely past the work-in-progress (aka mostly broken) stage and had made it to the partly-working stage

There is a different classification system used for "things we want the readership to be aware of" which is somewhat distinct from the above, and follows the classic software-severity-system of identifying error-messages and such:  

*  danger  (most severe) 
*  warning (pretty severe) 
*  note   (worth highlighting) 
*  explicitly stated (but not highlighted in color) 

I think these severity-ratings can be used by themselves to highlight worries and concerns that privacy-aware endusers might want to be cognizant about.  

* Danger:  encryption is off-by-default. 
* Warning:  stores a lot of metadata. 
* Note:  firm is based in the UK which is a FiveEyes country. 

But I also think we can combine the dev-stage labels with the severity-ratings:  

* Danger:  experimental 
* Warning:  early beta-stage 
* Note:  software only recently moved from late-beta to v1.0 development-stage.  

Depending on who the readership of the site is, would shift the points at which the shift from danger-to-warning occurs, and the shift from warning-to-note occurs.  Everyday endusers should typically not be using v1.0 software, other things being equal, so that might actually merit 'warning: software is not yet mature' type of label be slapped on it.  Anything that is not a *very* stable beta, should probably not be recommended to everyday endusers at all, and thus would merit the 'danger: software in an early and unstable stage of development' would be put on the late-alpha and early-beta stuff ... or those things would just NOT be counted as WorthMentioning *until* they achieved a modicum of stability, more likely.

Issue #1014, opened 2019-07-04
Mullvard VPN - Free Trial


C-Hodor:
Mullvard VPN does not offer a free trial account anymore, remove that feature from the vpn's list.
Cheers.

PR #1015, opened 2019-07-04
Update vpn.html

Offpics:
Resolves: #1014

Change Mullvad's free trial value.


jkhgvfgvsth:
@Mikaela Mullvad does have a free trial.

Click ""get account" to see more information. Then go to "https://mullvad.net/en/account/" and it will show you when it expires. I think it is but a few hours tho.

Mikaela:
Do you consider a few hours enough long to revert this PR?

Ping @privacytoolsIO/editorial sorry, I seem to have failed.

Offpics:
> @Mikaela Mullvad does have a free trial.
> 
> Click ""get account" to see more information. Then go to "https://mullvad.net/en/account/" and it will show you when it expires. I think it is but a few hours tho.

There is no such thing. After creating an account you have to pay for the service.

Issue #1018, opened 2019-07-05
💬 Discussion | Add Related issues to Minds


jkhgvfgvsth:
**Short Description**: Add Related Issues for Minds
**Longer Description**: Minds has many differences to the other listed in "Decentralized Social Networks", I assume this is why it is listed under worth mentioning and not main.

I propose the concept of adding related issues to help explain the differences in Minds, not necessarily making it worse but to better inform users. For instance, you can't really self-host Minds (yet).
Minds is also more active and has a significantly cleaner UI than alternatives like Diaspora or Friendica.

Some Issues to Discuss:
[Mind's Supporting Activity Pub isn't yet Active](https://gitlab.com/minds/engine/issues/307)

[Federated & Decentralized](https://gitlab.com/minds/engine/issues/183)

Mikaela:
> Minds - An open source and distributed social networking service, integrating the blockchain to reward the community.

How would you propose adding the issue links to that? I wonder if worth mentioning should also have their official forum threads (@BurungHantu1605)?

jkhgvfgvsth:
@Mikaela I figured you could add a warning badge like:

`<span class="badge badge-warning" data-toggle="tooltip" title="This software has some related information you may be interested in like <a href="https://gitlab.com/minds/engine/issues/183">#183</a> or <a href="https://gitlab.com/minds/engine/issues/307">#307</a>.">notes <i class="far fa-question-circle"></i></span>`

Mikaela:
I fear that tooltips cannot have links or they will be unclickable at least from desktop browsers. With the description I worry about it looking inconsistent, but I admit it's less of an issue than not having this information.

Would you like to send a PR?

jkhgvfgvsth:
@Mikaela :+1:

jkhgvfgvsth:
@Mikaela Would it be better to make "notes" redirect here?

blacklight447-ptio:
updates on this issue? is there any progress?

Mikaela:
It looks like @BurungHantu1605 rejected the PR without explanation.

blacklight447-ptio:
so what are we doing here, closing it?

Mikaela:
I guess

PR #1019, opened 2019-07-06
Add Related issues to Minds

jkhgvfgvsth:
Part of #1018


Mikaela:
Sorry, I don't feel comfortable merging this without an second opinion, so we will have to wait.

PR #1023, opened 2019-07-11
instant-messenger: add danger about Riot's privacy concerns

Mikaela:
Ref: #1007 #1004 (need changes if this is merged) #840 (some discussion
on the concerns).


Mikaela:
(I got a better idea #1024)

Issue #995, opened 2019-06-16
🆕 Software Suggestion | Add Yggdrasil to Self-contained Networks


oijkhg:
**Description**: Add Yggdrasil to Self-contained Networks
**Why?**: The current internet is mostly proprietary and owned by a few monopolies.
These monopolies have been known to over-exert their influence (at least where I am).
Yggdrasil may help us fix this with creating a FLOSS end-to-end encrypted IPv6 network.
**Does it protected your IP**: For the most part, no. But, ideally networks like i2p should work.
**Had this been discussed before?** This issue is made to revive #685 
**Links**:
https://yggdrasil-network.github.io/

https://www.reddit.com/r/darknetplan

BearerOfHonesty:
https://en.wikipedia.org/wiki/Unlicense#Reception sounds like CC0 might be a better choice even if it's more legalese.

Mikaela:
Has our opinion changed half an year later @privacytoolsIO/editorial @privacytoolsIO/services (https://github.com/privacytoolsIO/privacytools.io/issues/139#issuecomment-489439098) @Shifterovich ?

I seem to have referred to https://github.com/privacytoolsIO/privacytools.io/issues/685#event-2042260299 a few times.

Labeling I2P and Tor for them being supported and appearing at [other public peers](https://github.com/yggdrasil-network/public-peers/tree/master/other) (where funnily I have made the latest commit at the time of writing) and VPN as I have referred to it from so many VPN issues and it's the only VPN I am using. I seem to have 8 alive hosts (including those currently offline).

Mikaela:
Assigning myself as I care about this issue and am interested in making a PR to move this forwards at least a bit unless someone comes to close this.

By assigning I mean to only communicate that I care about the issue and I welcome others to submit the PR or say that they wish to do it by themselves, I don't intent to deserve exclusive rights on it.

Mikaela:
@oijkhg What do you think of https://github.com/privacytoolsIO/privacytools.io/pull/1035/files ?

PR #1035, opened 2019-07-17
self contained networks: add Yggdrasil as worth mentioning

Mikaela:
Resolves: #995


_includes/sections/self-contained-networks.html
```diff
@@ -49,4 +49,5 @@ <h3>Worth Mentioning</h3>
...
   <li><a href="https://ipfs.io/">IPFS</a> - A peer-to-peer hypermedia protocol to make the web faster, safer, and more open. (<a href="https://github.com/privacytoolsIO/privacytools.io/pull/361#issuecomment-344414022">important privacy warning</a>)</li>
+  <li><a href="https://yggdrasil-network.github.io/">Yggdrasil network</a> - Yggdrasil is an early-stage implementation of a fully end-to-end encrypted IPv6 network. It is lightweight, self-arranging, supported on multiple platforms and allows pretty much any IPv6-capable application to communicate securely with other Yggdrasil nodes. Yggdrasil does not require you to have IPv6 Internet connectivity - it also works over IPv4. <span class="badge badge-warning" data-toggle="tooltip" title="The project is currently in early stages but it is being actively developed.">experimental <i class=\"far fa-question-circle\"></i></span> <span class="badge badge-danger" data-toggle="tooltip" title="Yggdrasil doesn't have a goal of providing anonymity and your peers know your IP address unless you are only using Tor/I2P peers.">privacy warning <i class="far fa-question-circle"></i></span></li>
```
nitrohorse: Small suggestion for consistency: drop "network" from the name since the project looks like it's simply called, "Yggdrasil", and start the description as "An early-stage..."

nitrohorse:
Minor grammer suggestions but otherwise looks good to me.

_includes/sections/self-contained-networks.html
Mikaela: Done, I tend to call it as network, because sometimes older Linux users appear to say that Yggdrasil is a discontinued Linux distribution from 1995 (older than me) :smile: 

https://en.wikipedia.org/wiki/Yggdrasil_Linux

Issue #682, opened 2018-12-22
🆕 Software Suggestion | Ecosia search engine


Mikaela:
## Basic Information

**Name:** Ecosia
**Category:** Search engine
**URL:** https://www.ecosia.org/
**About URL** https://info.ecosia.org/

## Description

Ecosia is an search engine planting trees with the money they make. They appear to be transparent releasing financial documents at https://documents.ecosia.org/467540 (however I didn't read any).

Their privacy policy looks OK to me, https://info.ecosia.org/privacy . However their results come from Bing, but that is also one of the sources of DuckDuckGo.

asddsaz:
They have a decent privacy policy: https://tosdr.org/#ecosia
But are non-free, therefore don't meet the software criteria.

Mikaela:
DuckDuckGo and StartPage.com are also non-free as far as I can see?

Mikaela:
Oh, you opened https://github.com/privacytoolsIO/privacytools.io/pull/699 and forgot to mention it here.

libBletchley:
Ecosia has always been hostile toward Tor users which makes it a non-starter from a privacy viewpoint.

ATM it's just dropping packets from Tor and timing out, which IMO is even less diplomacy than a "403 forbidden".

Mikaela:
https://github.com/privacytoolsIO/privacytools.io/pull/960#issuecomment-498073312 ?

erciccione:
Why reopen this @Mikaela ?
From the comment of the ghost user looks like Ecosia is not Tor-friendly, which IMO is a big "no no" if we are talking about user privacy. I don't think that them being non-free was the stopper here.

Mikaela:
I don't think we have a formal rule of search engines needing to be Tor friendly in order to be listed. To [quote myself from the forum](https://forum.privacytools.io/t/what-is-your-daily-search-engine/174/7?u=mikaela):

> Ecosia, because I am more worried about the climate change than their Tor hostility.

What do you think @privacytoolsIO/editorial ?

erciccione:
I'm confused now. Isn't this **Privacy**tools.io? Shouldn't this community suggest the best tools to protect people's privacy?

> I don't think we have a formal rule of search engines needing to be Tor friendly in order to be listed

You don't need a formal rule. Supporting Tor is a big point if you are proposing a privacy-oriented browser, planting trees is not...

[...]

PR #1036, opened 2019-07-17
Search engines: add Ecosia

Mikaela:
Resolves: #682

The issue above is in disagreement and requires attention on whether Tor
blocking is a reason to not merge this.

### DIFF
--- a/_includes/sections/search-engines.html
+++ b/_includes/sections/search-engines.html
@@ -58,4 +58,5 @@ github="https://github.com/Qwant/"
   <li><a href="https://jivesearch.com/">Jive Search</a> - A free software search engine with a similar look and feel to Google.</a></li>  
   <li><a href="https://metager.de/en/">MetaGer</a> - An open source metasearch engine, which is based in Germany. It focuses on protecting the user's privacy.</li>
   <li><a href="https://www.mojeek.com/">Mojeek</a> - Independent and unbiased search results with no user tracking.</li>
+  <li><a href="https://www.ecosia.org/">Ecosia</a> - <span class="badge badge-warning" data-toggle="tooltip" title="Ecosia is reported to be blocking Tor nodes.">warning <i class="far fa-question-circle"></i></span> Search engine that plants trees.</li>
 </ul>


nitrohorse:
Hmm so...
1) I really do like that they use their profits to plant trees,
2) that their privacy policy is short and reasonably respectful for users, and
3) that they honor the Do Not Track browser setting.

On the other hand, in my perspective...

1) it's unfortunate Ecosia doesn't proxy requests to Bing like Startpage does to Google. Requests to Bing.com and Bing.net are made when searching. (Note: I also noticed (at least) on their [iOS app](https://apps.apple.com/us/app/ecosia/id670881887) that additional requests are made to "a-0001.a-afdentry.net.trafficmanager.net" (a [CNAME for Bing.com](https://urlscan.io/domain/www.bing.com)) and "dual-a-0001.a-msedge.net")
2) I wish they provided an option [in their settings page](https://www.ecosia.org/settings) (like [Startpage](https://eu.startpage.com/search/settings) and [DuckDuckGo](https://duckduckgo.com/settings) do) for using POST or GET when making the search request. Both have their privacy trade-offs but Ecosia locks users into using GET.
3) Looking at Ecosia's statement here:

> "...we forward the following information to our partner, Bing: IP address, user agent string, search term, and some settings like your country and language setting." -- https://info.ecosia.org/privacy

Besides Ecosia's tree-planting mission, I wonder if there is a significant privacy-gain over directly using Bing. Since requests from Ecosia are "forwarded" to Bing, is it correct to assume my queries on Bing's side fall under [Microsoft's user-hostile privacy policy](https://privacy.microsoft.com/en-us/privacystatement)?

Mikaela:
@nitrohorse so shall we close the Ecosia case *permanently*? "We love trees and planting them, but due to some concerns we won't be able to list it" ?

nitrohorse:
@Mikaela yeah, I think we should.

Mikaela:
Case closed, see comments of #682 and comments starting from https://github.com/privacytoolsIO/privacytools.io/pull/1036#pullrequestreview-263991424.

bcye:
Hey @nitrohorse and @Mikaela ,

I just tried using Ecosia on my computer via the TOR browser and it just worked fine. It is still not proxying search requests but sending them to bing, but I think this should not be a deal-breaker and that Ecosia should be at least added as a "Worth Mentioning".

CleanShot 2020-04-30 at 19 52 40

nitrohorse:
Hey @bcye, thanks for reaching out. I don't think my position has changed. Ecosia forwards requests to Bing which logs queries. And then "after 7 days [Ecosia] remove[s] all personal information (like the IP address) from [stored] search data." They take steps toward privacy but it's not as _high_ a priority like other search engines.

From my understanding, this is also unfortuanely the case for other environment-forward search engines that forward requests with personal information to Bing/Google such as [Ekoru](https://www.ekoru.org), [Lilo](https://search.lilo.org/), [OceanHero](https://oceanhero.today/), [YouCare](https://youcare.world), [Givero](https://www.givero.com/), and [Gexsi](https://gexsi.com/en).

Issue #1039, opened 2019-07-20
🆕 Software Suggestion | Snowflake extension


Mikaela:
## Basic Information

**Name:** Snowflake
**Category:**  Excellent Firefox Privacy Add-ons
**URL:** https://snowflake.torproject.org/

## Description

> Snowflake is a system to defeat internet censorship. People who are censored can use Snowflake to access the internet. Their connection goes through Snowflake proxies, which are run by volunteers. For more detailed information about how Snowflake works see our documentation wiki.

Snowflake extension turns users into Snowflake proxies letting censored Tor users use them as a bridge to Tor network.

See also https://github.com/privacytoolsIO/privacytools.io/issues/295 https://trac.torproject.org/projects/tor/ticket/23888

Mikaela:
Quoting a Matrix discussion:

> I guess it should be on a "please consider." And also to check if there's any possible way for the extention to be more harmful than not using it, in which case there should be a warning.

*I don't think there is much room for harm, snowflakes are more temporary than usual bridges and the usual bridges aren't publicly listed either*

> Well, fingerprinting, costing bandwidth, idk

I guess bandwidth cost may be a real issue, so this should probably be mentioned in case someone installed it on a phone.

* https://matrix.to/#/!tYmNuaXMRnHEZWdDqp:privacytools.io/$156364157721379ALURj:privacytools.io?via=privacytools.io&via=matrix.org&via=librem.one

nitrohorse:
Quoting what I mentioned in Riot:
> I can see some confusion from users who follow [PTIO's guide to disable WebRTC](https://www.privacytools.io/browsers/#webrtc) and then see they'd need to revert these steps for the extension to work.

But I do like the idea of putting this extension in a new "Please Consider"-type of category under the "For Experts Only" section since its use is to help _others_ achieve privacy who potentially can't by other means. So in that regard it does seem to stand on its own.

JonahAragon:
This is pretty sweet actually, I'm installing it now.

I don't think this needs to be in a separate category (unless there are many more similar tools that would fit in the category I suppose). We can just include this in "Firefox Privacy Add-ons", especially because that category in particular has long descriptions for each recommended items where we can explain essentially what's been said here: Snowflake doesn't provide privacy benefits to your own browser, rather it enables users with hostile ISPs to access the Tor network anonymously.

[...]

PR #1040, opened 2019-07-21
Browser Addons: Add Snowflake

nitrohorse:
## Description

Resolves: https://github.com/privacytoolsIO/privacytools.io/issues/1039

## Screenshot

snowflake-screen


_includes/sections/browser-addons.html
```diff
@@ -82,6 +82,18 @@ <h1 id="addons" class="anchor"><a href="#addons"><i class="fas fa-link anchor-ic
 safari=""
 %}
 
+{% include cardv2.html
+title="Snowflake"
+image="/assets/img/addons/snowflake.png"
+description="<strong>Snowflake</strong> is a system to defeat internet censorship. People who are censored can use Snowflake to access the internet. Their connection goes through Snowflake proxies, which are run by volunteers. If your internet access is not censored, you should consider installing the Snowflake extension to help users in censored networks. There is no need to worry about which websites people are accessing through your proxy. Their visible browsing IP address will match their Tor exit node, not yours."
+website="https://snowflake.torproject.org"
+forum="TODO"
```
nitrohorse: Need a forum link before this can be merged.

_includes/sections/browser-addons.html
```diff
@@ -82,6 +82,18 @@ <h1 id="addons" class="anchor"><a href="#addons"><i class="fas fa-link anchor-ic
...
+github="https://gitweb.torproject.org/pluggable-transports/snowflake.git"
+firefox="https://addons.mozilla.org/en-US/firefox/addon/torproject-snowflake"
```
nitrohorse: I'm also realizing that the browser icons aren't selectable (maybe they should be?) so I'll remove the firefox and chrome links.

_includes/sections/browser-addons.html
Mikaela: https://forum.privacytools.io/t/discussion-snowflake/1146 assuming there is no special magic in them

_includes/sections/browser-addons.html
```diff
@@ -82,6 +82,18 @@ <h1 id="addons" class="anchor"><a href="#addons"><i class="fas fa-link anchor-ic
 safari=""
 %}
 
+{% include cardv2.html
+title="Snowflake"
+image="/assets/img/addons/snowflake.png"
+description="<strong>Snowflake</strong> is a system to defeat internet censorship. People who are censored can use Snowflake to access the internet. Their connection goes through Snowflake proxies, which are run by volunteers. If your internet access is not censored, you should consider installing the Snowflake extension to help users in censored networks. There is no need to worry about which websites people are accessing through your proxy. Their visible browsing IP address will match their Tor exit node, not yours."
```
Mikaela: >  Snowflake Snowflake is a system to defeat internet censorship. People who are censored can use Snowflake to access the internet. Their connection goes through Snowflake proxies, which are run by volunteers. If your internet access is not censored, you should consider installing the Snowflake extension to help users in censored networks. There is no need to worry about which websites people are accessing through your proxy. Their visible browsing IP address will match their Tor exit node, not yours. 

I think this explanation should say more clearly that it's directly associated with Tor and lets users for whom Tor is blocked access the Tor network. And maybe there would also need to be a mention that it may use bandwidth?

nitrohorse: Hmm, good callout.

nitrohorse: We could add the words in bold text to address some of your points:

> Snowflake is a system **from the Tor project** to defeat internet censorship. People who are censored can use Snowflake to access the internet **through the Tor network.** Their connection goes through Snowflake proxies, which are run by volunteers. If your internet access is not censored, you should consider installing the Snowflake extension to help users in censored networks. There is no need to worry about which websites people are accessing through your proxy. Their visible browsing IP address will match their Tor exit node, not yours.

JonahAragon: How about:

> Snowflake is a system from the Tor Project to help people with censored connections access the internet via the Tor network. If you have an unsensored connection, running this extension volunteers your connection to be used as a Snowflake proxy to help users on censored networks. There is no need to worry about which websites people are accessing through your proxy. Their visible browsing IP address will match their Tor exit node, not yours. For other users, this extension will not add privacy or uncensor connections on its own, you should use the [Tor Browser](https://www.torproject.org/) instead.

nitrohorse: From the [Tor project list post](https://lists.torproject.org/pipermail/tor-dev/2016-January/010310.html), it's called out that Snowflake proxies are "short-lived" but I can't seem to find any clear documentation regarding bandwidth...

nitrohorse: How about this?
> Snowflake is a system **from the Tor project** to defeat internet censorship. People who are censored can use Snowflake to access the internet **through the Tor network**. Their connection goes through Snowflake proxies, which are run by volunteers. If your internet access is not censored, you should consider installing the Snowflake extension to help users in censored networks. There is no need to worry about which websites people are accessing through your proxy. Their visible browsing IP address will match their Tor exit node, not yours. **Also keep in mind that while Snowflake proxies are short-lived, they do use bandwidth.**

Mikaela: I think it's more of to access the Tor network through censorship rather than the internet, so the quotes are misleading.

> Also keep in mind that while Snowflake proxies are short-lived, they do use bandwidth.

I think we have a confusion of point of view here. From client point of view it's short-lived connection and may not use that much bandwidth, but what is there to prevent the same host from being consecutively proxying traffic from different user?

nitrohorse: > but what is there to prevent the same host from being consecutively proxying traffic from different user?

Ah yeah, good point; I assume nothing would prevent that so using "short-lived" in this context is also misleading..

nitrohorse: [Sync'd on Riot](https://matrix.to/#/!MGdJmTKoklgdThFeeL:privacytools.io/$156373647423625TgrcU:privacytools.io?via=privacytools.io&via=matrix.org) and will update the description to:
> Snowflake is a new pluggable transport from the Tor Project. If you have an uncensored connection, running this extension volunteers your connection to be used as a Snowflake proxy to help users unable to connect to the Tor network. Your IP will not be visible to the sites users visit using your proxy, as this extension will not make you an exit node. If your access to the Tor network is being blocked, this extension will not assist you, and you should use the Tor Browser instead.

PR #1042, opened 2019-07-22
Update source_code.md

ggg27:
Add TOS;DR and Snowflake to Source_Code.md

### DIFF
--- a/source_code.md
+++ b/source_code.md
@@ -27,6 +27,10 @@ https://github.com/privacytoolsIO/
  
  Decentraleyes: https://git.synz.io/Synzvato/decentraleyes
  
+ Terms of Service; Didn’t Read : https://github.com/tosdr/
+ 
+ Snowflake : https://gitweb.torproject.org/pluggable-transports/snowflake.git
+ 
  uMatrix: https://github.com/gorhill/uMatrix/
  
  NoScript: https://github.com/hackademix/noscript/


source_code.md
```diff
@@ -27,6 +27,10 @@ https://github.com/privacytoolsIO/
  
  Decentraleyes: https://git.synz.io/Synzvato/decentraleyes
  
+ Terms of Service; Didn’t Read : https://github.com/tosdr/
+ 
+ Snowflake : https://gitweb.torproject.org/pluggable-transports/snowflake.git
```
Mikaela: I am not entirely sure this is the correct place.

nitrohorse:
Thanks for catching this, @ggg27; didn't realize I need to update this file for changes.

source_code.md
nitrohorse: I think this is the right place from what I can tell. It looks like the order matches the order on PTIO.

Mikaela: I mean is that the source code for the browser extension too?

nitrohorse: Oh gotcha; yeah this was the same link I found in the [docs](https://trac.torproject.org/projects/tor/wiki/doc/Snowflake).

nitrohorse: Wow I misunderstood again, sorry 😓 hmm I can’t seem to find the extension source code myself 🤔 Maybe it’s not open-source at the moment...

nitrohorse: This is the same URL in the Addons page so I think once the extension itself becomes open-source we should update both locations.

PR #1048, opened 2019-07-24
Recommend Matrix instead of Riot

David-Beetle:
**Description**: Recommend Matrix instead of Riot
**Why?**: Riot isn't the only Matrix client. Others may offer better usability, like CLI access.


David-Beetle:
@Mikaela It is very similar to how we recommend XMPP.

I still recommend Riot and Weechat in the description.

Mikaela:
I had to check how the XMPP recommendation is currently at https://www.privacytools.io/software/im/ 

Screenshot_2019-07-24 Instant Messaging privacytools io

I guess XMPP should be a link, but there are clients clearly mentioned while the quick glance at this PR I didn't see any clients.

I don't support listing WeeChat, because it's primarily IRC client and the Matrix support is a script that AFAIK isn't even in [the official scripts repo](https://github.com/weechat/scripts), if it was a plugin, it would be almost equal to IRC and I would be open to listing it as a plugin, but currently it would be "for advanced users only" and that section doesn't exist with instant messengers at this time.

David-Beetle:
@Mikaela I personally like Weechat. However, if I'm the only one with this perspective I see no reason for privacytools.io to recommend it.

Mikaela:
> I personally like Weechat. However, if I'm the only one with this perspective I see no reason for it not to be removed.

I am fed up with the discussion whether or not WeeChat is a Matrix protocol client or a IRC protocol client, so I have asked authoritative source about this to end these discussions. https://github.com/weechat/weechat/issues/1386

Mikaela:
I got the impression that WeeChat does not support Matrix officially.

PR #1052, opened 2019-07-25
instant-messenger.html: make XMPP and OMEMO links

Mikaela:


Mikaela:
Oh, I can merge my own PRs as long as they have been approved by someone else. I think this is so non-controversial that I will merge, thanks for the review @nitrohorse

Issue #1057, opened 2019-07-25
🌐 Website Issue | Instant messengers: Gajim is not Linux-only


Mikaela:
However I am currently unable to open https://gajim.org/ so I cannot provide the correct platforms at the moment.

* https://www.privacytools.io/software/im/

nitrohorse:
Huh, that's interesting. It's down for me too.

nitrohorse:
[Looks like](https://gajim.org/downloads.php?lang=en) Gajim is available also on Windows.

![gajim](https://user-images.githubusercontent.com/1514352/62014170-435a3080-b18d-11e9-8a74-2508a62abc30.png)

Mikaela:
![Screenshot_2019-07-29 Gajim, an XMPP client](https://user-images.githubusercontent.com/831184/62033107-c5247b00-b1da-11e9-9e56-65231cd6dd1e.png)

hah, we took the same screenshot of https://gajim.org/downloads.php?lang=en

Anyway I would say FreeBSD, Linux, Windows based on that and will open a PR soon.

PR #1063, opened 2019-07-29
instant-messenger: correct supported platforms of Gajim

Mikaela:
Resolves: #1057

### DIFF
--- a/_includes/sections/instant-messenger.html
+++ b/_includes/sections/instant-messenger.html
@@ -69,7 +69,7 @@ web=""
   <ul>
     <li><a href="https://monal.im/">Monal</a> (iOS, MacOS) - An XMPP client in active development.</li>
     <li><a href="https://conversations.im/">Conversations</a> (Android) - An open source Jabber/XMPP client for Android 4.4+ smartphones. Supports end-to-end encryption with either OMEMO or OpenPGP.</li>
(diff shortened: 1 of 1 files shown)


_includes/sections/instant-messenger.html
```diff
@@ -69,7 +69,7 @@ <h3>Worth Mentioning</h3>
   <ul>
     <li><a href="https://monal.im/">Monal</a> (iOS, MacOS) - An XMPP client in active development.</li>
     <li><a href="https://conversations.im/">Conversations</a> (Android) - An open source Jabber/XMPP client for Android 4.4+ smartphones. Supports end-to-end encryption with either OMEMO or OpenPGP.</li>
-    <li><a href="https://gajim.org/">Gajim</a> (Linux) - An open source fully featured XMPP client.</li>
+    <li><a href="https://gajim.org/">Gajim</a> (FreeBSD, Linux, Windows) - An open source fully featured XMPP client.</li>
```
Mikaela: I wonder if this is now the only place where we say that BSD exists in addition to Linux and Windows?

nitrohorse: We call out BSD in a number of places throughout our code -- like in the email and file encryption sections but looks like the first time in the IM section.

PR #1073, opened 2019-08-01
Update discussion issue template to direct new discussions to the forum

nitrohorse:
## Description

Related to a comment @Mikaela made in # 977.


nitrohorse:
> 👍 but GitHub says that this will close the whole issue even if you refer only to a single comment.

Oh, interesting... I could then just remove the linked comment from the PR description?

PR #1081, opened 2019-08-03
GTK+ was renamed to GTK

dawidpotocki:
GTK+ was renamed to just GTK, some time ago.

https://mail.gnome.org/archives/gtk-devel-list/2019-February/msg00000.html


nitrohorse:
LGTM

PR #1083, opened 2019-08-03
Update logos

dawidpotocki:
StartMail, StartPage, F-Droid, Purism, Thunderbird, Tor Browser

There can be more, but I only know about these.


dawidpotocki:
I think it was something with previous png settings as I took the same
image. Made a new png and should look fine now.

PR #1090, opened 2019-08-03
Add warnings to VPN section, encourage Tor use

### LINKED ISSUE #914: 💬 Discussion | should VPN-section recommend Tor-use, versus VPN-use?

JonahAragon:
# Description

This PR adds a warning to the VPN page detailing the drawbacks and risks of using a VPN, and suggesting the use of Tor to ensure anonymity in most cases.

https://deploy-preview-1090--privacytools-io.netlify.com/providers/vpn/

Closes #914

### DIFF
--- a/pages/providers/vpn.html
+++ b/pages/providers/vpn.html
@@ -1,8 +1,57 @@
 ---
 layout: page
 permalink: /providers/vpn/
-title: "VPN Services for Privacy and Security"
+title: "VPN Services"
 description: "Find a no-logging VPN operator who isn't out to sell or read your web traffic."
 ---
 
+<div class="card border-danger">
+  <div class="card-header text-danger"><i class="fas fa-exclamation-circle fa-fw"></i> Warning</div>
+  <div class="card-body">
+    <p class="card-text text-danger">Using a VPN will <strong>not</strong> keep your browsing habits anonymous, nor will it add additional security to non-secure (HTTP) traffic.</p>
+    <p class="card-text text-danger">If you are looking for <strong>anonymity</strong>, you should use the Tor Browser <strong>instead</strong> of a VPN.</p>
+    <p class="card-text text-danger">If you're looking for added <strong>security</strong>, you should always ensure you're connecting to websites using HTTPS. A VPN is not a replacement for good security practices.</p>
+    <p class="card-text text-secondary">If you're looking for additional <strong>privacy</strong> from your ISP, on a public Wi-Fi network, or while torrenting files, a VPN may be the solution for you as long as you understand <a href="#info">the risks involved</a>.</p>
+    <a href="https://www.torproject.org/" class="btn btn-danger">Download Tor</a>
+    <a href="https://write.privacytools.io/my-thoughts-on-security/slicing-onions-part-1-myth-busting-tor" class="btn btn-outline-danger">Tor Myths &amp; FAQ</a>
+    <a href="#info" class="btn btn-outline-secondary">More Info</a>
+  </div>
+</div>
+
 {% include sections/vpn.html %}
+
+<h1 id="info" class="anchor"><a href="#info"><i class="fas fa-link anchor-icon"></i></a> Further Information and Dangers</h1>
+
+<div class="container">
+  <div class="row">
+    <div class="col-md-6">
+      <h3>Should I use a VPN?</h3>
+      <p>The answer to this question is not a particularly helpful one: <strong>It depends.</strong> It depends on what you're expecting a VPN to do for you, who you're trying to hide your traffic from, and what applications you're using.</p>
+      <p><strong>In most cases, VPNs do little to protect your privacy or enhance your security</strong>, unless paired with other changes.</p>
+      <p>VPNs cannot encrypt data outside of the connection between your device and the VPN server. VPN providers can see and modify your traffic the same way your ISP could. And there is no way to verify a VPN provider's "no logging" policies in any way.</p>
+      <h3>What if I need encryption?</h3>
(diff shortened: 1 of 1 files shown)


pages/providers/vpn.html
```diff
@@ -1,8 +1,57 @@
...
+      <h3>Are VPNs ever useful?</h3>
+      <p>A VPN may still be useful to you in a variety of scenarios, such as:</p>
+      <ol>
+        <li>Hiding your traffic from <strong>only</strong> your Internet Service Provider.</li>
+        <li>Hiding your downloads (such as Torrents) from your ISP and anti-piracy organizations.</li>
+      </ol>
+      <p>For use-cases like these, or if you have another compelling reason, the VPN providers we listed above are who we think are the most trustworthy. However, using a VPN provider still means you're <em>trusting</em> the provider. In pretty much any other scenario you should be using a secure<strong>-by-design</strong> tool such as Tor.</p>
+    </div>
+  </div>
+  <div class="row">
```
nitrohorse: Suggestion: could we make this a list? (this may work)
```html
<div class="row">
    <div class="col">
        <p><strong>Sources and Further Reading</strong>:
            <ol>
                <li><a href="https://schub.io/blog/2019/04/08/very-precarious-narrative.html">VPN - a Very Precarious Narrative</a> by Dennis Schubert</li>
                <li><a href="https://gist.github.com/joepie91/5a9909939e6ce7d09e29">Don't use VPN services</a> by Sven Slootweg</li>
                <li><a href="https://write.privacytools.io/my-thoughts-on-security/slicing-onions-part-1-myth-busting-tor">Slicing Onions: Part 1 – Myth-busting Tor</a> by blacklight447</li>
                <li><a href="https://write.privacytools.io/my-thoughts-on-security/slicing-onions-part-2-onion-recipes-vpn-not-required">Slicing Onions: Part 2 – Onion recipes; VPN not required</a> by blacklight447</li>
            </ol>
        <p>
    </div>
</div>
```

Issue #1077, opened 2019-08-01
🌐 Website Issue | Add more encrypted DNS servers


Mikaela:
## Description

Currently all DNS servers have warnings and we only list 4. I had been wondering if there really are no others than what we list, Google and Cloudflare, but there is a long list. I have no idea where I picked my previous list and I apologize.

* https://dnsprivacy.org/wiki/display/DP/DNS+Privacy+Test+Servers

Short list of what may be worth looking at:

* https://securedns.eu/ - no logs, all three protocols, optional adblock (alternative domain)
* ~~http://neutopia.org/~~ - missing privacy policy
* ~~https://tenta.com/dns-setup-guides~~ - I am not sure on their policy of counting counters and their privacy policy has the "information we don't collect" section and they talk about supporting law enforcement whentalking about VPN, so I am not sure if the DNS should be listed either.
* https://appliedprivacy.net/services/dns/ - DoH & DoT
* https://blog.uncensoreddns.org/ - "Absolutely nothing is being logged, neither about the users nor the usage of this service." - DoT
* https://dnsprivacy.at/ - " NO REQUESTS ARE LOGGED! This is an experimental service, only on a "best effort" basis." - DoT only?
* https://github.com/bhanupratapys/dnswarden/blob/master/README.md - "Use at your own risk. I will not be held responsible for any downtime of the servers , only on a "best effort" basis. No query logging . Completely disabled query logs related to the DNS traffic." - all three

This is based on the dnsprivacy.org list, from which homepages or additional information I found very easily by just checking the main domain.

Mikaela:
As there are starting to be options, maybe regional ones can join the list too

* https://www.nic.cz/odvr/ - enable cookies and click English

nitrohorse:
The larger the list grows, I’m wondering if it makes sense to create a new table like the Email and VPN providers have?

Mikaela:
Assigning to Nitrohorse who agreed to take this/tablesizing on Wire

Mikaela:
> However it's worth noting that these listings are based on what the sites are saying, I don't know if there are methods to ensure those claims are true.

Me at Reddit, maybe this should also be said on the site?

nitrohorse:
^^ hmm, yeah, I think that's a good callout.

nitrohorse:
How's this for the columns / DNS resolver criteria for a new table?
- DNS Provider
- Website
- Server Locations
- Logging
- Protocols
- Filtering
[...]

Issue #1070, opened 2019-08-01
🆕 Software Suggestion | DNSCloak


nitrohorse:
## Basic Information

**Name:** DNSCloak
**Category:** [DNS](https://www.privacytools.io/providers/dns/)
**URLs:**
- [Source Code](https://github.com/s-s/dnscloak) (licensed under [MPL 2.0](https://github.com/s-s/dnscloak/blob/master/LICENSE))
- From the [Center for the Cultivation of Technology gemeinnuetzige GmbH](https://techcultivation.org/), "a charitable non-profit host organization for international Free Software projects."
- [Privacy Policy](https://drive.google.com/file/d/1050No_pU74CAWUS5-BwQWyO2x_aiMzWc/view) (linked from the App Store listing)
- [App Store link](https://apps.apple.com/app/id1452162351)
- [Beta TestFlight link](https://testflight.apple.com/join/RlMeZBo7)
- [Some forum discussion](https://forum.privacytools.io/t/dnscloak-for-ios-is-open-source/827/2)

## Description

DNSCloak is an "iOS GUI and wrapper for dnscrypt-proxy 2." It "implements DNSCrypt & DNS-over-HTTPS/2 (DoH) protocols and allows you to secure your DNS traffic against of spoofing. It overrides your DNS settings with a single or multiple DNSCrypt or DNS-over-HTTPS/2 (DoH) resolvers."

## Related Issues:
- https://github.com/privacytoolsIO/privacytools.io/pull/1054
- https://github.com/privacytoolsIO/privacytools.io/issues/1068
- https://github.com/privacytoolsIO/privacytools.io/issues/1069

Issue #1068, opened 2019-07-31
🆕 Software Suggestion | Android (9+) Private DNS


Mikaela:
## Basic Information

**Name:** Private DNS
**Category:** Android Privacy Add-ons
**URL:** https://support.google.com/android/answer/9089903?hl=en

## Description

Android supports DNS over TLS natively and it can be configured in *Settings, Network & internet and then Advanced and then Private DNS*. Shouldn't we recommend users to enable it?

Blocker/link-to: #1054 ?

Edit/addition: Warning? If your DoT provider (or port 853) is blocked, Android is going to report that the network has no connectivity.

Mikaela:
Oh, I forgot that MiUi hides the feature due to China(?) and it might be nice to list some workaround, preferably F-Droid one

* https://www.xda-developers.com/enable-private-dns-xiaomi-devices-miui-10/

nitrohorse:
Good idea! I think we should recommend this; maybe below the [ICANN DNS section](https://www.privacytools.io/providers/dns/#icanndns)?

Mikaela:
> I think we should recommend this; maybe below the ICANN DNS section?

I am not entirely sure, because that is on a page for DNS providers, while we have separate Android section.

How about putting it to Android section, but having "see also Private DNS on our Android page" under ICANN DNS section?

Mikaela:
> How about putting it to Android section, but having "see also Private DNS on our Android page" under ICANN DNS section?

Or the other way probably works too?

Logo ideas:

![private-dns-dark](https://user-images.githubusercontent.com/831184/62280875-3fa00580-b43c-11e9-938b-f7ace7c7481b.png)
![private-dns-menu](https://user-images.githubusercontent.com/831184/62280877-40389c00-b43c-11e9-8f69-d80cf7ffbd72.png)

which don't look that good outside of my head, and all our DNS providers are warninged somehow.  I didn't think of the smaller one until I saw it and I would need to change my language to take the screenshot again, but I am in a bit hurry right now and there is no hurry until something is actually done.

beerisgood:
Remember that this doesn't work (yet) with Blokada. But they still work on it

Mikaela:
Oh, yes, the problem is that when a VPS is enabled, DoT is not used as the traffic goes through the VPN instead and the VPN decides what DNS servers it wants to use. I guess this should have a warning label while noting that DoT can possibly protect the DNS query to the VPN server assuming it's not using the IP address directly?

[...]

Issue #1055, opened 2019-07-25
🆕 Software Suggestion | DNS server software for advanced users


Mikaela:
This is salvaged from #1028 which solution #1054 I cut.

* [ ] Unbound (software)
* [ ] Bind (software)
* [ ] PowerDNS (software)
* [ ] the DNSCrypt server (I need to read more details) (software)

TODO: fill the fields

## Basic Information

**Name:**
**Category:**
**URL:**

## Basic Information

**Name:**
**Category:**
**URL:**

## Basic Information

**Name:**
**Category:**
**URL:**

## Basic Information

**Name:**
**Category:**
**URL:**

nitrohorse:
First go at it:

**Name:** DNSSEC-Trigger
**Category:** Software
**Description:** "DNSSEC-Trigger is experimental software that enables your computer to use DNSSEC protection for the DNS traffic."
**URLs:**
  - [Project](https://www.nlnetlabs.nl/projects/dnssec-trigger/about/)
  - [GitHub](https://github.com/NLnetLabs/dnssec-trigger)

**Name:** Unbound
**Category:** Software
**Description:** "Unbound is a validating, recursive, caching DNS resolver. It is designed to be fast and lean and incorporates modern features based on open standards."
**URLs:**
  - [Project](https://nlnetlabs.nl/projects/unbound/about/)
  - [GitHub](https://github.com/NLnetLabs/unbound)

**Name:** BIND 9
**Category:** Software
**Description:** "BIND 9 is the most widely used DNS software on the Internet. It performs both of the main DNS server roles – acting as an authoritative name server for one or more specific domains, and acting as a recursive resolver for the DNS system generally."
**URLs:**
  - [Project](https://www.isc.org/bind/)
  - [GitLab](https://gitlab.isc.org/isc-projects/bind9)

**Name:** PowerDNS
**Category:** Software
**Description**: "PowerDNS is a DNS server, written in C++ and licensed under the GPL."
**URLs:**
  - [Project](https://www.powerdns.com/)
  - [GitHub](https://github.com/PowerDNS/pdns)

**Name:** DNSCrypt
**Category:** Protocol
**Description:** "DNSCrypt is a network protocol which authenticates and encrypts Domain Name System (DNS) traffic between the user's computer and recursive name servers."
**URLs:**
  - [Project](https://dnscrypt.info/)
  - [GitHub](https://github.com/dnscrypt)

Mikaela:
[...]

PR #1097, opened 2019-08-05
Add Encrypted DNS providers table

nitrohorse:
## Description

Resolves: #1077 
Resolves: #1068
Resolves: #1070


nitrohorse:
@Mikaela I think we'll want to clarify criteria for adding a DNS provider to the list and also whether any that I've listed should be removed.

_includes/sections/dns.html
```diff
@@ -33,10 +33,223 @@ <h3>Worth Mentioning</h3>
...
-    <li><a href="https://powerdns.org/">PowerDNS</a> - A best effort DoH service. Servers located in the Netherlands.</li>
-    <li><a href="https://quad9.net/">Quad9 DNS</a> - A non-profit, anycast DNS provider founded by <a href="https://www-03.ibm.com/press/us/en/pressrelease/53388.wss">IBM</a>, <a href="https://www.pch.net/">PCH</a>, and <a href=https://www.globalcyberalliance.org/quad9/"">Global Cyber Alliance</a>. Provides malicious domain filtering and supports DoH, DoT, and DNSCrypt. <span class="badge badge-warning" data-toggle="tooltip" title="Founders of Global Cyber Alliance include: City of London Police & Manhattan District Attorney's Office">Warnings <i class="far fa-question-circle"></i></a></span></li>
 </ul>
+
+<h1 id="icanndns" class="anchor"><a href="#icanndns"><i class="fas fa-link anchor-icon"></i></a> Encrypted ICANN DNS Providers</h1>
+
+<div class="alert alert-warning" role="alert">
+  <strong>Note: Using an encrypted DNS provider will not make you anonymous. But it will give you a better privacy. Don't rely on a "no log" policy.</strong>
```
Mikaela: > But it will give you a better privacy.

In the future hopefully, but currently the sites you access get leaked by IP address and SNI. I think a better wording would be:

> But it will prevent DNS hijacking

another benefit is that it could make tracking your activity by a LAN watcher a bit more difficult, but I don't know what is the state of the tools that monitor SNI and IP addresses connected to.

_includes/sections/dns.html
```diff
@@ -33,10 +33,223 @@ <h3>Worth Mentioning</h3>
...
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title='"CZ.NIC resolvers neither collect any personal data nor gather information on pages where your computer sends personal data."'>
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+        <td data-value="0"><a data-toggle="tooltip" data-placement="bottom" data-original-title='"CZ.NIC is an interest association of legal entities, founded in 1998 by leading providers of Internet services."' href="https://www.nic.cz/page/351/about-association/">Association</a></td>
+        <td>N</td>
+        <td>IPv4, IPv6, DoH, DoT</td>
+        <td>Y</td>
+        <td>Y</td>
+        <td>N/A</td>
```
Mikaela: I would prefer saying `?` or `unknown` or `U/K` instead of `N/A` which I think gives the wrong impression.

_includes/sections/dns.html
Mikaela: Another note, as you seem to list IPv4 and IPv6, it should be noted that those generally aren't encrypted if people happen to add them.

_includes/sections/dns.html
```diff
@@ -33,10 +33,223 @@ <h3>Worth Mentioning</h3>
...
+        <td>Anycast (based in <span class="flag-icon flag-icon-us"></span> USA)</td>
+        <td>
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title="https://quad9.net/policy/" href="https://quad9.net/policy/">
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+        <td>Non-Profit</td>
+        <td><a href="https://quad9.net/policy/">Y</a></td>
+        <td>IPv4, IPv6, DoH, DoT, DNSCrypt</td>
+        <td>Y</td>
```
Mikaela: Note that their "insecure" DNS doesn't have DNSSEC either.

_includes/sections/dns.html
```diff
@@ -33,10 +33,223 @@ <h3>Worth Mentioning</h3>
...
+        <td>Y</td>
+        <td>Based on server choice</td>
```
Mikaela: (see the DNSSEC comment)

_includes/sections/dns.html
```diff
@@ -33,10 +33,223 @@ <h3>Worth Mentioning</h3>
...
+          <a href="https://quad9.net/">Quad9</a> <span class="badge badge-warning" data-toggle="tooltip" title="Founders include the Global Cyber Alliance, comprised of the City of London Police and Manhattan District Attorney's Office"><i class="fas fa-exclamation-triangle"></i></a></span>
+        </td>
+        <td>Anycast (based in <span class="flag-icon flag-icon-us"></span> USA)</td>
+        <td>
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title="https://quad9.net/policy/" href="https://quad9.net/policy/">
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+        <td>Non-Profit</td>
+        <td><a href="https://quad9.net/policy/">Y</a></td>
```
Mikaela: Isn't this also some as with applied privacy or where do you set the border between Y and Some?

I am looking at https://appliedprivacy.net/privacy-policy/ and https://quad9.net/policy/

_includes/sections/dns.html
```diff
@@ -33,10 +33,223 @@ <h3>Worth Mentioning</h3>
...
+        <td>IPv4, IPv6, DoT</td>
+        <td>Y</td>
+        <td>N</td>
+        <td>N</td>
+        <td>N/A</td>
+      </tr>
+    </tbody>
+  </table>
+
+  <h3>Worth Mentioning</h3>
```
Mikaela: Could we defiate from the usual "Worth Mentioning" and instead be "Worth Mentioning and additional information"?

_includes/sections/dns.html
```diff
@@ -33,10 +33,223 @@ <h3>Worth Mentioning</h3>
...
+
+  <ul>
+    <li><a href="https://www.isc.org/blogs/qname-minimization-and-privacy/">QNAME Minimization and Your Privacy</a> by the Internet Systems Consortium (ISC)</li>
+    <li><a href="https://www.isc.org/dnssec/">DNSSEC and BIND 9</a> by the ISC</li>
+    <li>Android 9 comes with a DoT client by <a href="https://android-developers.googleblog.com/2018/04/dns-over-tls-support-in-android-p.html">default</a>. <span class="badge badge-warning" data-toggle="tooltip" data-original-title="...but with some caveats"><a href="https://www.quad9.net/private-dns-quad9-android9/"><i class="fas fa-exclamation-triangle"></i></a></li>
```
Mikaela: I think the "default" link should be https://support.google.com/android/answer/9089903 (`?hl=en` to force language if you wish) instead as that blog doesn't even tell where to enable it.

_includes/sections/dns.html
Mikaela: Oh and there isn't the note that we cannot verify that providers actually follow their privacy policies?

Mikaela:
> I think we'll want to clarify criteria for adding a DNS provider to the list and also whether any that I've listed should be removed.

What I am thinking about:

* supports encrypted DNS (DoT/DoH/DNSCrypt), especially DoH as required by #785 
* preferably doesn't keep logs, but I would say that "some" are fine, but I guess that needs definition as currently we seem to disagree on it (AppliedPrivacy/Quad9)?
* preferably supports DNSSEC, but in optimal case the user would be validating that locally e.g. with unbound or dnssec-trigger. Whether that works is again a different question as e.g. OpenDNS (at least used to) strips queries by Unbound for DNSSEC validation breaking everything.

The server locations are again a question mark, should they be outside of 14 eyes? But I would argue that location matters with DNS servers as there is the question of speed (which may matter more to average user) and possibly CDN optimization. Then devices (especially phones/tablets) also travel which may be a bonus for anycast, but if I recall correctly OpenNIC is worried about that being centralization.

I think Quad9 currently has the widest anycast network and I wonder if delisting it would send traffic to `dns.google` or `cloudflare-dns.com`, but I respect your decision if you think it's better to delist it.

Which actually leads to question, do you think Cloudflare DNS should be listed or is it established as too big by #374 and some comments in #785 ? I am aware that many people in /r/privacytoolsio and possibly elsewhere in the community are using it. And that leads to a question should there be a warning about Cloudflare in addition to Google?

_includes/sections/dns.html
```diff
@@ -33,10 +33,223 @@ <h3>Worth Mentioning</h3>
...
+      <tr>
+        <th data-sorted="true" data-sorted-direction="descending">ICANN DNS Provider</th>
+        <th data-sortable="true">Server Locations</th>
+        <th data-sortable="false">Privacy Policy</th>
+        <th data-sortable="true">Type</th>
+        <th data-sortable="true">Logging</th>
+        <th data-sortable="true">Protocols</th>
+        <th data-sortable="true">DNSSEC</th>
+        <th data-sortable="true">QNAME Minimization</th>
+        <th data-sortable="true">Filtering</th>
```
Mikaela: I just remember that this should possibly be more explicitly explained in the table, what is being filtered?

_includes/sections/dns.html
```diff
@@ -33,10 +33,223 @@ <h3>Worth Mentioning</h3>
...
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title="https://adguard.com/en/privacy/dns.html" href="https://adguard.com/en/privacy/dns.html">
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+        <td>Commercial</td>
+        <td>N</td>
+        <td>IPv4, IPv6, DoH, DoT, DNSCrypt</td>
+        <td>N</td>
+        <td>Y</td>
+        <td>Y</td>
```
Mikaela: Ads, trackers, malicious domains?

_includes/sections/dns.html
```diff
@@ -33,10 +33,223 @@ <h3>Worth Mentioning</h3>
...
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title='"No logs."'>
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+        <td>Hobby Project</td>
+        <td>N</td>
+        <td>DoH, DoT, DNScrypt</td>
+        <td>Y</td>
+        <td>Y</td>
+        <td>Y</td>
```
Mikaela: " Filtered some ads, trackers, malware", some wildcards domains and international non-ASCII domains which can be a problem.

* https://github.com/ookangzheng/blahdns#default-blocked-wildcard-domain

I don't know how common they are, but I can think of two, I don't know if one of them still exists though, but they are likely more common in different countries.

_includes/sections/dns.html
Mikaela: Malicious domain filtering.

_includes/sections/dns.html
```diff
@@ -33,10 +33,223 @@ <h3>Worth Mentioning</h3>
...
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title="https://securedns.eu/#privacy" href="https://securedns.eu/#privacy">
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+        <td>Hobby Project</td>
+        <td>N</td>
+        <td>DoH, DoT, DNScrypt</td>
+        <td>Y</td>
+        <td>Y</td>
+        <td>Y</td>
```
Mikaela: Ads depending on server choice? https://github.com/notracking/hosts-blocklists

_includes/sections/dns.html
nitrohorse: > But it will prevent DNS hijacking

Good point, I stole that alert from the VPN page :sweat_smile: 

> Another note, as you seem to list IPv4 and IPv6, it should be noted that those generally aren't encrypted if people happen to add them.

Good catch; I think since this table is specifically for encrypted DNS I'll remove those from being included in the protocols.

> I guess I was thinking of something more explicit.

I'm wondering if adding something like "We also cannot verify that the below providers actually follow their own privacy policies so use caution." is really needed? Wouldn't this apply then to most everything then? I'm not sure...

nitrohorse: What about this:

> Note: Using an encrypted DNS provider will not make you anonymous, nor hide your internet traffic from your Internet Service Provider. But it will prevent DNS hijacking and spoofing, and make your DNS queries harder to share with third parties.

_includes/sections/dns.html
nitrohorse: Gotcha; ok, removed since it doesn't really make sense to include it in this table.

_includes/sections/dns.html
nitrohorse: Good callout; I'll go with `?` unless someone feels stronger about another method :smile:

_includes/sections/dns.html
nitrohorse: Good catch! Okay, will update to "Ads, trackers, malicious domains" rather than a boolean.

_includes/sections/dns.html
nitrohorse: Good catch! We can add a warning tooltip for this.

_includes/sections/dns.html
nitrohorse: Good catch, you're right!

_includes/sections/dns.html
nitrohorse: Nice, thanks!

_includes/sections/dns.html
nitrohorse: You're right; will update to "some" for consistency.

_includes/sections/dns.html
nitrohorse: Updated!

nitrohorse:
> > I think we'll want to clarify criteria for adding a DNS provider to the list and also whether any that I've listed should be removed.
> 
> What I am thinking about:
> 
>     * supports encrypted DNS (DoT/DoH/DNSCrypt), especially DoH as required by #785

That makes sense; it would also disqualify [UncensoredDNS](https://blog.uncensoreddns.org/) since they only support DoT.

>     * preferably doesn't keep logs, but I would say that "some" are fine, but I guess that needs definition as currently we seem to disagree on it (AppliedPrivacy/Quad9)?

Yeah, I agree that "some" is fine in certain cases like Applied Privacy and Quad9. Based on their polices, Applied Privacy and Q9 both log and aggregate metrics based primarily on counts and not IP address (which they treat as PII).

>     * preferably supports DNSSEC, but in optimal case the user would be validating that locally e.g. with unbound or dnssec-trigger. Whether that works is again a different question as e.g. OpenDNS (at least used to) strips queries by Unbound for DNSSEC validation breaking everything.

Good callout. AdGuard relies on Cloudflare, Google, and OpenDNS as upstream DNS providers, I think, while both Cloudflare and Google provide DNSSEC, OpenDNS currently does not. So I guess in AdGuard's case it's "partial" DNSSEC support since theoretically a third of queries wouldn't be validated? This was also mentioned in a [Reddit thread](https://www.reddit.com/r/Adguard/comments/bbb9md/adguard_dns_doesnt_validate_dnssec_signatures/) which AdGuard responded as:
> This is, in fact, a valid concern. We will consider removing OpenDNS from the list of our upstream servers. Thank you for the suggestion.

Interestingly, a product manager from OpenDNS [also replied on that thread](https://www.reddit.com/r/Adguard/comments/bbb9md/adguard_dns_doesnt_validate_dnssec_signatures/em3hozo/):

> I can confirm that we are working on partial DNSSEC support...rather than give up the filtering that makes our DNS resolvers unique, we've chosen to limit DNSSEC validation to upstream queries in order to provide data authenticity for data received from authorities, and then continue to support DNSCrypt as the means to provide data authenticity and authentication for clients connecting to us. 

> The server locations are again a question mark, should they be outside of 14 eyes? But I would argue that location matters with DNS servers as there is the question of speed (which may matter more to average user) and possibly CDN optimization. Then devices (especially phones/tablets) also travel which may be a bonus for anycast, but if I recall correctly OpenNIC is worried about that being centralization.
> I think Quad9 currently has the widest anycast network and I wonder if delisting it would send traffic to `dns.google` or `cloudflare-dns.com`, but I respect your decision if you think it's better to delist it.

 I wonder if rather than 'outside of 14 eyes' we'd go with what the VPN providers table uses, "outside of the US"? That would solely disqualify Quad9 though.

> Which actually leads to question, do you think Cloudflare DNS should be listed or is it established as too big by #374 and some comments in #785 ? I am aware that many people in /r/privacytoolsio and possibly elsewhere in the community are using it. And that leads to a question should there be a warning about Cloudflare in addition to Google?

From my perspective, I think if our criteria would allow "inside" the US (essentially no location criteria), we would be inclined to include Quad9 _and_ Cloudflare (especially with [Mozilla partnering with CF](https://blog.nightly.mozilla.org/2018/06/01/improving-dns-privacy-in-firefox/)). We need to decide where the balance and compromise is.

_includes/sections/dns.html
Mikaela: Sounds good, but I would add third party or something in front of spoofing as that is what the filtering DNS providers do and you have to trust them unless the zone is DNSSEC signed and you run a local DNSSEC validator (which I don't think is done by anyone on Android/iOS while it's somewhat rare also on desktops).

Mikaela:
>> supports encrypted DNS (DoT/DoH/DNSCrypt), especially DoH as required by #785

> That makes sense; it would also disqualify UncensoredDNS since they only support DoT.

Sorry, I was thinking / more of as "or" here and I would be fine with UncensoredDNS as the readers are going to see clearly that it's DoT-only and not consider it for their DoH requirement, but might take it for Android, thanks to your table :)

>  AdGuard relies on Cloudflare, Google, and OpenDNS as upstream DNS providers, I think

What is the source for this? I find it worrying as I would expect DNS providers to perform name resolution by themselves querying the root servers when they don't know which are the authoritative nameservers. Was the word recurse?

> If we still want to include AdGuard, I've added a tooltip to that table cell and updated the value to "Partial."

I guess this is the best solution for now

> I wonder if rather than 'outside of 14 eyes' we'd go with what the VPN providers table uses, "outside of the US"? That would disqualify Quad9 though and others like Cloudflare.

I don't think it's a good solution, because there are a lot of people in the US who also need DNS (who doesn't need DNS?) and using foreign resolvers could lead to slower speeds and content delivery happening from foreign instead of local datacenters and causing slower speeds.

> From my perspective, I think if our criteria would allow "inside" the US (essentially no location criteria), we would be inclined to include Quad9 and Cloudflare (especially with Mozilla partnering with CF) (and maybe even nextdns).

I guess we have to include them in that case. Can we at least have a warning "a big part of the internet is in Cloudflare's networks and it's a problem considering decentralization" or similar?

> I think regardless of including CF or not, we could add a snippet about Google in the top warning banner?

:+1: I am not entirely sure on "your ISP" part as I think they are more reliable in some countries, but then again they don't currently support DoT/DoH, but I hope in the future they will and if they have a good privacy policy, I guess we can include them or link elsewhere for comparsion of them assuming they are going to only serve their own customers?

_includes/sections/dns.html
```diff
@@ -33,10 +33,223 @@ <h3>Worth Mentioning</h3>
...
+        </td>
+        <td><span class="flag-icon flag-icon-ch"></span> Switzerland, <span class="flag-icon flag-icon-jp"></span> Japan, <span class="flag-icon flag-icon-de"></span> Germany</td>
+        <td>
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title='"No logs."'>
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+        <td>Hobby Project</td>
+        <td>N</td>
+        <td>DoH, DoT, DNScrypt</td>
```
Mikaela: Inconsistency, some parts say DNSCrypt some DNScrypt (note the `c` / `C`), DNSCrypt appears to be the upstream spelling judging by https://dnscrypt.info/

_includes/sections/dns.html
```diff
@@ -33,10 +33,223 @@ <h3>Worth Mentioning</h3>
...
+        </td>
+        <td><span class="flag-icon flag-icon-nl"></span> The Netherlands</td>
+        <td>
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title="https://securedns.eu/#privacy" href="https://securedns.eu/#privacy">
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+        <td>Hobby Project</td>
+        <td>N</td>
+        <td>DoH, DoT, DNScrypt</td>
```
Mikaela: Inconsistency, some parts say DNSCrypt some DNScrypt (note the `c` / `C`), DNSCrypt appears to be the upstream spelling judging by https://dnscrypt.info/

_includes/sections/dns.html
```diff
@@ -33,10 +33,223 @@ <h3>Worth Mentioning</h3>
...
+<div class="table-responsive">
+  <table class="table sortable-theme-bootstrap" data-sortable>
+    <thead>
+      <tr>
+        <th data-sorted="true" data-sorted-direction="descending">ICANN DNS Provider</th>
+        <th data-sortable="true">Server Locations</th>
+        <th data-sortable="false">Privacy Policy</th>
+        <th data-sortable="true">Type</th>
+        <th data-sortable="true">Logging</th>
+        <th data-sortable="true">Protocols</th>
```
Mikaela: Now I also realized that we don't expand what do DoH/DoT/DNSCrypt mean? Maybe they should have a short paragraph in the additional information in the bottom.

> DNS over TLS (DoT) is a \<explain https://en.wikipedia.org/wiki/Request_for_Comments_(identifier) here or just say RFC?\> for encrypted DNS on a dedicated port 853, DNS over HTTPS (DoH) is similar, but uses HTTPS instead being indistinguishable from "normal" HTTPS traffic on port 443 and DNSCrypt is an older regardless robust method of encrypting DNS.

Would we also need to give examples on what supports what? I guess the page already mentions DNSCrypt-proxy (possibly also as supporting DoH which it does) and Android 9 supports TLS, but would Firefox need to be mentioned or is it enough for #785 to refer to the table? I think it may be a non-issue.

<small>Sorry, my head isn't still working that well and I am not sure if this works as a base for you to improve.</small>

_includes/sections/dns.html
```diff
@@ -33,10 +33,223 @@ <h3>Worth Mentioning</h3>
...
+        <td>Anycast (based in <span class="flag-icon flag-icon-cy"></span> Cyprus)</td>
+        <td>
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title="https://adguard.com/en/privacy/dns.html" href="https://adguard.com/en/privacy/dns.html">
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+        <td>Commercial</td>
+        <td>N</td>
+        <td>DoH, DoT, DNSCrypt</td>
+        <td><a data-toggle="tooltip" data-placement="bottom" data-original-title="Of AdGuard's three upstream providers (Cloudflare, Google, and OpenDNS), OpenDNS does not support DNSSEC." href="https://old.reddit.com/r/Adguard/comments/bbb9md/adguard_dns_doesnt_validate_dnssec_signatures/">Partial</a></td>
```
Mikaela: Sorry, another small nitpick, but is there a particular reason for old.reddit.com privacy-wise or are you one of those people who are used to it and want to resist the change? I am just thinking on how the new interface is better on small screens (phones)?

Mikaela:
Feedback from friends:

* * * * *

> I find the term "DNS provider" slightly confusing as it's rather overloaded. Not sure if I can come up with a better one that doesn't sound like technical mumbo-jumbo.
>
> I really think there ought to be a paragraph about how DNS resolvers are different service from authoritative DNS hosts. And perhaps something about setting up your own as that's really easy, especially if you have *wrt device for a router already.

We have issues for the later already, #1055 and privacytoolsio/guides.privacytools.io#9

* * * * *

*The sorting is weird, if you sort by Protocol, it's alphabetic and DoH, DoT should rank higher than just DoH. When sorted by filtering, ? should rank the same as N.*

I don't know if that is easy to fix?

_includes/sections/dns.html
nitrohorse: https://i.reddit.com has the least amount of trackers with ‘old’ next and then the new UI having the most. “i” looks horrible on desktops so maybe we just stick with the new UI since it’s more user-friendly to all devices?

nitrohorse:
> > > supports encrypted DNS (DoT/DoH/DNSCrypt), especially DoH as required by #785
> 
> > That makes sense; it would also disqualify UncensoredDNS since they only support DoT.
> 
> Sorry, I was thinking / more of as "or" here and I would be fine with UncensoredDNS as the readers are going to see clearly that it's DoT-only and not consider it for their DoH requirement, but might take it for Android, thanks to your table :)

Ah, yeah, that makes sense :+1: 

> > AdGuard relies on Cloudflare, Google, and OpenDNS as upstream DNS providers, I think
> 
> What is the source for this? I find it worrying as I would expect DNS providers to perform name resolution by themselves querying the root servers when they don't know which are the authoritative nameservers. Was the word recurse?
> > If we still want to include AdGuard, I've added a tooltip to that table cell and updated the value to "Partial."
> 
> I guess this is the best solution for now

Huh, this is interesting. From some testing a month or two ago DNS results would return resolvers for Cloudflare/Google/OpenDNS which AdGuard after filtering, forwards requests to. But checking now, it looks like that isn't happening anymore; I'm connecting to a single Anycast resolver hosted on Vultr with DNSSEC support. I also can't find any official documentation in their [FAQ](https://kb.adguard.com/en/dns) regarding their upstream providers, so I'm inclined to change their DNSSEC value to `Y` and remove the warning.

> > I wonder if rather than 'outside of 14 eyes' we'd go with what the VPN providers table uses, "outside of the US"? That would disqualify Quad9 though and others like Cloudflare.
> 
> I don't think it's a good solution, because there are a lot of people in the US who also need DNS (who doesn't need DNS?) and using foreign resolvers could lead to slower speeds and content delivery happening from foreign instead of local datacenters and causing slower speeds.

Yeah, I agree with you; I think there's value in removing the location criteria for DNS.

> > From my perspective, I think if our criteria would allow "inside" the US (essentially no location criteria), we would be inclined to include Quad9 and Cloudflare (especially with Mozilla partnering with CF) (and maybe even nextdns).
> 
> I guess we have to include them in that case. Can we at least have a warning "a big part of the internet is in Cloudflare's networks and it's a problem considering decentralization" or similar?

Yeah, I think that'd be good. I can add a new row for Cloudflare with a warning. What do you think of [nextdns](https://www.nextdns.io/)? It's fairly new, but arguably on par with a "hobby project" now I guess which wouldn't disqualify it in that regard.

> > I think regardless of including CF or not, we could add a snippet about Google in the top warning banner?
> 
> +1 I am not entirely sure on "your ISP" part as I think they are more reliable in some countries, but then again they don't currently support DoT/DoH, but I hope in the future they will and if they have a good privacy policy, I guess we can include them or link elsewhere for comparsion of them assuming they are going to only serve their own customers?

Yeah, good callout, I can remove the ISP part unless someone things otherwise.

_includes/sections/dns.html
nitrohorse: I think since there doesn't appear to be partial support but rather full support for DNSSEC now we can just remove this.

_includes/sections/dns.html
nitrohorse: Great callout -- what do you think about something like this for the terms (viewable also from the preview deployment: https://deploy-preview-1097--privacytools-io.netlify.com/)

terms

And something like this for mentioning clients?

worth

_includes/sections/dns.html
```diff
@@ -33,10 +33,278 @@ <h3>Worth Mentioning</h3>
...
+        <td>Ads, trackers, malicious domains <span class="badge badge-warning" data-toggle="tooltip" data-original-title="And some wildcard, IDN, and non-ASCII domains."><a href="https://github.com/ookangzheng/blahdns#default-blocked-wildcard-domain"><i class="fas fa-exclamation-triangle"></i></a></span></td>
+        <td>
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title="https://github.com/ookangzheng/blahdns/" href="https://github.com/ookangzheng/blahdns/">
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+      </tr>
+
+      <tr>
+        <td data-value="Cloudflare">
```
nitrohorse: @Mikaela you'll notice I've added CF + nextdns for discussion here.

nitrohorse:
> Feedback from friends:
> 
> > I find the term "DNS provider" slightly confusing as it's rather overloaded. Not sure if I can come up with a better one that doesn't sound like technical mumbo-jumbo.
> > I really think there ought to be a paragraph about how DNS resolvers are different service from authoritative DNS hosts. And perhaps something about setting up your own as that's really easy, especially if you have *wrt device for a router already.
> 
> We have issues for the later already, #1055 and [privacytoolsIO/guides.privacytools.io#9](https://github.com/privacytoolsIO/guides.privacytools.io/issues/9)

Hmm, that's a good point... let's save that for later; we should address that though.
> 
> _The sorting is weird, if you sort by Protocol, it's alphabetic and DoH, DoT should rank higher than just DoH. When sorted by filtering, ? should rank the same as N._
> 

Okay, I resolved the `?/N` oddity so they're treated the same. I can't figure out right now how to custom sort by length and then alphanumerically but I have a hack in place where I give only "DoT" values, _"DoH"_ as the sort value... :sweat:

_includes/sections/dns.html
Mikaela: :+1: looks good

_includes/sections/dns.html
```diff
@@ -33,10 +33,278 @@ <h3>Worth Mentioning</h3>
...
+          <a href="https://1.1.1.1/dns/">Cloudflare</a>
```
```suggestion
          <a href="https://cloudflare-dns.com/dns/">Cloudflare</a>
```

1.1.1.1 is IPv4-only and I don't think we should hardcode it as there are already some IPv6-only networks in the world even if there are transition mechanisms.

Mikaela: (now I also finally learned to make suggestions)

Mikaela: But where do they actually document the DoT and DoH addresses?

_includes/sections/dns.html
```diff
@@ -33,10 +33,278 @@ <h3>Worth Mentioning</h3>
...
+  <ul>
+    <li>DNS over TLS (DoT) - a security protocol for encrypted DNS on a dedicated port 853.</li>
+    <li>DNS over HTTPS (DoH) - similar to DoT, but uses HTTPS instead, being indistinguishable from "normal" HTTPS traffic on port 443.</li>
+    <li>DNSCrypt - an older yet robust method of encrypting DNS.</li>
+  </ul>
+
+  <h3>Worth Mentioning and additional information</h3>
+
+  <ul>
+    <li>Firefox comes with built-in DoH support with Cloudflare set as the default resolver, but can be configured to use any DoH resolver. <span class="badge badge-warning" data-toggle="tooltip" title='"Cloudflare has agreed to collect only a limited amount of data about the DNS requests that are sent to the Cloudflare Resolver for Firefox via the Firefox browser."'><a href="https://developers.cloudflare.com/1.1.1.1/commitment-to-privacy/privacy-policy/firefox/"><i class="fas fa-exclamation-triangle"></i></a></span></li>
```
Mikaela: Is it also enabled by default already?

_includes/sections/dns.html
```diff
@@ -33,10 +33,278 @@ <h3>Worth Mentioning</h3>
...
+        <td><a data-toggle="tooltip" data-placement="bottom" data-original-title='"We do NOT log your IP address or DNS queries during normal operations. We do NOT share query data with third parties that are not directly involved with resolving the query (i.e. sending queries to authoritative nameservers for resolution)."' href="https://appliedprivacy.net/privacy-policy/">Some</a></td>
+        <td>DoH, DoT, DNS-over-Onion (experimental)</td>
+        <td>Y</td>
+        <td>Y</td>
+        <td>N</td>
+        <td>?</td>
+      </tr>
+
+      <tr>
+        <td data-value="nextdns">
```
Mikaela: I have a small problem with them, but apparently my problem doesn't even apply with encrypted DNS, so I have no problem :)

* https://forum.privacytools.io/t/secure-dns-alternative-nextdns-io/691/4?u=mikaela

_includes/sections/dns.html
nitrohorse: Todo: ads warning

nitrohorse: Updated and linked to https://codeberg.org/crimeflare/cloudflare-tor/ which looks more up-to-date compared to https://notabug.org/themusicgod1/cloudflare-tor/src/master.

_includes/sections/dns.html
nitrohorse: Hmm... I'm not 100% sure but I think not yet -- they're still conducting studies: https://blog.mozilla.org/futurereleases/2019/07/31/dns-over-https-doh-update-detecting-managed-networks-and-user-choice/

_includes/sections/dns.html
nitrohorse: Hmm good point! Need to find that...

nitrohorse: I guess both are documented here? https://developers.cloudflare.com/1.1.1.1/dns-over-https/

_includes/sections/dns.html
nitrohorse: You think that should be added here, that it's not enabled by default right now?

_includes/sections/dns.html
```diff
@@ -33,10 +33,278 @@ <h3>Worth Mentioning</h3>
...
+        <td>
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title='"Absolutely nothing is being logged, neither about the users nor the usage of this service. I do keep graphs of the total number of queries, but no personally identifiable information is saved. The data that is saved will never be sold or used for anything except capacity planning of the service."'>
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+        <td>Hobby Project</td>
+        <td>N</td>
+        <td data-value="doh">DoT</td>
+        <td>Y</td>
+        <td>N</td>
```
nitrohorse: @Mikaela would you be able to double-check that they don’t support QNAME Min?

Mikaela: Not as easily as the previous one as it's not listed in DNSCrypt-proxy, but I can reconfigure my Unbound, I guess.

Mikaela: I receive

```
└┌(%:~)┌- dig +short txt qnamemintest.internet.nl @127.0.0.1
a.b.qnamemin-test.internet.nl.
"NO - QNAME minimisation is NOT enabled on your resolver :("
```

when I configure Unbound with

```
# NOTE! Requires Unbound 1.7.3 or newer! Debian 9 has 1.6.0
# cp of forwards.conf updated to DNS over TLS time with  a lot took from
# https://www.ctrl.blog/entry/unbound-tls-forwarding.html

server:
  # Debian ca-certificates location
  tls-cert-bundle: /etc/ssl/certs/ca-certificates.crt
  # ctrl.blog says this is the Fedora location
  #tls-cert-bundle: /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem

# Forward queries to
forward-zone:
    name: "."
    forward-tls-upstream: yes
    forward-addr: 91.239.100.100@853#anycast.censurfridns.dk
    forward-addr: 2001:67c:28a4::@853#anycast.censurfridns.dk
```

and have commented qname-minimization.conf

```
server:
    # Send minimum amount of information to upstream servers to enhance
    # privacy. Only sends minimum required labels of the QNAME and sets
    # QTYPE to NS when possible.

    # See RFC 7816 "DNS Query Name Minimisation to Improve Privacy" for
    # details.

    #qname-minimisation: yes
```

Mikaela: wait, that is entirely different DNS provider I just tested? Let's try this again

_includes/sections/dns.html
Mikaela: I forgot that the studies exist and I am not sure, maybe mention that Mozilla is testing it with small subset of users?

_includes/sections/dns.html
Mikaela: No, it's the same DNS provider, but they just confusingly use different domain for their anycast DoT than website.

Mikaela:
I am not sure if the source_code file needs updating in this case as you already linked sources everywhere.

https://deploy-preview-1097--privacytools-io.netlify.com/classic/#icanndns also seems to work directly as it wasn't a new page that was added.

_includes/sections/dns.html
```diff
@@ -33,10 +33,278 @@ <h3>Worth Mentioning</h3>
...
+        <td>
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title="https://github.com/ookangzheng/blahdns/" href="https://github.com/ookangzheng/blahdns/">
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+      </tr>
+
+      <tr>
+        <td data-value="Cloudflare">
+          <a href="https://1.1.1.1/dns/">Cloudflare</a> <span class="badge badge-warning" data-toggle="tooltip" title="Cloudflare is one of the world's largest networks, and a problem considering anonymity and decentralization."><a href="https://codeberg.org/crimeflare/cloudflare-tor/"><i class="fas fa-exclamation-triangle"></i></a></span>
```
```suggestion
          <a href="https://developers.cloudflare.com/1.1.1.1/setting-up-1.1.1.1/">Cloudflare</a> <span class="badge badge-warning" data-toggle="tooltip" title="Cloudflare is one of the world's largest networks, and a problem considering anonymity and decentralization."><a href="https://codeberg.org/crimeflare/cloudflare-tor/"><i class="fas fa-exclamation-triangle"></i></a></span>
```

based on your comment I missed for a page that actually documents where to find DoT/DoH without being so much of an advertisement.

_includes/sections/dns.html
nitrohorse: Thank you!

Mikaela:
#1111 reminded me, what is the difference between an association and non-profit?

nitrohorse:
> #1111 reminded me, what is the difference between an association and non-profit?

Yeah, from my understanding in CZ.NIC’s context, it’s not technically a non-profit but rather a formal type of collective?

_includes/sections/dns.html
nitrohorse: Hmm what do you think about:
> Note: Using an encrypted DNS provider will not make you anonymous, nor hide your internet traffic from your Internet Service Provider. **But it will prevent DNS hijacking, and make your DNS requests harder to track, filter, and share with third parties.** If you are currently using Google as your DNS resolver, you should pick an alternative here.

_includes/sections/dns.html
nitrohorse: Updated to cloudflare-dns.com/dns :+1:

_includes/sections/dns.html
nitrohorse: What do you think about:
>  Firefox comes with built-in DoH support with Cloudflare set as the default resolver, but can be configured to use any DoH resolver. Currently Mozilla is conducting studies before enabling DoH by default for all US-based Firefox users.

_includes/sections/dns.html
nitrohorse: Found their docs:
- DoH:  https://developers.cloudflare.com/1.1.1.1/dns-over-https/
- DoT: https://developers.cloudflare.com/1.1.1.1/dns-over-tls/

nitrohorse: I wonder if https://cloudflare-dns.com/dns/ is good enough? Isn't super helpful because users have to dig into their docs to find these.

_includes/sections/dns.html
```diff
@@ -33,10 +33,279 @@ <h3>Worth Mentioning</h3>
...
+        </td>
+        <td>Anycast (based in <span class="flag-icon flag-icon-us"></span> US)</td>
+        <td>
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title="https://www.cloudflare.com/privacypolicy/">
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+        <td>Commercial</td>
+        <td><a data-toggle="tooltip" data-placement="bottom" data-original-title='"We will collect limited DNS query data that is sent to the resolvers. This data does not contain user IP addresses or any other personally identifiable information, and the bulk of the data is only stored for 24 hours."' href="https://developers.cloudflare.com/1.1.1.1/commitment-to-privacy/privacy-policy/privacy-policy/">Some</a></td>
+        <td>DoH, DoT, DNSCrypt, DNS-over-Tor</td>
```
nitrohorse: Added [DNS-over-Tor](https://new.blog.cloudflare.com/welcome-hidden-resolver/).

_includes/sections/dns.html
```diff
@@ -33,10 +33,279 @@ <h3>Worth Mentioning</h3>
...
+      </tr>
+    </tbody>
+  </table>
+
+  <h4>Terms</h4>
+
+  <ul>
+    <li>DNS-over-TLS (DoT) - A security protocol for encrypted DNS on a dedicated port 853.</li>
+    <li>DNS-over-HTTPS (DoH) - Similar to DoT, but uses HTTPS instead, being indistinguishable from "normal" HTTPS traffic on port 443.</li>
+    <li>DNS-over-Tor - An experimental security protocol for "not only protecting lookups from eavesdroppers but also providing anonymity for clients towards the resolver" (<a href="https://arxiv.org/pdf/1906.09682.pdf">more details)</a>.</li>
```
nitrohorse: @Mikaela - now that Cloudflare and the Foundation of Applied Privacy include DNS-over-Tor, (for consistency I changed Applied Privacy's "DNS-over-Onion" to "Tor"), I've added this new term description. Thoughts?

_includes/sections/dns.html
Mikaela: I am otherwise fine, but a small issue is that the filtering may be done by the provider and even intentionally if the user is after adblocking.

_includes/sections/dns.html
Mikaela: :+1: ~~but where does "US-based Firefox users" come from? I don't think I have heard of that criteria before connected to this issue.~~

> Users that do not wish to participate in this study can opt-out by typing “about:studies’ in the navigation bar, looking for an active study titled “Detection Logic for DNS-over-HTTPS”, and disabling it. (Not all users will receive this study, so don’t be alarmed if you can’t find it.) Users may also opt out of participating in any future studies from this page.
> 
> As always, we are committed to maintaining a transparent relationship with our users. We believe that DoH significantly improves the privacy of our users. As we move toward a rollout of DoH to all United States-based Firefox users, we intend to provide explicit mechanisms allowing users and local DNS administrators to opt-out.
from your link.

_includes/sections/dns.html
Mikaela: I don't think it's good enough to link here as they advertise their own app instead of the open source apps we suggest (including Android 9+'s native support) and they don't even tell the addresses. The only link to DoT again is in "how to integrate" section which I think is more aimed at device manufacturers or something in my opinion.

I think it would be OK to link at if we didn't have a focus on encrypted DNS, but the plain DNS makes no sense in my opinion as it could be hijacked, read and everything making it unfit as a privacy tool.

_includes/sections/dns.html
Mikaela: I think Cloudflare and Foundation of Applied Privacy mean different things with it.

For Cloudflare it's a normal DoH resolver which has a valid EV certificate.

For FoAP

> This is a TCP (not DoT or DoH) endpoint. Transport level encryption is provided by the onion layer.

I think Firefox at least is not going to accept it as it requires a valid certificate for DoH and only .onions can have a EV certificate, so I wonder if it would be better to not mention DNS-over-Tor at all as it currently seems to be restricted only for bigger players, but I guess it's interesting to follow in case the situation improves in the future.

Do you know what kind of software would take the FoAP link? I know Firefox/TorBrowser are fine with Cloudflare's and dnscrypt-proxy has a support for it.

ghbjklhv1:
What might be a better idea is to instead have a sortable table of 'no logging' OpenNic Servers.
Basically anybody can host a OpenNic server, so it is very democratic. :)

Mikaela:
> Only 2 support DNS-Over TOR: Only Cloudflare and Foundation for Applied Privacy use TOR.

And the two that do support DNS over Tor mean two different things by it (https://github.com/privacytoolsIO/privacytools.io/pull/1097/#discussion_r311923362) and the original DNS over Tor would mean

```
DNSPort [address:]port|auto [isolation flags]
    If non-zero, open this port to listen for UDP DNS requests, and resolve them anonymously. This port only handles A, AAAA, and PTR
    requests---it doesn’t handle arbitrary DNS request types. Set the port to "auto" to have Tor pick a port for you. This directive
    can be specified multiple times to bind to multiple addresses/ports. See SocksPort for an explanation of isolation flags. (Default:
    0)
```

from `man torrc`.

> None use I2p: None of the recommendations use, i2p.

Do you have any recommendations that support DNS over I2P? I am not aware of any and I think Nitrohorse would have suggested them if he knew.

> Only OpenNic has a larger list of TLDs:
> Only OpenNic supports uncommon TLDs like .bit.
> https://www.wikipedia.org/wiki/OpenNIC#OpenNIC_namespaces

This is a separate issue, but do those have valid SSL certitificates or are they all plain text? We are already recommending OpenNIC on the top of the page though.

> In my opinion, the current state of DNS sucks.

I guess there is always room for improvement.

> IMO, we need to inform them by adding tabs like supports i2p, Supports TOR, Uses Free Software, and Supports NameCoin. Preferably, requiring them to at least use/support one of these.

I don't see additional value of them.

> Edit: Why isn't OpenNic listed?

Because it's already listed. https://www.privacytools.io/providers/dns/#dns

> What might be a better idea is to instead have a sortable table of 'no logging' OpenNic Servers.
> Basically anybody can host a OpenNic server, so it is very democratic. :)

OpenNIC doesn't fullfil our requirement of supporting DNS over TLS or DNS over HTTPS, I have opened an issue at https://github.com/opennic/opennic-web/issues/68. OpenNIC servers are also already listed at https://servers.opennic.org/ (see also the previous link).

_includes/sections/dns.html
nitrohorse: Hmm that's true... the wording is a bit misleading since filtering can be by choice or not. Will think of something a bit clearer.

nitrohorse: > Note: Using an encrypted DNS provider will not make you anonymous, nor hide your internet traffic from your Internet Service Provider. But it will prevent DNS hijacking, **and make your DNS requests harder to eavesdrop on, tamper with, and share with third parties.** If you are currently using Google as your DNS resolver, you should pick an alternative here.

Not sure if that makes it a bit clearer?

_includes/sections/dns.html
nitrohorse: Yeah, it's kind of annoying that there isn't a general page with both DoH and DoT endpoints listed that's also bookmark-able; I think the best we could link to would be the DoH page: https://developers.cloudflare.com/1.1.1.1/dns-over-https/request-structure/ 

cf

Users would then see the DoT page available on the left sidebar :confused:

_includes/sections/dns.html
nitrohorse: > I wonder if it would be better to not mention DNS-over-Tor at all...

Yeah, I think that'd be best for now until we decide a more consistent way to list it without confusing users (including me :sweat_smile:).

nitrohorse:
> IMO, we need to inform them by adding tabs like supports i2p, Supports TOR, Uses Free Software, and Supports NameCoin

Thanks for the suggestions, @ghbjklhv1! I don't think a table with this criteria should override _this_ table. Rather be in addition to possibly? We'll need to iterate over and clarify DNS-over-Tor/I2P + NameCoin support more (I'm still learning), and I think having this table for encrypted ICANN DNS resolvers that support DoH/DoT/DNSCrypt is valuable for PTIO users now and a good launching point for enhancing the DNS page overall.

_includes/sections/dns.html
nitrohorse: Removed for now.

nitrohorse:
> I am not sure if the source_code file needs updating in this case as you already linked sources everywhere.

Ah, good catch -- I don't mind; I think it'll be useful to add :+1:

_includes/sections/dns.html
nitrohorse: I've updated the code with this change for easier previewing.

_includes/sections/dns.html
nitrohorse: Updated Cloudflare's link to point to https://developers.cloudflare.com/1.1.1.1/dns-over-https/request-structure/ until we figure out something better.

_includes/sections/dns.html
```diff
@@ -33,10 +33,278 @@ <h3>Worth Mentioning</h3>
...
-    <li><a href="https://quad9.net/">Quad9 DNS</a> - A non-profit, anycast DNS provider founded by <a href="https://www-03.ibm.com/press/us/en/pressrelease/53388.wss">IBM</a>, <a href="https://www.pch.net/">PCH</a>, and <a href=https://www.globalcyberalliance.org/quad9/"">Global Cyber Alliance</a>. Provides malicious domain filtering and supports DoH, DoT, and DNSCrypt. <span class="badge badge-warning" data-toggle="tooltip" title="Founders of Global Cyber Alliance include: City of London Police & Manhattan District Attorney's Office">Warnings <i class="far fa-question-circle"></i></a></span></li>
 </ul>
+
+<h1 id="icanndns" class="anchor"><a href="#icanndns"><i class="fas fa-link anchor-icon"></i></a> Encrypted ICANN DNS Providers</h1>
+
+<div class="alert alert-warning" role="alert">
+  <strong>Note: Using an encrypted DNS resolver will not make you anonymous, nor hide your internet traffic from your Internet Service Provider. But it will prevent DNS hijacking, and make your DNS requests harder to eavesdrop on, tamper with, and share with third parties. If you are currently using Google's DNS resolver, you should pick an alternative here.</strong>
```
```suggestion
  <strong>Note: Using an encrypted DNS resolver will not make you anonymous, nor hide your internet traffic from your Internet Service Provider. But it will prevent DNS hijacking, and make your DNS requests harder to eavesdrop on and tamper with. If you are currently using Google's DNS resolver, you should pick an alternative here.</strong>
```

I think sharing implies consent and there is nothing that could prevent the DNS resolver from sharing it to third parties (except their policies, but some like Quad9 do say to share anonymized data).

I guess this line is not going to be perfect, so I would be fine with merging like this too.

_includes/sections/dns.html
```diff
@@ -33,10 +33,278 @@ <h3>Worth Mentioning</h3>
...
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title='"CZ.NIC resolvers neither collect any personal data nor gather information on pages where your computer sends personal data."'>
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+        <td><a data-toggle="tooltip" data-placement="bottom" data-original-title='"CZ.NIC is an interest association of legal entities, founded in 1998 by leading providers of Internet services."' href="https://www.nic.cz/page/351/about-association/">Association</a></td>
+        <td>No</td>
+        <td>DoH, DoT</td>
+        <td>Yes</td>
+        <td>Yes</td>
+        <td data-value="N">?</td>
```
Mikaela: Should this say?

```suggestion
        <td data-value="No">?</td>
```

_includes/sections/dns.html
```diff
@@ -33,10 +33,278 @@ <h3>Worth Mentioning</h3>
...
+        </td>
+        <td>Anycast (based in <span class="flag-icon flag-icon-dk"></span> Denmark)</td>
+        <td>
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title='"Absolutely nothing is being logged, neither about the users nor the usage of this service. I do keep graphs of the total number of queries, but no personally identifiable information is saved. The data that is saved will never be sold or used for anything except capacity planning of the service."'>
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+        <td>Hobby Project</td>
+        <td>No</td>
+        <td data-value="doh">DoT</td>
```
Mikaela: data-value=doh DoT?

_includes/sections/dns.html
```diff
@@ -33,10 +33,278 @@ <h3>Worth Mentioning</h3>
...
+      </tr>
+    </tbody>
+  </table>
+
+  <h4>Terms</h4>
+
+  <ul>
+    <li>DNS-over-TLS (DoT) - A security protocol for encrypted DNS on a dedicated port 853.</li>
+    <li>DNS-over-HTTPS (DoH) - Similar to DoT, but uses HTTPS instead, being indistinguishable from "normal" HTTPS traffic on port 443.</li>
+    <li>DNSCrypt - On older yet robust method of encrypting DNS.</li>
```
```suggestion
    <li>DNSCrypt - An older yet robust method of encrypting DNS.</li>
```

source_code.md
```diff
@@ -258,21 +258,34 @@ Backend: closed-source
...
+
 DNSCrypt-proxy: https://github.com/jedisct1/dnscrypt-proxy/
 
- OpenNic: https://github.com/opennic/ (mostly)
+OpenNic: https://github.com/opennic/ (mostly)
+
 Webpage: https://github.com/opennic/opennic-web
 
-  Worth Mentioning
+### Worth Mentioning
```
Mikaela: You missed https://github.com/s-s/dnscloak ?

blacklight447-ptio:
I'm fine with the merge, on the cloudflare thing, most critisms seem to be mostly speculation. I dislike them for captchaing tor users, but they do seem to improve the situation by design alt SVC onions. They also do a lot of other good stuff like rolling out encrypted sni.

_includes/sections/dns.html
nitrohorse: Going to stay with https://developers.cloudflare.com/1.1.1.1/setting-up-1.1.1.1/ :+1:

_includes/sections/dns.html
nitrohorse: Good point! Updated to
> But it will prevent DNS hijacking, and make your DNS requests harder for third parties to eavesdrop on and tamper with.

_includes/sections/dns.html
nitrohorse: Good catch! Will update.

_includes/sections/dns.html
nitrohorse: Yeah, this is the "hack" I mentioned for "DoT" values when sorted to be grouped with "DoH" values. Don't really like it myself but it's what I could figure out for now.

PR #1114, opened 2019-08-07
 fix icon for Snowflake git service

victorhck:
## Git icon for Snowflake

Snowflake don't use GitHub service.


victorhck:
I found it updating my spanish translated page :)

PR #1129, opened 2019-08-10
Conditionally add top padding to Android addons icons

nitrohorse:
## Description

This PR hopefully increases the UX on the [Android Addons page](https://www.privacytools.io/operating-systems/#aaddons) by adding 50 pixels of top padding to the icons conditionally (using a [media query](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries)) on the "large" Bootstrap layout until the layout collapses for smaller screens.

### Before

addons-before

### After

addons-after

### Not modified (small screen layout)

addons-sm

#### Check List 

- ~~[ ] This project has an [associated discussion](https://github.com/privacytoolsIO/privacytools.io/issues).~~

### DIFF
--- a/_includes/sections/android-addons.html
+++ b/_includes/sections/android-addons.html
@@ -44,7 +44,7 @@
 
 <div class="container-fluid">
   <div class="row mb-2">
-    <div class="col-lg-3 col-sm-12">
+    <div class="col-lg-3 col-sm-12 pt-lg-5">
       <img src="/assets/img/addons/xprivacylua.png" class="img-fluid d-block mr-auto ml-auto align-middle" alt="XPrivacyLua">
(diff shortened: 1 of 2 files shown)


dawidpotocki:
Why are you making your own CSS class? No need for that, you can use utility classes from Bootstrap, like, `pt-lg-5`.

nitrohorse:
> Why are you making your own CSS class? No need for that, you can use utility classes from Bootstrap, like, `pt-lg-5`.
> 
> EDIT: Btw cleaner solution would be using flexbox instead of XYZ padding.

Oh TIL! I'm not really familiar with Bootstrap or flexbox but will look into that instead.

Issue #781, opened 2019-03-16
🆕 Software Suggestion | Replace Ricochet with Cwtch?


Mikaela:
## Basic Information

**Name:** Cwtch
**Category:**  Encrypted Instant Messenger
**URL:** https://cwtch.im/ (source: https://git.openprivacy.ca/cwtch.im/cwtch)

## Description

I get the impression that it's the continuation of Ricochet of which there is a lot of discussion in closed pull request to remove it in https://github.com/privacytoolsIO/privacytools.io/pull/476 and especially in open issue to add a warning to update the  shipped Tor binary onto it https://github.com/privacytoolsIO/privacytools.io/issues/474.

blacklight447-ptio:
Interesting, tried it out but it still seems in early alpha and not ready for practicle use yet, what about Briar? Maybe thats a replacement

Mikaela:
I think Briar is Android app and currently has no way to add contacts who aren't physically in the same space, while Richochet and Cwtch are desktop apps, so it wouldn't work as a replacement in my opinion.

blacklight447-ptio:
@Mikaela There is already a headless client in the works btw. But they are currently focused on the remote contacts feature, which I predict should be ready somewhere this year.  after that briar will become a very interesting project to follow, as it will provide e2e encryption, will be peer to peer, so have no server to seize, and will hide almost all metadata because everything hides inside the tor network( so exit nodes are not a problem either)

blacklight447-ptio:
update: briar remote contacts are now in alpha and expected to release at the end of next month.

odiferousmint:
Cwtch is commonly associated with Ricochet, but I think it would be worth noting that their primary focus *seems* to be group chat and whatnot. I am not entirely sure that they are NOT willing to sacrifice security for these features.

Mikaela:
I am currently somewhat low energy and cannot go into your comment in great detail, but do you think Cwtch would be worth considering as a worth mentioning team chat application?

odiferousmint:
> I am currently somewhat low energy and cannot go into your comment in great detail, but do you think Cwtch would be worth considering as a worth mentioning team chat application?

Yes, definitely! The mentioned "shortcomings" can be fixed, and personally I like their threat model.

blacklight447-ptio:
cwtch is still really alpha software though.

PR #1135, opened 2019-08-11
Delist Ricochet

Mikaela:
## Description

Resolves: #781

Ricochet's development appears to be stopped and in the linked issue I proposed replacing it with Cwtch where I understand the development to continue. However there @blacklight447-ptio said that it's not in usable state and proposed replacing it with Briar instead, which we list as worth mentioning, so I think Ricochet was forgotten.

How long is Ricochet going to stay secure if it's unmaintained? Are users going to follow the instructions to update Tor binary? Are users going to follow Tor's releases and keep the Tor binary up-to-date? I think keeping it listed is a disaster waiting to happen.

Also Tor is misspelled as TOR in our description which in my opinion looks bad and if we won't delist it, we should fix that. I don't know if the misspelling looks like a warning to stay away though, so maybe it can be positive.


JonahAragon:
I was under the impression Ricochet was still safe if the Tor binary was updated. It also doesn’t appear to be completely dead, it looks like development is just more focused on the protocol than the actual client.

nitrohorse:
> I was under the impression Ricochet was still safe if the Tor binary was updated. It also doesn’t appear to be completely dead, it looks like development is just more focused on the protocol than the actual client.

Based on the open [issues](https://github.com/ricochet-im/ricochet/issues) and PRs it sure looks unmaintained. Hmm... some digging...

>The reason I am reluctant to add anyone to the github team is because I know the issues that lurk in the codebase, and the amount of work required to fix them - rolling out a new legacy ricochet release with a new tor version won't fix those problems - a new release without those gives users a false sense of security.

> If there truly is desire to revive the old ricochet, I would strongly encourage you to redo both the authentication protocol and the regex handling - both are currently a source of legacy issues, and known vulnerabilities - neither are trivial to fix but If there are secure PRs for those submitted I will try and find time to review & merge them.

> If there really is willingness and effort to fund work /input energy into metadata resistant communications, I would ask you to deeply consider joining us to move Cwtch forward rather than investing effort into reviving the original Ricochet.

https://github.com/ricochet-im/ricochet/issues/600#issuecomment-511488605

In addition to Cwtch it looks like some of the community has moved to https://ricochetrefresh.net/.

PR #1154, opened 2019-08-13
[WIP] Rewrite navbar

### LINKED ISSUE #872: 🌐 Website Issue | Can't scroll down to the bottom of the 'Software'-submenu (mobile)
## Description
When I visit privacytools.io on my smartphone, the last entry of the 'Software' submenu (Productivity Tools) gets cut off, making it very hard to click on that entry(it's impossible to scroll down further). My smartphone has a 5.1" 16:9 1080p screen, but it's pretty easy to reproduce this issue on desktop browsers by making the window size smaller.

## Screenshots

Screenshot

### LINKED ISSUE #877: 🌐 Website Issue | Services dropdown menu on navbar gets cut
## Description

I was translating the website into Portuguese with a friend, and he noticed this small UI issue. Apparently, there is a bug on the navbar where it cuts the dropdown menu of Services when the screen is between 990px and ~1300px of width. I took this screenshot, where you can notice the bug on the top right corner:

## Screenshots
Screenshot_2019-04-21_01-16-05

dawidpotocki:
Fixes #877, Fixes #872, Closes #1195.

No Javascript needed, that's why I even started it :)

- [x] Dropdowns
- [x] Hide dropdown when clicking anything else
- [x] Mobile Hamburger

image

### DIFF
--- a/assets/css/style.scss
+++ b/assets/css/style.scss
@@ -183,3 +183,155 @@ footer {
     margin-right: 1rem;
   }
 }
+
+/*
+ * Navbar
+ */
+
+#navbar {
+  margin-left: auto;
+  margin-right: auto;
+  max-width: 1140px;
+  padding: .75rem 3rem;
+  color: rgba(255,255,255,0.5);
+  -webkit-touch-callout: none;
+  -webkit-user-select: none;
+  -khtml-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
+}
+
+#navbar .menu {
+  position: relative;
+}
+
+.menu {
+  display: flex;
+  justify-content: space-between;
+}
+
+.nav-anchor {
+  display: inline;
+  color: inherit;
+}
+
+.nav-anchor:hover {
+  text-decoration: none;
+}
+
+.nav-anchor:hover,
+.nav-summary:hover,
+.nav-details[open] > summary > .nav-summary {
+  color: rgba(255,255,255,0.8);
+}
+
+.nav-dropdown {
+  color: var(--dark);
+  background-color: var(--light);
+  padding: .5rem 0;
+  margin-top: .5rem;
+  position: absolute;
+  border: 1px solid #dee2e6;
+  border-radius: .25rem;
+  box-shadow: 0 .5rem 1rem rgba(0,0,0,.15);
+  z-index: 10;
+}
+
+#nav-right > details > .nav-dropdown {
+  right: 0;
+}
+
+.nav-details {
+  display: inline;
+  position: relative;
+}
+
+.nav-details > summary {
+  list-style: none;
+}
+.nav-details > summary::-webkit-details-marker {
+  display: none;
+}
+
+#nav-right > .nav-details,
+#nav-right > .nav-anchor {
+  padding-left: 0.5rem;
+}
+#nav-left > .nav-details,
+#nav-left > .nav-anchor {
+  padding-right: 0.5rem;
+}
+
+input#nav-toggle,
+#nav-toggle-label {
+  display: none;
+}
+
+#nav-home-mobile {
+  display: none;
+}
+
+
+/* Mobile hamburger menu */
+
+@media only screen and (max-width: 992px) {
+  .menu,
+  .nav-details,
+  .nav-anchor {
+    display: block;
+  }
+  .menu {
+    margin-top: 0.5rem;
+  }
+  #nav-home {
+    display: none;
+  }
+  #nav-home-mobile {
+    display: inline;
+  }
+  input#nav-toggle:checked ~ .menu {
+    display: none;
+  }
+  input#nav-toggle + label {
+    display: block;
+    cursor: pointer;
+    margin-top: 0.25rem;
+    float: right;
+  }
+  .nav-dropdown {
+    display: block;
+    position: relative;
+    max-height: 15.75em;
+    overflow-y: auto;
+    left: 0;
+    right: 0;
+  }
+  #nav-right > .nav-details,
+  #nav-right > .nav-anchor,
+  #nav-left > .nav-details,
+  #nav-left > .nav-anchor {
+    padding: 0 0;
+  }
+  #navbar {
+    padding: 1rem 3rem;
+  }
+}
+
+@media only screen and (max-width: 768px) {
+  #navbar {
+    padding: 1rem 2rem;
+  }
+}
+
+@media only screen and (max-width: 575px) {
+  #navbar {
+    padding: 1rem 1rem;
+  }
+}
+
+@media only screen and (max-height: 350px) {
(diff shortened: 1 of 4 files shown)


dawidpotocki:
> I don't understand the code, but it seems to work even if it's a bit funny how everything can be opened simultaneously.

```
<details>
	<summary>
		This is summary!
	</summary>
	This is some text
</details>
```

<details>
	<summary>
		This is summary!
	</summary>
	This is some text
</details>


Now, how I made this thing ^ work like a dropdown? I had to first create
a `<span>` after `</summary> and make `position: absolute` so it would
"fly" and not show itself in the navbar, which would destroy everything.
Next to fix bug with "Services" dropdown being cut of, I had to set
`<details>` to `positon: relative` so dropdown is shown next to it and
then `right: 0` to the dropdown so it would go to the left side.

It's pretty simple, considering that one from BS4 is using Javascript
for the same thing and everything you need is builtin to the HTML!

assets/css/style.scss
```diff
@@ -182,4 +182,38 @@ footer {
...
-}
\ No newline at end of file
+}
+
+/*
+ * Navbar
+ */
+
+.nav-text {
+  color: #9a9da0;
```
nitrohorse: Small suggestion, since we're using SASS could we use color variables for these new classes to allow the colors to be re-usable? I think like this:
```
$color-gray: #9a9da0;

.nav-text {
  color: $color-gray;
}
```

nitrohorse:
This is really cool! Though I think we'll need `Hide dropdown when clicking anything else` to be resolved; the usability without isn't great.

assets/css/style.scss
dawidpotocki: I don't think that there is need for that as it probably will not be used anywhere else.
It was the color used before in navbar, that was probably built-in to BS4.
But if you want, I can do that.

nitrohorse: Ah, that makes sense. I suppose we can make the colors re-usable if we need to reference them again in the future somewhere else.

dawidpotocki:
> Though I think we'll need Hide dropdown when clicking anything else to be resolved; the usability without it isn't great.

Got it working with JS. I don't believe it is possible without it. But at least navbar dropdowns _sort of work_, which was not the case before.

assets/css/style.scss
```diff
@@ -182,4 +182,41 @@ footer {
...
+}
+
+.nav-anchor {
+  display: inline;
+  color: inherit;
+}
+
+.nav-link:hover,
+.nav-summary:hover {
+  color: #c4c9cc;
```
JonahAragon: This `:hover` color seemingly does not apply to navbar items without dropdowns (i.e. the home button, participate link, and donate link).

assets/css/style.scss
JonahAragon: The actual color is `rgba(255,255,255,0.5)`, FYI

dawidpotocki:
It is because I renamed `.nav-link` to `.nav-anchor`, because it was
already defined by Bootstrap, so I renamed it and forgot to update it.
Should be good now.

dawidpotocki:
> > +/*
> + * Navbar
> + */
> +
> +.text-nav {
> +  color: #9a9da0;
> +}
> +
> +.nav-anchor {
> +  display: inline;
> +  color: inherit;
> +}
> +
> +.nav-link:hover,
> +.nav-summary:hover {
> +  color: #c4c9cc;
>
> This `:hover` color seemingly does not apply to navbar items without dropdowns (i.e. the home button, participate link, and donate link).

It is because I renamed `.nav-link` to `.nav-anchor`, because it was
already defined by Bootstrap, so I renamed it and forgot to update it.
Should be good now.

assets/css/style.scss
dawidpotocki: It is because I renamed `.nav-link` to `.nav-anchor`, because it was
already defined by Bootstrap, so I renamed it and forgot to update it.
Should be good now.

(if you got the same reply on emails, sorry but GitHub does not like me replying to these comments with email :C)

Perelandra0x309:
W3.CSS does nice navbars with no JS.
https://www.w3schools.com/w3css/w3css_navigation.asp

dawidpotocki:
> W3.CSS does nice navbars with no JS.

They are on hover, this one is on click. The collapsing one is using
JavaScript, so it's not really useful.

Perelandra0x309:
They do have clickable ones too.  The menu on that website itself is a clickable menu, so can be used as an example.

dawidpotocki:
> They do have clickable ones too.  The menu on that website itself is a clickable menu, so can be used as an example.

They. Require. JavaScript.

JonahAragon:
This is coming along great! I wonder if you would be able to push the dropdown box down a bit...

Old:

image

Current in PR:

image

For some reason that dropdown being so close to the navbar text is unsettling to me. Also, want to make sure changing the color of an active dropdown to white is on your to-do list. I would imagine `summary:focus` would do the trick, maybe not perfectly but good enough, unless you have other plans.

dawidpotocki:
> This is coming along great! I wonder if you would be able to push the dropdown box down a bit...
> For some reason that dropdown being so close to the navbar text is unsettling to me. Also, want to make sure changing the color of an active dropdown to white is on your to-do listw I would imagine `summary:focus` would do the trick, maybe not perfectly but good enough, unless you have other plans.

Yeah, I'm planning to do that, but my priority is making a mobile
hamburger ui now. I'm trying to implement it without JavaScript and it
seems to be possible, all the possible ways are weird hacks, but
whatever.

Perelandra0x309:
> They. Require. JavaScript.

Where do you see that?  I am using it just fine without JS.

dawidpotocki:
> Where do you see that? I am using it just fine without JS.

https://invidio.us/watch?v=mNttgpgNFS4

Perelandra0x309:
If you insist on having clickable menus (on desktop) then yes JS in needed for that one feature.  But I thought you wanted this to be JS free?  Are hoverable menus that bad?  Those do not require JS.  Also the hoverable menus from W3 are actually clickable on mobile devices.

BTW W3 is just a framework of standard CSS attributes, just makes it easier to do what you want.  You are probably doing the exact same with with your straight CSS if you are able to get hoverable menus without JS.  So it's a decision you and the site owners need to make- use JS to get clickable menus or drop JS and hav menus be hoverable on desktop (clickable on mobiles).

dawidpotocki:
> If you insist on having clickable menus (on desktop) then yes JS in needed for that one feature.

But. I. Implemented. Them. In. This. Pull. Request. Without. JavaScript.

Perelandra0x309:
Ah, OK your last update from yesterday indicated you were having to use "weird hacks" to do it without JS.  If you got it working, great!

dawidpotocki:
__Everything is finished.__

I think.

Issue #1130, opened 2019-08-10
🆕 Software Suggestion | Fedora Workstation


dawidpotocki:
**Name:** Fedora
**Category:** OS
**URL:** https://getfedora.org

I think we lack easy-to-use distribution that will work on most devices.
At the moment there are 3 operating systems for PCs listed:
* Debian
* Trisquel
* Qubes OS

### Problems with Debian
It is not the most noob-friendly distro and does comes with 100% free kernel, which means that a lot of devices will not work.

### Problems with Trisquel
Comes with 100% free kernel and 100% free repos, which means that a lot of devices will not work… and sorry, but it looks like trash.

### Problems with Qubes OS
It isn't the easiest distro to use, for sure.

### Why Fedora?
It is distro which comes with easy to use installer and _"standard"_ Linux kernel.
It means that it should work fine for most devices.
Repos are 100% free (other than Linux), but when you open GNOME Software, there is an option to enable 3rd party repo to install NVIDIA drivers, which may be needed.
Fedora uses SELinux, Flatpak (but without any repo enabled like Flathub) and Wayland (GNOME Shell) by default, which is better for security, I guess.
There are also [Fedora Spins](https://spins.fedoraproject.org/), which are just Fedora + different desktop environment, everyone can choose whatever they like.
It is developed by Fedora Project, which is sponsored by Red Hat.

### Why not just Ubuntu?
There was an Amazon controversy.
> From October 2012, it sent the user's queries through a secure HTTPS connection from the home lens to productsearch.ubuntu.com, which then polled Amazon.com to find relevant products; Amazon then sent product images directly to the user's computer through HTTP. If the user clicked on one of these results and then bought something, Canonical received a small commission on the sale.
…as the home lens is the normal means to search for content on the local machine, reviewers were concerned about the disclosure of queries that were intended to be local, creating a privacy problem.

Because of it and new system usage data collection, some people are skeptical.

### Why not Linux Mint?
Their site was compromised few times and was pointing at modified ISOs with a backdoor in it.

### Why not Manjaro?
It is not the most user-friendly OS, but better than Arch.
They hold updates (2-3 weeks), just for the sake of it (it is supposed to be more _"stable"_), even security updates.
They also forgot to update SSL certificates, twice and suggested stupid workarounds, like changing system time.
First time: https://archive.fo/JeOLo
Second time: https://archive.fo/XV70t

### Why not elementaryOS?
I don't like what developers of it done/do.
[...]

Issue #449, opened 2018-04-19
Review PC Operating Systems


c0rdis:
Let's review the section and make sure we all agree on the top-3 recommended options. It's not clear to me why Debian (?) and Trisquel are there, and Whonix / MOFO are not.

https://itsfoss.com/mofo-linux-censorship/

Shall we replace them?

Atavic:
Debian is the most important linux distro, with many others based upon it.
Trisquel GNU/Linux doesn't include proprietary software.
Qubes is a privacy/securityt distro, maybe with high hardware requirements when compared to Whonix.

c0rdis:
Well, Debian maybe the most important Linux distro with many others based upon it but I cannot see why it is necessarily the best for user's privacy. I mean, Tails is based on Debian, and we're rightly recommending it in the following section. 

Whonix is also based on Debian but it's very purpose is  "to provide privacy, security and anonymity on the internet".
MOFO has plenty of tools bundled from the box, most of which we're recommending on the page. 

Also, gNewSense/Dragora do not include proprietary software either, and it's unclear for me why it's specifically Trisquel in top-3.

Shifterovich:
Before I made some changes to the section it already included Debian, so I have no idea why and how it got there, but I kept it there. Now there are many operating systems focused on security. Some are worse than others, Qubes is obviously superior. Subgraph has/had [some flaws](https://micahflee.com/2017/04/breaking-the-security-model-of-subgraph-os/), but it's better than recommending Debian, which might lead users to think that Debian actually has some extra security features and decide to go with it as it's easier to install than Qubes.

We should remake the section. Qubes is the starting point, the rest should be figured out based on *current* available operating systems.

Trisquel was there along with Debian iirc.

angela-d:
Why the hate on for Debian?

[url=https://www.debian.org/social_contract]Debian social contract[/url]

Omitting the nonfree repositories (which is what PureOS does & is based on Debian) is as close to you can get to a free/private operating system of proprietary blobs.

Shifterovich:
Debian is not specifically focused on security. A secure OS looks like Qubes, not like Debian.

Atavic:
Debian should stay because many other distros are based on it. While debian closes the gaps in the OS, ubuntu, mint and other forks just follow.

angela-d:
[...]

PR #1156, opened 2019-08-13
Replace Trisquel with Fedora

JonahAragon:
Replaces Trisquel with Fedora as a top-3 choice, moves Trisquel to Worth Mentioning. This makes a lot more sense to me as a general recommendation IMO, when I was going through the changes for #1148 I didn't really understand why Trisquel was recommended in the first place.

I would also be open to recommending elementaryOS instead of Trisquel or Fedora despite the comments against it at #1130, if we'd prefer. But Fedora is also fine. I do think it's important to have an easier recommendation, Qubes (and arguably even Debian) is somewhat extreme.

Resolves: #1130; Resolves: #449


nitrohorse:
I have a soft-spot for elementary since it was the first Linux distro that seemed to get the UI/UX just right for me. It was also much easier to recommend to friends/family over using Windows in regards to its usability. I don't have experience with Fedora but it also sounds like a solid recommendation. I'm just wondering if elementary would bring a healthy balance to our OS recommendations that also keeps new users to Linux in mind?

gjhklfdsa1:
> elementary
@nitrohorse My issue with ElementaryOS is that they seem to be "convenience first, security second".
You can set up Pantheon on Debian (`[pantheon-debian](https://gitlab.com/pantheon-debian/pantheon-debian/wikis/home)`) but it has some issues:
https://gitlab.com/pantheon-debian/pantheon-debian/issues/26

Plus, the NX Software Center works basically everywhere:
https://github.com/Nitrux/nx-software-center

JonahAragon:
> Trisquel is a popular OS recommended by privacy experts worldwide for its ease-of-use and free/libre components and support.

Both of which are not privacy-centric features.

---

> Yum is basically only used in Fedora and Redhat. I do not understand how this benefits users in any way not provided by APT or GUIX.

Supporting a monoculture in the package manager space seems like not the best idea. Fedora, CentOS, and RHEL are **huge** distros too, so saying its "only" used in those distros understates its popularity.

---

> Fedora is for profit

As I stated at https://github.com/privacytoolsIO/privacytools.io/issues/1130#issuecomment-520963685, Fedora is a community led project:

> Saying Fedora is owned by Red Hat would be like saying Debian is owned by Canonical

RHEL "owning" Fedora seems to be a common misinterpretation, but it's pretty obviously debunked.

---

> From a UI standpoint

Fedora runs GNOME. That's basically the easiest-to-understand desktop environment around. The only better DE from a purely UI perspective (IMO) is Elementary's. Trisquel looks straight out of 2012.

---

> I have a soft-spot for elementary since it was the first Linux distro that seemed to get the UI/UX just right for me. 

Agree, which is why I'd also be fine listing it, it is fantastic software. But from especially a security perspective Fedora seems to come out on top. Also we currently recommend Debian (and Trisquel for the time being), both of which are apt-based, so I think throwing another package manager in the mix is a good thing. Elementary is basically Ubuntu with another DE and custom apps, but the underlying software is the same.

---

The main things I like about Fedora are 1) it isn't Debian-based like our other recommendations, and 2) it has a very security-focused approach. As @blacklight447-ptio stated at https://github.com/privacytoolsIO/privacytools.io/issues/1130#issuecomment-520738986: 

> also, fedora is out of the box a lot more secure then debian. it has Selinux out of the box, it applies a ton of hardening patches aswell, it also uses wayland, which is all listed above.

nitrohorse:
> from especially a security perspective Fedora seems to come out on top. Also we currently recommend Debian (and Trisquel for the time being), both of which are apt-based, so I think throwing another package manager in the mix is a good thing.

Ah, that's a good point I didn't think about. Makes sense! :+1:

ggg27:
Just to be clear:

> The project was founded in 2003 as a result of a merger between the Red Hat Linux (RHL) and Fedora Linux projects. Red Hat employees make up only 35% of project contributors, and most of the over 2,000 contributors are unaffiliated members of the community.

________________

If you want a diffrent package mangaer may I recommend GUIX of Pacman?

PR #1169, opened 2019-08-17
Remove Brave

### LINKED ISSUE #161: Remove Brave

### LINKED ISSUE #649: ❌ Software Removal | Brave

### LINKED ISSUE #758: ❌ Software Removal | Brave Browser
More evidence that Brave is just Malware to push profit towards their own cryptocurrency. 

https://glitch.social/@tessaracht/101572578668968885

"first, they sold exceptions to privacy tracking to Facebook and Twitter, now it comes out they've got a header override that allows them to set header exceptions on demand."

https://laptop-updates.brave.com/promo/custom-headers

JonahAragon:
Honestly, I don't see any advantages to using Brave over either Firefox or Tor Browser for any conceivable use-case. The Brave team also clearly [does not appreciate](https://github.com/privacytoolsIO/privacytools.io/pull/657#issue-237832186) the free advertising (after [requesting their addition to the site](https://github.com/privacytoolsIO/privacytools.io/pull/78#issuecomment-252093024) no less), so I'm not particularly inclined to keep petitioning for their inclusion 😆 

With Firefox Quantum, the security benefits of using Chromium over Firefox are largely nonexistent nowadays. If anyone has points to the contrary I'd love to hear them, otherwise I'm good to see it go.

Resolves #161, Resolves #657, Resolves #649, Resolves #758. Reverts #78.


blacklight447-ptio:
While I agree that is should no longer be included, shouldn't we open a last issue and poll the community in it?

nitrohorse:
Community discussions:
- https://old.reddit.com/r/privacytoolsIO/comments/crq532/should_ptio_delist_brave/
- https://forum.privacytools.io/t/discussion-brave-browser-candidate-for-delisting/280/10

JonahAragon:
...plus the 5 GitHub discussions linked in the OP.

mr-perseus:
Please check how often and fast the respective Chrome forks are updated. For example I checked out „Irdium“ a few weeks ago and it was based on a Chromium version released in April. Ungoogled Chromium is solo run by Eloston and while he is currently actively updating it, that could change at any time. Also he needs a few days to update it usually.

Brave pushes Chromium stable updates within hours, thats one reason why I like it. And I prefer a Chromium based browser over FF because of the sandboxing.

I just commented this so you don‘t recommend browsers which might have improved privacy but are a security nightmare.

blacklight447-ptio:
Which exactly why i investigated and recommend that we keep chromium based browsers,  the alternatives i mentioned are nowhere final and were just a few options i know at the top of my head that will require further investigation. after some chatting with the team over Wire, we came to a few conclusions:

First we will split the browser recommendations in 3 sections: android ios and desktop.
for android we will most likely list tor browser for android maybe Firefox focus/firefox and bromite (this last one was recommend by daniel micay) for ios we are looking into Snowhaze and onion browser. for desktop we keep Firefox and tor browser, while we are still searching for an chromium based alternative to Brave which is actively maintained. question now is: do we de-list brave and search for an alternative, or keep brave listed until we found an alternative. remember though that all of this is all future work which we will still need to investigate, non the options i listed above here is guaranteed to end up on the list.

blacklight447-ptio:
After discussion with the team, we have decided to remove brave for now, and start researching alternatives for chromium on desktop. I will soon create a ticket for the above described redesign of the browser page.

ciampolo:
I am one of the guys who suggested a Brave removal.

While it is true that Firefox has no real protection of its own there are two things to note here:
1. Chrome uses User namespaces to enable their sandbox which are beyond insecure and allow for so much retarded stuff to happen that couldn't ever happen without any sandbox at all (I don't know if it's still possible but you could get root rather easily just through JS explicitly thanks to Chromes use of User namespace as sandbox)
2. Any nix application run as a different user with proper rights set is more than enough of a sandbox. Put something like Firejail on top of it and this will stop virtually all attacks and it is without a doubt infinitely more secure more than running sandboxed Chrome under your own user. 

Also a side note: I tried Brave again and main vectors I listed nearly a year ago are still fingerprintable; I said it a year ago I can only say it again it is either this is intentional cash grabbing (buhh huhh Brendan) or it is incompetency of the devs which you would doubt with someone like Eich. After 4 years you finally realized  that Webkit/Blink/Chrom* is the wrong choice yet you keep going? I will forever do my best to stop anyone from ever installing the terrible scam that Brave is dont worry Mr. Eich.

Mikaela:
The user namespaces or an alternative to them are tracked in https://github.com/brave/brave-browser/issues/3420.

Another potentially interesting side note, [PRISM Break is considering Brave](https://gitlab.com/prism-break/prism-break/issues/1566#note_e7c3a12a6612c1362c5a73fcd0077e96ae008bf6).

digitalblossom:
May I ask why the community was polled when its response was mostly positive towards Brave, and then it was removed regardless?

> It would be a poor decision to delist them

> Delisting them would be pretty stupid

> I think that's where Brave comes in good at

> Brave makes sense for me

> I think you should consider keeping brave

> Fortunately, it’s completely open source

> I think it’s almost ignorant we’re even having this debate when we’re trying to introduce people to privacy

> It is a good option for who isn't extreme privacy focused

> But if it comes to browsers for me it‘s brave at the moment

> I don't advocate for brave to be removed from the recommended browser list on PTIO

> Don't remove Brave

> No, Brave shouldn't be delisted

jonaharagon:
@digitalblossom was the community polled?

The reasons Brave was removed are in the original post.

Issue #949, opened 2019-05-26
🌐 Website Issue | Wire has misleading warning, contacts are not stored in plaintext, chat participants are


Mikaela:
I read WhatsApp and Wire privacy policies earlier today and I think Wire said that they hash contacts.

I am opening this as a note to self or if someone else takes a look at the linked article and compares this.

Mikaela:
> Wire stores metadata such as your contacts in plaintext (= not encrypted).

I think our warning is misleading or misunderstood from the actual problem.

>  "Hey Wire, why does your database schema include plaintext storage of threads between users?" security researcher Thomas H. Ptáček‏ tweeted on Wednesday, with a link to some of Wire's code (Wire is open source).

* https://www.vice.com/en_us/article/gvzw5x/secure-messaging-app-wire-stores-everyone-youve-ever-contacted-in-plain-text

So Wire knows between whom chats are happening, not the actual contacts as their privacy policy says:

> If you agree to give Wire access to your address book contacts, only hashed phone numbers will be used to match you with other users. The content of your address book is never uploaded to or stored on our servers.

* https://wire.com/en/legal/#privacy

I will leave this open until this is resolved or I can find if this has been fixed.

PR #1172, opened 2019-08-17
Update Wire's metadata warning to increase clarity

nitrohorse:
## Description

This was brought up by abreast:privacytools.io in the Matrix chat room:
> The warning for Wire on ptio website is a bit misleading: "Wire stores metadata such as list of your connections/conversations in plaintext (= not encrypted)."
> Including conversations in the warning will lead people to think that their actual conversation is stored in paintext which would be incorrect.

So this PR attempts to make the warning a bit more clear.

Relates to: https://github.com/privacytoolsIO/privacytools.io/issues/949

#### Check List 


nitrohorse:
> wouldn't the word "contacts" be more clear than word "connections"?

Yeah good catch! Ideally I think this would be good:
> Wire stores some plaintext metadata about its users on their servers. This data includes profile name, username, profile picture, and users' lists of contacts and conversations.

_However_ commas (`,`) are breaking the tooltip lol so I'll try to figure out what's going on with that.

Mikaela:
> wouldn't the word "contacts" be more clear than word "connections"?

I don't think so as that would seem like phone contacts that Wire server doesn't have plaintext access to those.

>  However commas (,) are breaking the tooltip lol so I'll try to figure out what's going on with that.

This is just a guess, but how about `\,` ?

blacklight447-ptio:
> > wouldn't the word "contacts" be more clear than word "connections"?
> 
> Yeah good catch! Ideally I think this would be good:
> 
> > Wire stores some plaintext metadata about its users on their servers. This data includes profile names, profile pictures, usernames, and users' lists of contacts and conversations.
> 
> _However_ commas (`,`) are breaking the tooltip lol so I'll try to figure out what's going on with that.

if you'd change it to that, then it would be perfect IMO. ill wait for you to make the changes and figure the tooltip issue and approve it after that.

nitrohorse:
Hmm @Mikaela is right; they hash contacts but store metadata about user-to-user connections in plaintext I think...

nitrohorse:
> If you agree to give Wire access to your address book contacts, only hashed phone numbers will be used to match you with other users. The content of your address book is never uploaded to or stored on our servers. -- https://wire.com/en/legal/#privacy

_includes/cardv2.html
```diff
@@ -27,7 +27,7 @@ <h3 class="h5">
       <img src="{{include.image}}" height="120" width="120" class="panel-pic" alt="{{include.title}}">
       {{include.description}}
       {% if include.labels %}
-      {% assign labels = include.labels | replace:", ", "," | split:"," %}
+      {% assign labels = include.labels | split:";" %}
```
nitrohorse: @JonahAragon is it alright to change the delimiter in labels to a different character other than `,`?

nitrohorse: Maybe something not used normally like `|` would be better than `;`.

blacklight447-ptio: I wouldn't see any reason why not.

_includes/cardv2.html
```diff
@@ -27,7 +27,7 @@ <h3 class="h5">
       <img src="{{include.image}}" height="120" width="120" class="panel-pic" alt="{{include.title}}">
       {{include.description}}
       {% if include.labels %}
-      {% assign labels = include.labels | replace:", ", "," | split:"," %}
+      {% assign labels = include.labels | split:"|" %}
```
Mikaela: Is this going to break anything else?

nitrohorse: Good catch! I’m going to put the “wip” label on this and verify. 🙏🏼

_includes/cardv2.html
JonahAragon: You can change it to whatever, but you will need to update everything that uses the labels functionality to match. You should also change the badges delimiter to match, since it uses the same syntax as labels.

_includes/cardv2.html
JonahAragon: Most Likely

_includes/cardv2.html
nitrohorse: Okay, that's a good catch. I'll update the badges delimiter to match + validate this doesn't break anything.

nitrohorse: Alright, so I went through the repo's code and couldn't find any place where multiple badges or labels are being used in a v2 card aside from Wire. So I think we're fine with making the cardv2 delimiter change without breaking anything.

nitrohorse: I've also updated the formatting of `cardv2.html` to make editing and reading a bit more easier for future contributors.

JonahAragon: Maybe Wire was the one I was thinking of, I knew I had seen one with two labels recently lol

_includes/cardv2.html
```diff
@@ -2,88 +2,254 @@
...
-              {% assign help_icon = '<i class="far fa-question-circle"></i>' %}
-              <span class="badge badge-{{color}}" {% if tooltip %} data-toggle="tooltip" title="{{ tooltip }}">{{text}} {{ help_icon
-                  }}
-                  {% else %}
-                  >{{text}}
-                  {% endif %}
-              </span>
-          {% endfor %}
+      <span class="mr-2">{{ include.title }}</span>
+      {% assign badges = include.badges | split:"|" %}
```
nitrohorse: Updated to "|"

JonahAragon: I am somewhat surprised that works, seeing as Liquid uses "|"s internally. There's even one in that line in particular 😆

nitrohorse: Oh really? I missed that lol. Great that’s probably a bug I created...

JonahAragon: Idk, it appears to function.

nitrohorse: Ah I must have mis-understood -- I didn't realize we're using a module called _Liquid_ lol! Yeah, then this is alright since the | is within quotes so it's treated simply as a string :smile:

PR #1174, opened 2019-08-17
Reworked VPN Providers

JonahAragon:
This PR adds a lot of information regarding the factors that go into our VPN provider recommendations. 

- The criteria for inclusion has been updated.
- NordVPN has been delisted (Resolves: #1162)
- The VPN table has been removed and been replaced with a more transparent and information-filled list of recommendations.

In this PR, the only recommended provider is Mullvad. I have also chosen to list ProtonVPN and IVPN as "worth mentioning" despite not having public security audits published, a new criteria for inclusion. Because they are not full recommendations, those listings do not include links to their homepage or to purchase.

I'm considering this the final version of the VPN recommendations until the end of 2019, when we will reevaluate IVPN and ProtonVPN. Therefore, unless someone else has something to add, this PR closes #1139.

PR #1175, opened 2019-08-17
Clarify and expand VPN criteria

JonahAragon:
Building on https://github.com/privacytoolsIO/privacytools.io/pull/1174#discussion_r314960176, this PR expands our criteria to be more "complete". It also adds the number of servers and locations to each provider.


pages/providers/vpn.html
```diff
@@ -20,6 +20,118 @@
...
+      <p><strong>Best Case:</strong></p>
+      <ul>
+        <li>Operating outside the USA or other Fourteen Eyes countries.</li>
+        <li>Operating inside a country with strong consumer protection laws.</li>
+      </ul>
+    </div>
+
+    <div class="col-12">
+      <h3><span class="badge badge-info">Technology</span></h3>
+      <p>We require all our recommended VPN providers to provide OpenVPN configuration files to be used in any client. <strong>If</strong> a VPN provides their own custom client, we required a kill-switch to block network data leaks when disconnected.</p>
```
nitrohorse: Change “required” to “require”?

pages/providers/vpn.html
```diff
@@ -20,6 +20,118 @@
...
+        <li>Perfect Forward Secrecy (PFS).</li>
+        <li>Published security audits from a reputable third-party firm.</li>
+      </ul>
+    </div>
+    <div class="col-md-6">
+      <p><strong>Best Case:</strong></p>
+      <ul>
+        <li>Strongest Encryption: RSA-4096.</li>
+        <li>Perfect Forward Secrecy (PFS).</li>
+        <li>Comprehensive published security audits from a reputable third-party firm.</li>
```
nitrohorse: Should we specify a desired time since the audit was conducted? Like for these audit(s) to be “recent/relevant”?

JonahAragon: On L89 I did say `...on a repeated (yearly) basis.` which I think covers that. That would be for the best-case scenario though. As far as making it a minimum criteria, I'm not sure how many providers would be running audits frequently. We're getting to the point where if we're any more strict we can't recommend anybody.

Issue #1152, opened 2019-08-13
🌐 Website Issue | Small section on how to confirm that encrypted DNS is working


Mikaela:
## Description

I initially saw @nitrohorse suggest this and later also saw a question [on Reddit how to verify DoH working](https://www.reddit.com/r/privacytoolsIO/comments/cplstn/doh_confirmation_test/) where he had answered.

* [ ] DoH / DoT
    * General advice: check the website of your DNS provider, it may have a test telling "you are using X DNS".
    * Firefox: go to `about:networking#dns`, if the TRR field says `True`, you are using DoH.
        * Should we note here that it's normal that some first queries such as captive portal checking may be going plaintext? Is it OK that our instructions allow downgrading to insecure DNS if DoH is down? I guess as optimally the user has global DNS encryption and eDNS is just going to benefit as DoH?
* [ ] DNSCrypt-proxy
    * see above or just stop it, if you have configured it correctly, your DNS queries stop working (with the exception of possible DoH in Firefox)
* [ ] DNSSEC
    * http://dnssec.vs.uni-due.de/
* [ ] QNAME minimization
    * if you have access to the dig command, `dig +short txt qnamemintest.internet.nl`
        * @nitrohorse would you happen to remember the source for this test?
* Anything else I am forgetting?

nitrohorse:
Hmm this looks good 😄 the source of that I found from [here](https://discourse.pi-hole.net/t/unbound-and-qname-minimisation/10038/2), specifically on [slide 25](https://nlnetlabs.nl/downloads/presentations/unbound_qnamemin_oarc24.pdf).

nitrohorse:
I see DNSCrypt [also links](https://github.com/pi-hole/pi-hole/wiki/DNSCrypt-2.0#test-dnscrypt) to https://www.dnsleaktest.com for testing.

PR #1176, opened 2019-08-18
Add section to validate DNS connection

nitrohorse:
## Description

This PR adds a new section for giving users easy to follow steps to validate their encrypted DNS connection is working for them. Definitely open to any feedback to make this clearer and more readable.

@Mikaela did note in the issue:
> Should we note here [in Firefox's `about:networking#dns` step] that it's normal that some first queries such as captive portal checking may be going plaintext? Is it OK that our instructions allow downgrading to insecure DNS if DoH is down? I guess as optimally the user has global DNS encryption and eDNS is just going to benefit as DoH?

I address this by saying `If the TRR column says "true" for some fields, you are using DoH.` Maybe this could be changed to something more clearer?

Resolves: https://github.com/privacytoolsIO/privacytools.io/issues/1152 

#### Check List 

### DIFF
--- a/_includes/sections/dns.html
+++ b/_includes/sections/dns.html
@@ -290,15 +290,15 @@ github="https://github.com/jedisct1/dnscrypt-proxy"
   <ul>
     <li>DoH / DoT
       <ul>
-        <li>Check <a href="https://www.dnsleaktest.com/">https://www.dnsleaktest.com/</a>.</li>
+        <li>Check <a href="https://www.dnsleaktest.com/">DNSLeakTest.com</a>. <span class="badge badge-warning" data-toggle="tooltip" data-placement="bottom" data-original-title="If your DNS provider relies on anycast, the displayed ISP, hostname, or location may be misleading although valid."><i class="fas fa-exclamation-triangle"></i></span></li>
         <li>Check the website of your DNS provider. They may have a page for telling "you are using X DNS." Examples include <a href="https://adguard.com/en/adguard-dns/overview.html">AdGuard</a> and <a href="https://1.1.1.1/help">Cloudflare</a>.</li>
-        <li>If using Firefox's trusted recursive resolver (TRR), navigate to <code>about:networking#dns</code>. If the TRR column says "true" for some fields, you are using DoH.</li>
+        <li>If using Firefox's trusted recursive resolver (TRR), navigate to <code>about:networking#dns</code>. If the TRR column says "true" for some fields, you are using DoH. <span class="badge badge-warning" data-toggle="tooltip" data-placement="bottom" data-original-title='Some fields will say "false" depending on the the value of network.trr.mode in about:config'><a href="https://wiki.mozilla.org/Trusted_Recursive_Resolver"><i class="fas fa-exclamation-triangle"></i></a></span></li>
       </ul>
     </li>
-    <li>dnscrypt-proxy - Try the above steps or attempt to stop it. If you have configured it correctly, your DNS requests will stop working (with the exception having Firefox's TRR configured).
+    <li>dnscrypt-proxy - Check <a href="https://github.com/jedisct1/dnscrypt-proxy/wiki/Checking">dnscrypt-proxy's wiki on how to verify that your DNS is encrypted</a>.
     </li>
-    <li>DNSSEC - Check <a href="https://dnssec.vs.uni-due.de/">https://dnssec.vs.uni-due.de/</a>.</li>
-    <li>QNAME Minimization - Run <code>dig +short txt qnamemintest.internet.nl</code> from the command-line (taken from <a href="https://nlnetlabs.nl/downloads/presentations/unbound_qnamemin_oarc24.pdf">this NLnet Labs presentation</a>). You should see this display: <code>"HOORAY - QNAME minimisation is enabled on your resolver :)!"</code></li>
+    <li>DNSSEC - Check <a href="https://dnssec.vs.uni-due.de/">DNSSEC Resolver Test by Matthäus Wander</a>.</li>
+    <li>QNAME Minimization - Run <code><a href="https://en.wikipedia.org/wiki/Dig_(command)">dig</a> +short txt qnamemintest.internet.nl</code> from the command-line (taken from <a href="https://nlnetlabs.nl/downloads/presentations/unbound_qnamemin_oarc24.pdf">this NLnet Labs presentation</a>). You should see this display: <code>"HOORAY - QNAME minimisation is enabled on your resolver :)!"</code></li>
   </ul>
 
   <h3>Worth Mentioning and Additional Information</h3>
--- a/_includes/sections/dns.html
+++ b/_includes/sections/dns.html
@@ -297,7 +297,7 @@ github="https://github.com/jedisct1/dnscrypt-proxy"
     </li>
     <li>dnscrypt-proxy - Check <a href="https://github.com/jedisct1/dnscrypt-proxy/wiki/Checking">dnscrypt-proxy's wiki on how to verify that your DNS is encrypted</a>.
     </li>
-    <li>DNSSEC - Check <a href="https://dnssec.vs.uni-due.de/">DNSSEC Resolver Test by Matthäus Wander</a>.</li>
+    <li>DNSSEC - Check <a href="https://dnssec.vs.uni-due.de/">DNSSEC Resolver Test by Matthäus Wander</a> or <a href="https://www.dnssec.cz/">DNSSEC Test by CZ.NIC</a>.</li>
     <li>QNAME Minimization - Run <code><a href="https://en.wikipedia.org/wiki/Dig_(command)">dig</a> +short txt qnamemintest.internet.nl</code> from the command-line (taken from <a href="https://nlnetlabs.nl/downloads/presentations/unbound_qnamemin_oarc24.pdf">this NLnet Labs presentation</a>). You should see this display: <code>"HOORAY - QNAME minimisation is enabled on your resolver :)!"</code></li>
   </ul>
 
--- a/_includes/sections/dns.html
+++ b/_includes/sections/dns.html
@@ -285,7 +285,7 @@ github="https://github.com/jedisct1/dnscrypt-proxy"
     <li>DNSCrypt - An older yet robust method of encrypting DNS.</li>
   </ul>
 
-  <h4>How to Validate</h4>
+  <h4>How to verify DNS is encrypted</h4>
 
   <ul>
     <li>DoH / DoT
--- a/_includes/sections/dns.html
+++ b/_includes/sections/dns.html
@@ -291,7 +291,7 @@ github="https://github.com/jedisct1/dnscrypt-proxy"
     <li>DoH / DoT
       <ul>
         <li>Check <a href="https://www.dnsleaktest.com/">DNSLeakTest.com</a>. <span class="badge badge-warning" data-toggle="tooltip" data-placement="bottom" data-original-title="Your DNS provider may not appear with their own name, so compare the responses to what you know or can find about your DNS provider. Just ensure you don't see your ISP or old unencrypted DNS provider."><i class="fas fa-exclamation-triangle"></i></span></li>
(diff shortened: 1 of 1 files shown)


_includes/sections/dns.html
```diff
@@ -37,7 +37,7 @@ <h1 id="icanndns" class="anchor"><a href="#icanndns"><i class="fas fa-link ancho
   <table class="table sortable-theme-bootstrap" data-sortable>
     <thead>
       <tr>
-        <th data-sorted="true" data-sorted-direction="descending">ICANN DNS Provider</th>
+        <th data-sorted="true" data-sorted-direction="ascending">ICANN DNS Provider</th>
```
nitrohorse: Fix small issue with sorting "caret" pointing in the wrong direction on page load.

_includes/sections/dns.html
```diff
@@ -285,6 +285,22 @@ <h4>Terms</h4>
     <li>DNSCrypt - An older yet robust method of encrypting DNS.</li>
   </ul>
 
+  <h4>How to Validate</h4>
+
+  <ul>
+    <li>DoH / DoT
+      <ul>
+        <li>Check <a href="https://www.dnsleaktest.com/">https://www.dnsleaktest.com/</a>.</li>
```
Mikaela: I think this need additional information.

```suggestion
        <li>Check <a href="https://www.dnsleaktest.com/">https://www.dnsleaktest.com/</a> You should see something else than your ISP's DNS servers that is at least in the country of your chosen encrypted DNS server.</li>
```

I am not how good my suggestion is though. Also I am not sure how clear it is that Quad9 often appears as WoodyNet for people in the USA while in Finland they appear as TREX Exchange Services Oy where the node is hosted. My Keybase DNS leak test with my config file is possibly a good showcasing of this.

_includes/sections/dns.html
```diff
@@ -285,6 +285,22 @@ <h4>Terms</h4>
...
+        <li>Check the website of your DNS provider. They may have a page for telling "you are using X DNS." Examples include <a href="https://adguard.com/en/adguard-dns/overview.html">AdGuard</a> and <a href="https://1.1.1.1/help">Cloudflare</a>.</li>
```
Mikaela: I guess the dnsleak.com test is higher due to being more reliable? I am thinking on how that test may fail at times due to caching or just in general like happens to BlahDNS https://github.com/ookangzheng/blahdns/issues/42 who doesn't know why it happens. I am not sure if this comment needs to be addressed.

_includes/sections/dns.html
```diff
@@ -285,6 +285,22 @@ <h4>Terms</h4>
...
+        <li>If using Firefox's trusted recursive resolver (TRR), navigate to <code>about:networking#dns</code>. If the TRR column says "true" for some fields, you are using DoH.</li>
```
```suggestion
        <li>If using Firefox's trusted recursive resolver (TRR), navigate to <code>about:networking#dns</code>. If the TRR column says "true", then that DNS query was done using DoH.</li>
```

Can we assume that people read our insturctions and we don't need to separately warn about not all queries possibly being TRR?

image

_includes/sections/dns.html
```diff
@@ -285,6 +285,22 @@ <h4>Terms</h4>
...
+      </ul>
+    </li>
+    <li>dnscrypt-proxy - Try the above steps or attempt to stop it. If you have configured it correctly, your DNS requests will stop working (with the exception having Firefox's TRR configured).
```
```suggestion
    <li>dnscrypt-proxy - Check <a href="https://github.com/jedisct1/dnscrypt-proxy/wiki/Checking">DNSCrypt-proxy's wiki on how to verify that your DNS is encrypted</a>.
```

This discussion made me check the upstream.

_includes/sections/dns.html
```diff
@@ -285,6 +285,22 @@ <h4>Terms</h4>
...
+    </li>
+    <li>DNSSEC - Check <a href="https://dnssec.vs.uni-due.de/">https://dnssec.vs.uni-due.de/</a>.</li>
```
```suggestion
    <li>DNSSEC - Check <a href="https://dnssec.vs.uni-due.de/">dnssec.vs.uni-due.de</a>.</li>
```

I want the link to look better, but I am not sure what to call that test as they don't call themselves as "DNSSEC capability checker by ??? University" or similar.

_includes/sections/dns.html
```diff
@@ -285,6 +285,22 @@ <h4>Terms</h4>
...
+    <li>QNAME Minimization - Run <code>dig +short txt qnamemintest.internet.nl</code> from the command-line (taken from <a href="https://nlnetlabs.nl/downloads/presentations/unbound_qnamemin_oarc24.pdf">this NLnet Labs presentation</a>). You should see this display: <code>"HOORAY - QNAME minimisation is enabled on your resolver :)!"</code></li>
```
Mikaela: Should it be noted that dig is part of `bind-dnsutils` in Debian and ??? in Fedora, if it says command not found? And Windows?

_includes/sections/dns.html
nitrohorse: Yeah, I guess I'm thinking the dnsleaktest maybe isn't more reliable but applicable to _all_ providers we list? Some don't have their own test page is what I'm thinking... I guess it's implicit for users to try several of the points we make here, e.g. check dnsleaktest & see if a test page is available.

_includes/sections/dns.html
nitrohorse: Yeah, good point :thinking: the server location difference is due to Anycast, correct (which we call out in the table)? Maybe we should mention here in a warning or something about for Anycast providers, this may not be the most reliable check then...

_includes/sections/dns.html
nitrohorse: Hmm... that's a good point... there's ongoing discussions about how the about:config tweaks section could potentially go or not so maybe its worth calling out in a warning or something here about this?

_includes/sections/dns.html
nitrohorse: yeah :thinking:

_includes/sections/dns.html
nitrohorse: hmm, yeah good callout :thinking:

_includes/sections/dns.html
JonahAragon: I would literally just call it "DNSSEC Resolver Test by University of Duisburg-Essen"

JonahAragon: Alternatively, "DNSSEC Resolver Test by Matthäus Wander"

_includes/sections/dns.html
JonahAragon: On Ubuntu and Debian (and presumably Trisquel) it's just `dnsutils`. On Fedora and CentOS it is `bind-utils`. On Arch (and presumably Parabola) it's `bind-tools`.

nitrohorse: Hmm, should we just link to https://en.wikipedia.org/wiki/Dig_(command)?

_includes/sections/dns.html
nitrohorse: I'll push an update.

_includes/sections/dns.html
nitrohorse: Will push an update.

_includes/sections/dns.html
nitrohorse: I'll push an update for this.

_includes/sections/dns.html
```diff
@@ -285,6 +285,22 @@ <h4>Terms</h4>
     <li>DNSCrypt - An older yet robust method of encrypting DNS.</li>
   </ul>
 
+  <h4>How to Validate</h4>
```
```suggestion
  <h4>How to verify DNS is encrypted</h4>
```

This is not a perfect suggestion either, but "Validate" is a bit unclear on what is being validated.

_includes/sections/dns.html
```diff
@@ -285,6 +285,22 @@ <h4>Terms</h4>
...
+
+  <ul>
+    <li>DoH / DoT
+      <ul>
+        <li>Check <a href="https://www.dnsleaktest.com/">DNSLeakTest.com</a>. <span class="badge badge-warning" data-toggle="tooltip" data-placement="bottom" data-original-title="If your DNS provider relies on anycast, the displayed ISP, hostname, or location may be misleading although valid."><i class="fas fa-exclamation-triangle"></i></span></li>
```
Mikaela: It's not only the case with anycast or at least I think DNSLeaktest showed BlahDNS Swizerland as etna.switch.cz or similar.

```suggestion
        <li>Check <a href="https://www.dnsleaktest.com/">DNSLeakTest.com</a>. <span class="badge badge-warning" data-toggle="tooltip" data-placement="bottom" data-original-title="Your DNS provider may not appear with their own name, so compare the responses to what you know or can find about your DNS provider. The main point is that you shouldn't see your ISP or old unencrypted DNS provider."><i class="fas fa-exclamation-triangle"></i></span></li>
```

JonahAragon: The tests will always show the owner of the IP address block, which is typically the upstream ISP of the VPN provider. This suggestion is worded fine though I think. However, I would change "The main point is that you shouldn't" to "Just ensure you don't".

_includes/sections/dns.html
```diff
@@ -285,6 +285,22 @@ <h4>Terms</h4>
...
+        <li>Check the website of your DNS provider. They may have a page for telling "you are using X DNS." Examples include <a href="https://adguard.com/en/adguard-dns/overview.html">AdGuard</a> and <a href="https://1.1.1.1/help">Cloudflare</a>.</li>
+        <li>If using Firefox's trusted recursive resolver (TRR), navigate to <code>about:networking#dns</code>. If the TRR column says "true" for some fields, you are using DoH. <span class="badge badge-warning" data-toggle="tooltip" data-placement="bottom" data-original-title='Some fields will say "false" depending on the the value of network.trr.mode in about:config'><a href="https://wiki.mozilla.org/Trusted_Recursive_Resolver"><i class="fas fa-exclamation-triangle"></i></a></span></li>
+      </ul>
+    </li>
+    <li>dnscrypt-proxy - Check <a href="https://github.com/jedisct1/dnscrypt-proxy/wiki/Checking">dnscrypt-proxy's wiki on how to verify that your DNS is encrypted</a>.
+    </li>
+    <li>DNSSEC - Check <a href="https://dnssec.vs.uni-due.de/">DNSSEC Resolver Test by Matthäus Wander</a>.</li>
```
nitrohorse: Also found https://www.dnssec.cz/ by CZ.NIC.

nitrohorse: Added as a second option.

_includes/sections/dns.html
```diff
@@ -285,6 +285,22 @@ <h4>Terms</h4>
...
+        <li>Check the website of your DNS provider. They may have a page for telling "you are using our DNS." Examples include <a href="https://adguard.com/en/adguard-dns/overview.html">AdGuard</a> and <a href="https://1.1.1.1/help">Cloudflare</a>.</li>
+        <li>If using Firefox's trusted recursive resolver (TRR), navigate to <code>about:networking#dns</code>. If the TRR column says "true" for some fields, you are using DoH. <span class="badge badge-warning" data-toggle="tooltip" data-placement="bottom" data-original-title='Some fields will say "false" depending on the the value of network.trr.mode in about:config'><a href="https://wiki.mozilla.org/Trusted_Recursive_Resolver"><i class="fas fa-exclamation-triangle"></i></a></span></li>
+      </ul>
+    </li>
+    <li>dnscrypt-proxy - Check <a href="https://github.com/jedisct1/dnscrypt-proxy/wiki/Checking">dnscrypt-proxy's wiki on how to verify that your DNS is encrypted</a>.
```
Mikaela: Does the CZ NIC page have English somewhere or do we assume that everyone can read Czech? :smile: 

I am learning it and can somewhat understand the gist on having DNSSEC, but I am not entirely sure what it's trying to tell me.

nitrohorse: Hmm so they don't have an English translation... yeah, maybe this is more confusing to add. From my understanding is that as long as there's a big green lock icon next to "DNSSEC", DNSSEC is enabled :smile: But I think the original link is fine to have. I'll remove the CZ.NIC link.

PR #1177, opened 2019-08-18
Add contrib badge to Arch Linux

dawidpotocki:
- [x] &zwnj;


ggg27:
@nitrohorse @Mikaela This makes it seem like Parabola is non-free.

PR #1178, opened 2019-08-18
Indicate which DNS providers support DoT over port 443

nitrohorse:
## Description

This PR adds a tooltip to the providers who support DoT over port 443 in addition to the expected port 853. This is an important indication because like DoH, DoT/443 makes eavesdropping more difficult for third parties and provides the benefit of not easily being blocked by firewalls compared to 853.

Resolves: #none

#### Check List 

- ~~[ ] This project has an [associated discussion](https://github.com/privacytoolsIO/privacytools.io/issues).~~ (via Wire chat)

### DIFF
--- a/_includes/sections/dns.html
+++ b/_includes/sections/dns.html
@@ -85,7 +85,7 @@ github="https://github.com/jedisct1/dnscrypt-proxy"
         </td>
         <td>Hobby Project</td>
         <td>No</td>
-        <td>DoH, <span data-toggle="tooltip" data-placement="bottom" data-original-title="Supports port 443 in addition to 853" class="text-success">DoT</span>, DNSCrypt</td>
+        <td data-value="dot/443">DoH, <span data-toggle="tooltip" data-placement="bottom" data-original-title="Supports port 443 in addition to 853" class="text-success"><strong>DoT</strong></span>, DNSCrypt</td>
         <td>Yes</td>
         <td>Yes</td>
         <td>Ads, trackers, malicious domains <span class="badge badge-warning" data-toggle="tooltip" data-original-title="And some wildcard, IDN, and non-ASCII domains."><a href="https://github.com/ookangzheng/blahdns#default-blocked-wildcard-domain"><i class="fas fa-exclamation-triangle"></i></a></span></td>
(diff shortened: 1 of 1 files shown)


nitrohorse:
Hmm, these are good points :thinking: 

> I think the colour is too light or difficult to spot

Yeah that makes sense, what if I also bold the text instead of relying on just a color change? Will push an update to try it.

> 
> I wish the sorting preferred DoT in port 443 over DoT in port 853 due to below
> 

Okay, I'll give those three providers' protocol field a custom value so they're grouped together while everything else sorts by length? I'll push an update for this.

> Could the terms be updated to clarify this? I am thinking that DoT should mention "...which is often blocked by restrictive firewalls, while 443 generally works everywhere"?

Good callout; will do :+1: 
>       
> I started wondering about DPI that can distinguish SSH from https, but can it distinguish TLS encrypted traffic from other https traffic?

That's a good question; I'll do some digging into this.

JonahAragon:
I wonder if green is the best color choice, does it imply that configuration is more secure than others? I'd probably prefer using `text-info` or maybe `text-primary`. Or even just black with bold text.

nitrohorse:
> I wonder if green is the best color choice, does it imply that configuration is more secure than others? I'd probably prefer using `text-info` or maybe `text-primary`. Or even just black with bold text.

I've removed the color; you're right; DoT/443 isn't necessarily more secure from my understanding, but doesn't stand out as much. In other words, it's more difficult to eavesdrop on and block.

Mikaela:
> In other words, it's more difficult to eavesdrop on and block.
> 
> @Mikaela this makes me think our description for DoT/443 in the terms could mention that eavesdropping is more difficult. Currently it says:

What do you mean by eavesdropping? The queries are encrypted from you to the DNS server and no one else can read them.

nitrohorse:
@Mikaela ah sorry, by eavesdropping, I mean DoT/443 could potentially be more difficult for your ISP/3rd parties to _analyze_ and _detect_ than DoT/853, yeah?

Mikaela:
> by eavesdropping, I mean DoT/443 could potentially be more difficult for your ISP/3rd parties to analyze and detect than DoT/853, yeah?

Possibly yes, but I am not comfortable making that claim personally until someone wiser tells me so

Mikaela:
As I went jabbering about this [on the forum](https://forum.privacytools.io/t/dns-over-https-doh-vs-dns-over-tls-dot-where-do-you-stand/1433?u=mikaela) could we have a review and merge? :)

_includes/sections/dns.html
```diff
@@ -150,7 +150,7 @@ <h1 id="icanndns" class="anchor"><a href="#icanndns"><i class="fas fa-link ancho
         </td>
         <td>Hobby Project</td>
         <td>No</td>
-        <td>DoH, DoT, DNSCrypt</td>
+        <td data-value="dot/443">DoH, <span data-toggle="tooltip" data-placement="bottom" data-original-title="Supports port 443 in addition to 853"><strong>DoT</strong></span>, DNSCrypt</td>
```
dawidpotocki: What about adding question mark next do DoT?
It is hard to tell that there is hidden info.

PR #1204, opened 2019-08-24
Add dark theme

### LINKED ISSUE #1151: ✨ Feature Suggestion | Dark Theme
When I go to ptio, my eyes are dying, just like on most websites.
My suggestion is to add dark theme.
We could use `prefers-color-scheme` CSS media query, which would automatically set preferred by visitor color scheme.
Here you can see table of browsers that support it: https://caniuse.com/#feat=prefers-color-scheme.
To change them I would suggest adding 🌙/☀ button to the right corner of navbar.
Adding it should not be that hard, so if everyone will agree, I will start implementing it.

Thanks.

dawidpotocki:
Fixes #1151

- [x] Add a switch
- [x] Remove white background from images (https://github.com/privacytoolsIO/privacytools.io/pull/1210)

### How to test?

* Install Firefox Developer Edition (up-to-date Firefox and Safari (https://github.com/privacytoolsIO/privacytools.io/pull/1204#issuecomment-524510658) should also be fine):
https://www.mozilla.org/en-US/firefox/developer/

* Firefox gets data from operating system to decide if user prefers dark themes, so set it to dark in system settings or execute it like this (Adwaita is included with GTK, so you should have it installed):

White theme:
```
$ GTK_THEME=Adwaita ./firefox-developer/firefox
```
Dark theme:
```
$ GTK_THEME=Adwaita-dark ./firefox-developer/firefox
```

Or you can go to `about:config` and add a property named `ui.systemUsesDarkTheme`, `integer` and set it to `1`.
__Note:__ `privacy.resistFingerprinting` must be false.

After you do that, there is an option to open devtools, go to Style Editor and disable `dark.css` to compare it with regular light version.
image

### Screenshot

screenshot

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -44,4 +44,5 @@
 
   <!-- CSS stylesheets -->
   <link href="/assets/css/style.css?v=5" rel="stylesheet">
+  <link id="dark-css" href="/assets/css/dark.css?v=1" rel="stylesheet" media="(prefers-color-scheme: dark)">
 </head>
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -143,6 +143,9 @@
         <a href="/donate/" class="nav-anchor">
           Donate <span class="fas fa-heart text-danger"></span>
         </a>
+        <a id="nav-switch-theme" class="nav-anchor" href="javascript:void(0)" onClick="changeColorScheme()">
+          Theme <span class="nav-theme-icon fas"></span>
+        </a>
       </div>
     </div>
   </nav>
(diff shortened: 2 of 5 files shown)


JonahAragon:
FYI for visitors who don't want to or can't change their OS theme, dark mode can be toggled in the Safari developer console (on Catalina. Don't know about Mojave support, maybe only in Safari Tech Preview on Mojave?):

image

...and in Firefox I don't know how to do it per-site, but you can update the style used in the browser by going to `about:config` and adding a new property `ui.systemUsesDarkTheme` of integer type with value `1`.

PR #1209, opened 2019-08-24
Remove translation with broken TLS

dawidpotocki:
image
image
image
image
image
image
image
image
image
image
image
image


nitrohorse:
Wonder if the owner of the site knows :eyes:

dawidpotocki:
> Wonder if the owner of the site knows :eyes:

Probably not, when you go through HTTP, website still has single-page
design.

nitrohorse:
Also [looks like](https://fediverse.network/all) the domain hosts https://peertube.dreads-unlock.fr/ but that's also inaccessible.

PR #1210, opened 2019-08-24
Remove white backgrounds, update and remove (not used) images

dawidpotocki:
Needed for https://github.com/privacytoolsIO/privacytools.io/pull/1204


dawidpotocki:
> LGTM, however StartPage image seems to have changed entirely, is that a correct logo?

Yes, it is a correct one, full logo was too wide and looked bad.

nitrohorse:
> LGTM, however StartPage image seems to have changed entirely, is that a correct logo?
> Yes, it is a correct one, full logo was too wide and looked bad.

Not sure if that’s really their logo or just what they use for their favicon. Same for SearX, we don’t use their magnifying glass favicon here either.

dawidpotocki:
> Not sure if that’s really their logo or just what they use for their favicon.

https://twitter.com/StartPageSearch

PR #1234, opened 2019-08-28
Add Resolve-DnsName to QNAME Minimization under DNS validation section

Strappazzon:
## Description

`dig` on Windows doesn't exist so I added [`Resolve-DnsName`](https://docs.microsoft.com/en-us/powershell/module/dnsclient/resolve-dnsname?view=win10-ps).

#### Check List

* Netlify preview for the mainly edited page: https://deploy-preview-1234--privacytools-io.netlify.com/providers/dns/

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -32,8 +32,7 @@ Note: This criteria applies to all of privacytools.io
 - supports DoH or DoT (We love DNSCrypt, but there is already https://github.com/DNSCrypt/dnscrypt-resolvers which is directly supported by dnscrypt-proxy, so we don't consider useful to list providers only supporting it).
 - supports DNSSEC (https://dnssec.vs.uni-due.de/ can test your current DNS provider).
 - doesn't log IP addresses during normal operation (If your suggestion logs, please compare its privacy policy with other servers on our table that keep logs).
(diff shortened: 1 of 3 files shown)


nitrohorse:
> I think the link makes it a bit more difficult to see that Resolve-DnsName is part of the command, but this isn't a blocker for me as it still has monospacing...

Yeah, I wonder if we should just remove the hyperlinks from `dig` and `Resolve-DnsName`? I suspect users would research themselves more about these commands if they run into issues?

Strappazzon:
> * Would you mind updating also the two other instances where `dig` is mentioned?
>    * https://github.com/privacytoolsIO/privacytools.io/blob/master/.github/CONTRIBUTING.md#dns
>    * https://github.com/privacytoolsIO/privacytools.io/blob/master/.github/ISSUE_TEMPLATE/7_DNS_provider.md (a HTML comment so invisible in the preview)

I made the requested changes.

> Yeah, I wonder if we should just remove the hyperlinks from `dig` and `Resolve-DnsName`? I suspect users would research themselves more about these commands if they run into issues?

I agree, I removed both links.

Mikaela:
@JonahAragon @BurungHantu1605 Any idea when you could check this? This is stuck on you due to codeowners.

Issue #1221, opened 2019-08-25
dns/#icanndns: list the hosting provider / CDN


Mikaela:
#1218 points that 5/11 of the DNS resolvers we list are behind Cloudflare.  In the team chat there is concensus that it may not be so good idea to only focus on Cloudflare while there are a lot other CDN providers too that are more invisible and I view that if every DNS provider is hosted with e.g. DigitalOcean, it's not much better for decentralization.

Thus I propose instead adding a new column for hosting provider where it can be clearly distinguished ~~(some like Quad9 might have "multiple" or "several" due to being at individual IXPs and I am only aware of DNS Leak Test naming it as WoodyNet and TREX)~~.

*Edit: the strikethroughed part doesn't apply with the testing method that is as accurate as `whois <ip address`.*

*Assigning @nitrohorse who is author of the table and knows it the best.*

nitrohorse:
Here's a some info we could use for the providers currently listed:

AdGuard 
  - IPv4
    - RU-MNETWORKS-20111214 (Serveroid, LLC)
  - IPv6
    - RU-MNETWORKS-20130326 (Serveroid, LLC)

BlahDNS
  - `whois 2a0a:e5c0:2:2:0:c8ff:fe68:bf48`
    - Data_Center_Light_Linthal (Data Center Light)
  - `whois 2001:19f0:7001:1ded:5400:01ff:fe90:945b`
    - CHOOPA (Choopa, LLC)
    - NET-2001-19F0-7001-48 (Vultr Holdings, LLC)
  - `whois 108.61.201.119`
    - CHOOPA-NETBLK08 (Choopa, LLC)
    - NET-108-61-200-0-23 (Vultr Holdings, LLC)
  - `whois 159.69.198.101`
    - RIPE (RIPE Network Coordination Centre)
    - DE-HETZNER-19920313 (Hetzner Online GmbH)

Cloudflare
  - `whois 1.1.1.1`
    - APNIC-LABS (APNIC Research and Development)

CZ.NIC
  - `whois 193.17.47.1`
    - CZ.NIC (self)
  - `whois 185.43.135.1`
    - CZ-NIC-ODVR (self)

dnswarden
  - IPv4
    - APNIC-AP (APNIC)
  - IPv6
    - DE-HETZNER-20071010 (Hetzner Online GmbH)

Foundation for Applied Privacy
  - `whois 37.252.185.232`
    - AT-IPAX-20120420 (IPAX OG)
[...]

PR #1261, opened 2019-09-02
Add hosting providers to DNS table

nitrohorse:
## Description

Adds a new "hosting provider" column to the DNS table. Also attempted to give the "filtering" column more width for readability. Can add a separate "hosting provider location" column in this or a separate PR.

Resolves: https://github.com/privacytoolsIO/privacytools.io/issues/1221

#### Check List 

* Netlify preview for the mainly edited page: https://deploy-preview-1261--privacytools-io.netlify.com/providers/dns/#icanndns


_includes/sections/dns.html
```diff
@@ -114,12 +127,37 @@ <h1 id="icanndns" class="anchor"><a href="#icanndns"><i class="fas fa-link ancho
...
+            <a href="https://www.choopa.com/">Choopa, LLC</a>,
+          </span>
+          <span class="no-text-wrap">
+            <a href="https://www.datacenterlight.ch/">Data Center Light</a>,
+          </span>
+          <span class="no-text-wrap">
+            <a href="https://www.hetzner.com/">Hetzner Online GmbH</a>,
+          </span>
+          <span class="no-text-wrap">
+            <a href="https://www.ripe.net/">RIPE Network Coordination Centre</a>,
```
Mikaela: I think this may be wrong and the IP range just didn't have owner specified or something.

_includes/sections/dns.html
```diff
@@ -202,8 +242,20 @@ <h1 id="icanndns" class="anchor"><a href="#icanndns"><i class="fas fa-link ancho
...
-        <td>Based on server choice</td>
+        <td>
+          <span class="no-text-wrap">
+            Based on server choice
+          </span>
+        </td>
         <td>?</td>
+        <td>
+          <span class="no-text-wrap">
+            <a href="https://www.apnic.net/">APNIC</a>,
```
Mikaela: I think this may be wrong and the IP range just didn't have owner specified or something.

_includes/sections/dns.html
```diff
@@ -328,6 +404,14 @@ <h1 id="icanndns" class="anchor"><a href="#icanndns"><i class="fas fa-link ancho
             <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
           </a>
         </td>
+        <td>
+          <span class="no-text-wrap">
+            <a href="https://www.ripe.net/">RIPE Network Coordination Centre</a>,
```
Mikaela: I think this may be wrong and the IP range just didn't have owner specified or something.

_includes/sections/dns.html
```diff
@@ -374,8 +468,20 @@ <h1 id="icanndns" class="anchor"><a href="#icanndns"><i class="fas fa-link ancho
...
+            Based on server choice
+          </span>
+        </td>
         <td>?</td>
+        <td>
+          <span class="no-text-wrap">
+            <a href="https://www.digitalocean.com/">DigitalOcean, Inc.</a>,
+          </span>
+          <span class="no-text-wrap">
+            <a href="https://www.ripe.net/">RIPE Network Coordination Centre</a>
```
Mikaela: I think this may be wrong and the IP range just didn't have owner specified or something.

nitrohorse:
> I think it looks good overall, but some comments in case there is anything to do towards them:
> 
>     * RIPE and APNIC are who distribute IP addresses initially, I don't think they are hosting providers so I would prefer to instead say Unknown or ??? or maybe even reach out to those providers and ask, if their websites don't reveal the information?
> 
>     * Vertical scrollbar appears on my system, but it really cannot be helped with the amount of detail that we have.
> 
>     * The table headers aren't on a straight line, but I am not sure if anything can be done to that.
> 
> 
> Otherwise I am surprised at how the amount of Cloudflare has dropped, maybe the providers using it only have their websites.
> 
> Thanks a lot purple_heart

Good points :+1: Scroll bars are inevitable if we want this table to remain readable I think. And the table headers are annoyingly on different lines for some columns due to the cell _values_ having a shorter length than the column title. Maybe this can be explored further in a separate PR.

Issue #1184, opened 2019-08-19
✨ Feature Suggestion | <wifi|wlan>.privacytools.io (openwireless.org revived)?


Mikaela:
## Description:

I was thinking about open WLAN networks and https://openwireless.org/ and it's issues and started wondering at the team chat if it's something PTIO could do. I am not sure how to format the issue, so I am quoting myself from Wire team chat:

*if you aren't familiar with [the Open Wireless Movement, you may want to check openwireless.org](https://openwireless.org/)*

> Silly idea: https://wlan.privacytools.io/ or similar to https://openwireless.org/ with actually having a community? And fixing it's issues like _nomap and having 5Ghz WiFi?
> 
> I should probably forum it though
>
> see also:
> * https://github.com/EFForg/OpenWireless/issues?utf8=✓&q=author%3AMikaela+
> * https://github.com/EFForg/OpenWireless/issues?utf8=✓&q=commenter%3AMikaela+
> 
> I think the base could be similar to openwireless, but 5GHz would need to be docuemnted as something like wlan.privacytools.io_fast_nomap and there should be the be nice badge and probably some tools on staying safe on the open WLAN like linking to our encrypted DNS page as otherwise the hotspot owner can control the traffic and maybe a suggestion to use Tor Browser
> 
> I have given up hope with openwireless a long time ago judging by the issue tracker and my WLAN has been Piraattipuolue.fi_nomap and now stationary piraattipuolue.fi_nomap advertising PPFI, but I have a properish router now that can have multiple SSIDs and happened to think about it again as it would be nice if the openwireless idea could be reality and WLAN roaming just worked

@nitrohorse commented:

> That’s a really cool idea; til about openwireless.org. Now I’m sad it’s no longer a focus for the EFF 😢

and I continued:

> Oh, another problem is what to do if there are multiple similar SSIDs in close proximity

replied to @nitrohorse :

> It never was for more than two blog posts or so, judging by me on the website
> 
> I have apparently last talked about it in January and gotten suggested to start a movement in Helsinki, but I didn't get pirates interested, because we were simultaneously in the middle of elections
> https://github.com/EFForg/OpenWireless/issues/305#issuecomment-456495091

@nitrohorse :

> yeah I went through the issues you posted which were good questions
>  We should create an issue to track this; i like this idea.

me again:

> 👍 I will after returning home, currently in supermarket

and here is the issue.

Mikaela:
[...]

PR #1279, opened 2019-09-07
privacy-resources: delist Keybase, list Open Wireless Movement

Mikaela:
## Description

* openwireless.org, see also: https://github.com/privacytoolsIO/privacytools.io/issues/1184
* Keybase: we list it on other sections and the description is a bit unhelpful.

>  <li><a href="https://keybase.io/"><strong>Keybase.io</strong></a> - Get a public key, safely, starting just with someone's social media username.</li>

While there is a "search Keybase on top", I think this description should either be removed or changed a lot as Keybase's homepage doesn't advertise this capability and instead advertises itself as kind of a team chat and file sync.

* https://deploy-preview-1279--privacytools-io.netlify.com/#resources


Mikaela:
I don't get notifications of reactions, but I assume you want some arguments here why it's in-scope rather than on the website.

* Founders of Open Wireless Movement include Electronic Frontier Foundation
* [openwireless.org](https://openwireless.org) enforces that IP addresses don't map into individual people
* MAC address randomization doesn't help with your WISP, while it may help with open WLAN networks. They also refer to a carrier that has at least used to log traffic of their customers for advertising purpouses
* Useful information on open WLAN networks including that encrypted ones won't protect against a determined attacker

Issue #1301, opened 2019-09-12
❌ Software Removal | Privacy Badger


dawidpotocki:
Privacy Badger does not do anything beyond what uBlock Origin/uMatrix/NoScript does.
If we don't want to remove it, it should at least be listed last.

ghbjklhv1:
I have a bit of an issue with this as not all advertisements are bad (_just most_).
Even DuckDuckGo has advertisements and you can get privacy respecting ads (AdEx is interesting).

__________________________________

As for NoScript and uMatrix, they are too hard to use.
It is too difficult to manually allow basically every website.

nitrohorse:
I think Privacy Badger does fill a specific use case for some users, specifically when using uBO with blocklists rather than allowlists.
> Privacy Badger uses heuristics to locate trackers and then blocks them. Basically, this means that Privacy Badger picks up where Ublock Origin leaves off. Should a tracker not located on the Ublock Origin blacklist find its way to your browser, Privacy Badger's heuristics will catch it and block it. - https://www.techrepublic.com/article/how-to-use-ublock-origin-and-privacy-badger-to-prevent-browser-tracking-in-firefox/

Further context: https://www.eff.org/privacybadger/faq#How-does-Privacy-Badger-work

Since we don't document a recommendation for which method to use (blocklist vs. allowlist) then it seems safer to leave Privacy Badger listed. But moving it to the bottom or at least lower than uBO/uMatrix makes sense to me since those are more critical in my opinion.

blacklight447-ptio:
I vote for removal.

Mikaela:
I vote for keeping.

Privacy Badger is the only extension we list that non-experts can use that doesn't block based on a external blocklist and instead learns what trackers are by itself.

µBlock Origin depends on external blocklists, µMatrix and noScript would be closer to PB by blocking everything, but they require too much configuration for basic users.

Mikaela:
I opened #1306 which does the suggestion of moving it on top of "for everyone" addons (I am not comfortable putting it below noScript and µMatrix as those two are under a separate heading "for power users").

See also: https://www.eff.org/privacybadger/faq#Is-Privacy-Badger-compatible-with-other-extensions,-including-other-adblockers

Mikaela:
What would you all say on puttng the list to alphabetical order?

See also: https://github.com/privacytoolsIO/privacytools.io/pull/1306/files#r324430513

nitrohorse:
[...]

PR #1306, opened 2019-09-13
browser-addons: put Privacy Badger on bottom

Mikaela:
## Description

Resolves: #1301 where moving Privacy Badger on the bottom was given as a requirement for keeping it.

* Netlify preview for the mainly edited page: https://deploy-preview-1306--privacytools-io.netlify.com/browsers/#addons

### DIFF
--- a/_includes/sections/browser-addons.html
+++ b/_includes/sections/browser-addons.html
@@ -5,17 +5,6 @@
 </div>
 
 
-{% include cardv2.html
-title="Privacy Badger: Stop Tracking"
-image="/assets/img/addons/privacy-badger.png"
-description="<strong>Privacy Badger</strong> is a browser add-on that stops advertisers and other third-party trackers from secretly tracking where you go and what pages you look at on the web. Privacy Badger learns about trackers as you browse."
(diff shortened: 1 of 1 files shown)


_includes/sections/browser-addons.html
```diff
@@ -90,6 +79,17 @@ <h1 id="addons" class="anchor"><a href="#addons"><i class="fas fa-link anchor-ic
 chrome="https://chrome.google.com/webstore/detail/snowflake/mafpmfcccpbjnhfhjnllmmalhifmlcie"
 %}
 
+{% include cardv2.html
```
nitrohorse: I'm wondering if instead at the bottom, we should move it below Decentraleyes? Or if we decide to move it, should we just reorder the list based on importance for the average user?

Mikaela: I think order of most importance is something that we could debate forever without reaching a conclusion. --> #1301

nitrohorse: Yeah, that's true :bulb:

PR #1317, opened 2019-09-15
Remove "Contact" from navbar

dawidpotocki:
I don't see any other solution to this (no, changing paddings is not the solution).

image


nitrohorse:
Currently it’s in the footer and it makes sense that it should remain there. An About page would make more sense in the header but we don’t have that yet.

PR #1337, opened 2019-09-22
teamchat: link to ToSDR instead of web.archive.org

Mikaela:
I am just wondering if it's a better link than an web.archive.org link to a closed feature request for a single privacy feature where they say to not have any plans for it, while there are more things wrong with them.


nitrohorse:
What about instead of replacing the archive link we just add the ToS;DR link?

Mikaela:
Which word would you make the other link or add for it?

nitrohorse:
Ah, you're right; I didn't realize "Discord" itself was a hyperlink. Yeah, then I think the ToS;DR link would be better overall.

Issue #1207, opened 2019-08-24
🗄️ DNS provider suggestion | Snopyta DNS


Mikaela:
## Basic Information

**Name:** Snopyta
**Filtering:** no
**Privacy policy:** https://snopyta.org/privacy_policy (doesn't mention DNS / IP logging?)
**Protocols:** DoT (853), DoH
**Server/Location:** Finland
**Source:** -
**Type:**  Informal collective
**Website:** https://snopyta.org/service/dns/index.html

## Description

I find their DNS server being in Finland interesting (as the other DNS providers I am aware of are Cloudflare and Quad9 anycast) and they are part of Librehosters (#816) where they were requested to be listed earlier.

I am hoping @Perflyst (please tell me my name memory works and you are the Snopyta admin) can fill us in on the features and list them on the website assuming they are supported.

### Required features:

* [x] supports DoH or DoT 
* [x] supports DNSSEC 
* [x] doesn't log IP addresses during normal operation 

#### Desired features

* [x] supports QNAME minimization

Perflyst:
DNSSEC is supported and of course I do not log anything (expect needed for debugging).
QNAME minimization is enabled.

Please do **not** add the service yet to your website. I want to add it to the privacy policy first and change some setup things in the background.

Perflyst:
Added it to the privacy policy: https://git.snopyta.org/snopyta/snopyta.org/commit/210a3cc8ccd31ba974fefa4e19685b3ce48b7d5d

If this policy is okay for privacytools, feel free to add the it to your website now.

PR #1339, opened 2019-09-23
DNS: add Snopyta

Mikaela:
## Description

Resolves: #1207

#### Check List 

- [x] This project is [free/libre software](https://www.wikipedia.org/wiki/Free_software).

* Netlify preview for the mainly edited page: https://deploy-preview-1339--privacytools-io.netlify.com/providers/dns/#icanndns

### DIFF
--- a/_includes/sections/dns.html
+++ b/_includes/sections/dns.html
@@ -464,6 +464,43 @@
       </tr>
 
       <tr>
+        <td data-value="Snopyta">
+          <a href="https://snopyta.org/service/dns/index.html">Snopyta</a>
+        </td>
+        <td>
+          <span class="no-text-wrap">
+            <span class="flag-icon flag-icon-fi"></span>
+            Finland
+          </span>
+        </td>
+        <td>
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title="https://snopyta.org/privacy_policy/" href="https://snopyta.org/privacy_policy/">
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+        <td>
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title='Part of LibreHosters, "a network of cooperation and solidarity that uses free software to encourage decentralisation through federation and distributed platforms."' href="https://libreho.st/">
+            Informal collective
+          </a>
+        </td>
+        <td>No</td>
+        <td>DoH, DoT</td>
+        <td>Yes</td>
+        <td>Yes</td>
+        <td>
+          <span class="no-text-wrap">
+            No
+          </span>
+        </td>
+        <td>?</td>
+        <td>
+          <span class="no-text-wrap">
+            <a href="https://www.hetzner.com/rechtliches/datenschutz">Hetzner Online GmbH</a>,
+          </span>
+        </td>
+      </tr>
+
+      <tr>
         <td data-value="UncensoredDNS">
           <a href="https://blog.uncensoreddns.org/">UncensoredDNS</a>
         </td>
--- a/_includes/sections/dns.html
+++ b/_includes/sections/dns.html
@@ -150,7 +150,7 @@
             <a href="https://www.datacenterlight.ch/">Data Center Light</a>,
           </span>
           <span class="no-text-wrap">
-            <a href="https://www.hetzner.com/">Hetzner Online GmbH</a>,
+            <a href="https://www.hetzner.com/">Hetzner Online GmbH</a>
           </span>
         </td>
       </tr>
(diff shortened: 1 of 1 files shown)


_includes/sections/dns.html
```diff
@@ -463,6 +463,43 @@ <h1 id="icanndns" class="anchor"><a href="#icanndns"><i class="fas fa-link ancho
         </td>
       </tr>
 
+      <tr>
+        <td data-value="Snopyta">
+          <a href="https://snopyta.org/service/dns/index.html">Snopyta</a>
```
Mikaela: I wonder if I should call it as Snopyta DNS?

_includes/sections/dns.html
```diff
@@ -463,6 +463,43 @@ <h1 id="icanndns" class="anchor"><a href="#icanndns"><i class="fas fa-link ancho
...
+            Finland
+          </span>
+        </td>
+        <td>
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title="https://snopyta.org/privacy_policy/" href="https://snopyta.org/privacy_policy/">
+            <img alt="WWW" src="/assets/img/layout/www.png" width="35" height="35">
+          </a>
+        </td>
+        <td>
+          <a data-toggle="tooltip" data-placement="bottom" data-original-title='Part of LibreHosters, "a network of cooperation and solidarity that uses free software to encourage decentralisation through federation and distributed platforms."' href="https://libreho.st/">
```
Mikaela: I just copy-pasted this and wonder if LibreHosters should be a type?

_includes/sections/dns.html
```diff
@@ -463,6 +463,43 @@ <h1 id="icanndns" class="anchor"><a href="#icanndns"><i class="fas fa-link ancho
...
+            Informal collective
+          </a>
+        </td>
+        <td>No</td>
```
Mikaela: > There are no IP addresses nor any queries logged by default. In order to prevent abuse we monitor traffic volume which may result in temporary enabling logs to block IP addresses. These logs are deleted immediately after action has been taken and are kept for a maximum of 24 hours.

but I think similar scenarios were marked as not logging. I wonder if the title should have a warning "on normal operation" or similar (as that is in our requirements)?

_includes/sections/dns.html
```diff
@@ -463,6 +463,43 @@ <h1 id="icanndns" class="anchor"><a href="#icanndns"><i class="fas fa-link ancho
...
+        <td>DoH, DoT</td>
+        <td>Yes</td>
+        <td>Yes</td>
+        <td>
+          <span class="no-text-wrap">
+            No
+          </span>
+        </td>
+        <td>?</td>
```
Mikaela: Similar to a lot of others, I guess in theory I could link to Unbound source code or https://git.snopyta.org/snopyta/ but neither seems that helpful to me?

_includes/sections/dns.html
```diff
@@ -463,6 +463,43 @@ <h1 id="icanndns" class="anchor"><a href="#icanndns"><i class="fas fa-link ancho
...
+        <td>
+          <span class="no-text-wrap">
+            <a href="https://www.hetzner.com/rechtliches/datenschutz">Hetzner Online GmbH</a>,
```
```suggestion
            <a href="https://www.hetzner.com/">Hetzner Online GmbH</a>
```

I guess I should remove this comma and there are a lot of excess commas with a closer look. And I thought I force pushed the privacy policy away?

_includes/sections/dns.html
nitrohorse: For consistency if we do, then I think we should update these also, yeah?
- AdGuard DNS
- NixNet DNS

I think it's also fine how it is without "DNS."

_includes/sections/dns.html
nitrohorse: I'm not sure, maybe? Leaving the type as "informal collective" makes it easy to add future DNS that are under a different collective I guess.

_includes/sections/dns.html
nitrohorse: I think if you look at Quad9, it's the same situation: we state "Some" with a tooltip regarding this type of scenario.

_includes/sections/dns.html
nitrohorse: Yeah, I guess I'd also leave it as-is.

_includes/sections/dns.html
Mikaela: Ok, I guess I will let it be.

_includes/sections/dns.html
Mikaela: I need to check later, but I think Nixnet or Lelux or similar had this.

Perflyst: Everyone needs to enable logging if there is abuse and if the provider writes you an email because your server was part of a DNS reflection DDoS. If I claim I will never log than this is a lie and I also would hurt the full DNS network.

Perflyst: But ok, we can do it like this: I do never log any DNS queries.
I still can use tcpdump or similar to get the information needed to stop the abuse. Is this logging?

Mikaela:
This seems ready for merging enough

Issue #1314, opened 2019-09-15
🌐 Website Issue | The VPN page doesn't mention encrypted DNS


Mikaela:
## Description

![Screenshot from 2019-09-15 11-24-14](https://user-images.githubusercontent.com/831184/64918749-5fe91100-d7ab-11e9-8366-9d4f5abe32f6.png)

I think the warning should also include recommending encrypting DNS as otherwise the network administrator or attacker in local network can manipulate DNS queries or even anyone between you and the DNS server can do that.

I don't know if this would require more explanation including that if you use a VPN, your DNS goes through the VPN and using both simultaneously would require trusting both providers. And if the user is using Tor, they should let Tor/exit-nodes perform DNS to not taint circuits or send all traffic through one circuit (even while Tor exit-node is comparable to untrusted/unsecured open network).

* https://www.privacytools.io/providers/dns/#icanndns

blacklight447-ptio:
Im not sure about this one, the warning is already pretty long, if we make it any longer people my skip reading it at all.

Mikaela:
I think encrypted DNS is an important base for https as the untrusted network could otherwise be sending users to wrong places and trigger certificate warnings.

blacklight447-ptio:
Could you make a pull request to show how you would like it?

Mikaela:
Sure, seems simple enough and I think HTTPS could become a link alongside it. I cannot give schedule promises though:

PR #1340, opened 2019-09-23
vpn: encourage encrypted DNS & linkify https

Mikaela:
Resolves: #1314

https://deploy-preview-1340--privacytools-io.netlify.com/providers/vpn/


pages/providers/vpn.html
```diff
@@ -10,7 +10,7 @@
   <div class="card-body">
     <p class="card-text text-danger">Using a VPN will <strong>not</strong> keep your browsing habits anonymous, nor will it add additional security to non-secure (HTTP) traffic.</p>
     <p class="card-text text-danger">If you are looking for <strong>anonymity</strong>, you should use the Tor Browser <strong>instead</strong> of a VPN.</p>
-    <p class="card-text text-danger">If you're looking for added <strong>security</strong>, you should always ensure you're connecting to websites using HTTPS. A VPN is not a replacement for good security practices.</p>
+    <p class="card-text text-danger">If you're looking for added <strong>security</strong>, you should always ensure you're connecting to websites using <a href="https://en.wikipedia.org/wiki/HTTPS">HTTPS</a> and <a href="/providers/dns/#icanndns">encrypted DNS</a>. A VPN is not a replacement for good security practices.</p>
```
Mikaela: Logic flaw. Encrypted DNS comes before HTTPS.

Mikaela:
I rubber duck debugged this in the new issue page.

* * * * *

🌐 Website Issue | VPN questions / encryption doesn't mention encrypted DNS and may be a bit misleading

#1207 not fixed by #1340.

> ## What if I need encryption?
> 
> In most cases, your traffic is already encrypted! Over 98% of the top 3000 websites offer HTTPS, meaning your traffic is safe regardless of using a VPN. It is incredibly rare for applications that handle personal data to not support HTTPS in 2019, especially with services like Let's Encrypt offering free HTTPS certificates to any website operator.

Nope, your DNS traffic is announced and modifiable in plain text unless you run a local DNSSEC validating resolver (which will only help with DNSSEC-signed domains), so better advice would be to encrypt DNS.

However that doesn't prevent SNI from leaking where you are connecting to except in the rare cases of using FIrefox + DoH + Cloudflared domain.

> Even if a site you visit doesn't support HTTPS, a VPN will not protect you, because a VPN cannot magically encrypt the traffic between the VPN's servers and the website's servers. Installing an extension like HTTPS Everywhere and making sure every site you visit uses HTTPS is far more helpful than using a VPN.

So maybe there is a missing question on whether I should use encrypted DNS with a VPN?

Mikaela:
How about additional informationing self-contained networks to poke more holes into why someone would want a VPN? :smile_cat:

pages/providers/vpn.html
```diff
@@ -10,7 +10,7 @@
   <div class="card-body">
     <p class="card-text text-danger">Using a VPN will <strong>not</strong> keep your browsing habits anonymous, nor will it add additional security to non-secure (HTTP) traffic.</p>
     <p class="card-text text-danger">If you are looking for <strong>anonymity</strong>, you should use the Tor Browser <strong>instead</strong> of a VPN.</p>
-    <p class="card-text text-danger">If you're looking for added <strong>security</strong>, you should always ensure you're connecting to websites using HTTPS. A VPN is not a replacement for good security practices.</p>
+    <p class="card-text text-danger">If you're looking for added <strong>security</strong>, you should always ensure you're connecting to websites using <a href="/providers/dns/#icanndns">encrypted DNS</a> and <a href="https://en.wikipedia.org/wiki/HTTPS">HTTPS</a>. A VPN is not a replacement for good security practices.</p>
```
Mikaela: This is the main fix for #1314.

pages/providers/vpn.html
```diff
@@ -142,8 +142,11 @@ <h3>Should I use a VPN?</h3>
...
       <p>VPNs cannot encrypt data outside of the connection between your device and the VPN server. VPN providers can see and modify your traffic the same way your ISP could. And there is no way to verify a VPN provider's "no logging" policies in any way.</p>
       <h3>What if I need encryption?</h3>
-      <p>In most cases, your traffic is already encrypted! Over 98% of the top 3000 websites offer <strong>HTTPS</strong>, meaning your traffic is safe regardless of using a VPN. It is incredibly rare for applications that handle personal data to not support HTTPS in 2019, especially with services like Let's Encrypt offering free HTTPS certificates to any website operator.</p>
+      <p>In most cases, most of your traffic is already encrypted! Over 98% of the top 3000 websites offer <strong>HTTPS</strong>, meaning your non-DNS traffic is safe regardless of using a VPN. It is incredibly rare for applications that handle personal data to not support HTTPS in 2019, especially with services like Let's Encrypt offering free HTTPS certificates to any website operator.</p>
```
Mikaela: This was missing the fact that DNS was most likely unencrypted.

pages/providers/vpn.html
```diff
@@ -142,8 +142,11 @@ <h3>Should I use a VPN?</h3>
...
       <p>Even if a site you visit doesn't support HTTPS, a VPN will not protect you, because a VPN cannot magically encrypt the traffic between the VPN's servers and the website's servers. Installing an extension like <a href="https://www.eff.org/https-everywhere">HTTPS Everywhere</a> and making sure every site you visit uses HTTPS is far more helpful than using a VPN.</p>
+      <h4>Should I use <a href="/providers/dns/#icanndns">encrypted DNS</a> with a VPN?</h4>
```
Mikaela: I am not entirely happy with this, but I welcome better suggestions. Otherwise I would say that it's good enough.

pages/providers/vpn.html
```diff
@@ -142,8 +142,11 @@ <h3>Should I use a VPN?</h3>
...
+      <p>The answer to this question is also the not very helpful <strong>it depends</strong>. Your VPN provider may have their own DNS servers, but if they don't, the traffic between your VPN provider and the DNS server isn't encrypted. You need to trust the <a href="/providers/dns/#icanndns">encrypted DNS provider</a> in addition to the VPN provider and unless your client and target server support <a href="https://www.eff.org/deeplinks/2018/09/esni-privacy-protecting-upgrade-https">encrypted SNI</a>, the VPN provider can still see which domains you are visiting.</p>
+      <p>However <strong>you shouldn't use encrypted DNS with Tor</strong> as that would put all your DNS queries through a single circuit instead of separating them to different exit nodes and thus allowing the encrypted DNS provider to deanonymize you.</p>
```
Mikaela: I think it's a fair warning even if maybe misplaced.

pages/providers/vpn.html
```diff
@@ -167,6 +170,7 @@ <h3>Are VPNs ever useful?</h3>
         <ol>
           <li><a href="https://schub.io/blog/2019/04/08/very-precarious-narrative.html">VPN - a Very Precarious Narrative</a> by Dennis Schubert</li>
           <li><a href="https://gist.github.com/joepie91/5a9909939e6ce7d09e29">Don't use VPN services</a> by Sven Slootweg</li>
+          <li><a href="/software/networks/">Self-contained networks</a> on privacytools.io are able to replace a VPN that allows access to services on local area network</li>
```
Mikaela: Don't forget what VPNs were originally for. And keep Tor hidden services and similar in mind :)

pages/providers/vpn.html
```diff
@@ -142,8 +142,11 @@ <h3>Should I use a VPN?</h3>
...
       <p>Even if a site you visit doesn't support HTTPS, a VPN will not protect you, because a VPN cannot magically encrypt the traffic between the VPN's servers and the website's servers. Installing an extension like <a href="https://www.eff.org/https-everywhere">HTTPS Everywhere</a> and making sure every site you visit uses HTTPS is far more helpful than using a VPN.</p>
+      <h4>Should I use <a href="/providers/dns/#icanndns">encrypted DNS</a> with a VPN?</h4>
+      <p>The answer to this question is also the not very helpful <strong>it depends</strong>. Your VPN provider may have their own DNS servers, but if they don't, the traffic between your VPN provider and the DNS server isn't encrypted. You need to trust the <a href="/providers/dns/#icanndns">encrypted DNS provider</a> in addition to the VPN provider and unless your client and target server support <a href="https://www.eff.org/deeplinks/2018/09/esni-privacy-protecting-upgrade-https">encrypted SNI</a>, the VPN provider can still see which domains you are visiting.</p>
```
nitrohorse: suggestion to add a semicolon after "helpful":
```suggestion
      <p>The answer to this question is also the not very helpful: <strong>it depends</strong>. Your VPN provider may have their own DNS servers, but if they don't, the traffic between your VPN provider and the DNS server isn't encrypted. You need to trust the <a href="/providers/dns/#icanndns">encrypted DNS provider</a> in addition to the VPN provider and unless your client and target server support <a href="https://www.eff.org/deeplinks/2018/09/esni-privacy-protecting-upgrade-https">encrypted SNI</a>, the VPN provider can still see which domains you are visiting.</p>
```

pages/providers/vpn.html
nitrohorse: Suggestion: I think we don't need to hyperlink in this title since we already link to that page within the body of text.
```suggestion
      <h4>Should I use encrypted DNS with a VPN?</h4>
```

pages/providers/vpn.html
nitrohorse: Just a suggestion, a bit more terse:
```suggestion
      <p>However <strong>you shouldn't use encrypted DNS with Tor</strong>. This would direct all of your DNS requests through a single circuit, and would allow the encrypted DNS provider to deanonymize you.</p>
```

PR #707, opened 2018-12-31
Add Source Code

gjhklfdsa:
**Description**: Add a list of every projects source code.
**Why**?  Both to prove that a project is indeed free software and for people to find a utilize the source code and bug trackers.

I'm open to suggestions and feedback. 
I still am looking for several item's source code. If you know where those are and could tell me it would be greatly appreciated. :)

Edit: I've been working on this for a few days (offline at first), that is partially why there are so many individual commits.

PR #1348, opened 2019-09-25
Remove source_code.md

dawidpotocki:
This file seems useless to me. We already give link to the source code in
the cardv2 recommendations. A lot of repositories are wrong in the
`source_code.md`, we still list "Secure Hosting Provider" as [WIP], some
sections are outdated like "PC Operating Systems" since it lists Trisquel as
main recommendation and doesn't have Fedora, in some places we give all
recommendations listed and put "closed-source" to closed onces, but in some
we just skip them, formatting is a big mess and… that file actually is not
accessible from the privacytools.io, even though it has a Jekyll front
matter and is accessible from local build on my machine.

So yeah, nobody gives a fsck about that file and just makes maintaining site
a bit more… annoying ;p.


Mikaela:
I would like to keep it as a lot of work went towards it in https://github.com/privacytoolsIO/privacytools.io/pull/707 and it also contains worth mentioning etc.

dawidpotocki:
> I would like to keep it as a lot of work went towards it in https://github.com/privacytoolsIO/privacytools.io/pull/707 and it also contains worth mentioning etc.

That a lot of work was put into something, doesn't mean it shouldn't be
removed. If you want, you can add icon next to worth mentioning items,
but `source_code.md` is just a mess and makes contributing more
annoying.

PR #1364, opened 2019-09-29
operating-systems: add Linux in front of kernel

Mikaela:
Co-authored-by: Dawid Potocki <dpot@disroot.org>


Mikaela:
> Don't explain it because you are just wasting your time

:confused:

Issue #1373, opened 2019-10-03
✨ Website Suggestion | Decentralized Website Hosting


JonahAragon:
Two questions: Should we upload our static content to IPFS? Should we mirror this site on ZeroNet?

Mikaela:
> Should we upload our static content to IPFS? 

I like the idea and it could also lead into removing binary files from the git repo.

> Should we mirror this site on ZeroNet?

I am a bit meh about zeroNet, but it has been requested in https://github.com/privacytoolsIO/privacytools.io/issues/139#issuecomment-489437354.

PR #1374, opened 2019-10-03
Add media assets to IPFS

JonahAragon:
## Description

Testing

Related: #1373 

#### Check List 


JonahAragon:
```
<meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'self' 'unsafe-inline' https://*.privacytools.io 127.0.0.1; style-src 'self' 'unsafe-inline' 127.0.0.1; img-src 'self' data: https://*.privacytools.io 127.0.0.1; object-src 'none'; frame-src https://stats.privacytools.io; font-src 'self'; manifest-src 'self';">
```

Mikaela:
I wonder if the IPFS gateway should be a variable in `_config.yml` like [I do with mikaela.info](https://github.com/Mikaela/mikaela.github.io/blob/f75d710b6ea3f38fd4b532301e7f08f8acf86b9e/_config.yml#L35-L43) ?

JonahAragon:
I don't see why we'd ever change it though. I wouldn't want to trust a third-party gateway for non-IPFS users (especially since most traffic would be going to whatever gateway we specify).

Mikaela:
> I don't see why we'd ever change it though. I wouldn't want to trust a third-party gateway for non-IPFS users (especially since most traffic would be going to whatever gateway we specify).

What if the gateway ever gets renamed or there will be IPFS gateway redirector instead or something like that? I think it could also be beneficial for forks and maybe local testing.

Issue #1106, opened 2019-08-06
✨ Help Wanted! | Add Weblate i18n Support


JonahAragon:
## Checklist

- [x] Homepage
- [ ] Providers 
  - [ ] Avoid US Services
  - [ ] Cloud Storage
  - [ ] DNS
  - [ ] Email
  - [ ] Hosting
  - [ ] Pastebin
  - [ ] Search
  - [ ] Social Networks
  - [ ] Social News Aggregators
  - [ ] VPN
- [ ] Browser
  - [ ] Fingerprint
  - [x] Browser Addons (#1138)
  - [ ] Tweaks
  - [ ] Recommendation
  - [ ] WebRTC Section
- [ ] Software
  - [ ] Calendar/Contacts Sync
  - [ ] Notebook
  - [ ] Email Alternatives
  - [ ] Email Clients
  - [ ] File Encryption
  - [ ] File Sharing
  - [ ] Password Manager
  - [ ] Productivity
  - [ ] Real-Time Communication
    - [ ] IM
    - [ ] VoIP
    - [ ] Team Chat
  - [ ] File Sync
  - [ ] Self-Contained Notebooks
  - [ ] Self-Hosted Cloud
- [ ] OS
  - [x] Android Add-ons
  - [ ] Windows 10 (In progress: https://github.com/privacytoolsIO/privacytools.io/commit/cb2ae143ed66239ac3dc54cb4b2708b84fdc1f8e)
  - [x] Live CD (#1163)
  - [x] Mobile (#1163)
  - [x] PC (#1163)
  - [x] Router (#1163)
- [x] [Services](https://www.privacytools.io/services/) Page (#1138)
- [x] Donate
- [x] [Contact](https://www.privacytools.io/contact/)
- [x] Navbar
- [x] Footer (#1138)

## Description

Alright. So it turns out if we use [jekyll-multiple-languages-plugin](https://github.com/kurtsson/jekyll-multiple-languages-plugin) and throw *every single string on the site* into `/_i18n/en.yml` we can then use [our Weblate install](https://weblate.nablahost.com/projects/privacytoolsio/) to translate the site much easier.

The problem of course, is moving all the strings manually 😅 

So that's going to be my project for now. As of df38bc4 I have successfully moved... the index page to `en.yml`, which took about an hour 🙄 

[...]

PR #1457, opened 2019-11-03
i18n: Key donate and sponsors, fix navbar, merge master

djoate:
This is a part of https://github.com/privacytoolsIO/privacytools.io/issues/1106. Please take a look at my reviews before merging

Keys the strings of /donate and /sponsors into en.yml; fix some unkeyed strings in navbar; merge master.
PR also fixes Whonix string taking up two lines and some minor punctuation


_i18n/en.yml
```diff
@@ -246,12 +247,45 @@ participate:
...
+  tax_and_financial_info: "Tax and Financial Information"
+  contribs_handled_by_opencol: "Your contribution to privacytools.io will be handled by the Open Collective Foundation 501(c)(3). For US companies and taxpayers, this means your contribution is"
+  tax_deducible: "tax-deductible"
+  non_profit_and_anons_no_sponsorship: "As a non-profit, your sponsorship contribution will not be used for private profit and will only be used to cover expenses incurred by the project. All of our transactions (donations and expenses) are published transparently on OpenCollective. For the benefit of our readership, anonymous contributions will not be eligible for the sponsorship opportunities outlined above."
```
djoate: These three strings break up one paragraph because "tax-deductible" is surrounded with a `<strong>` tag. Should we not do this and just put `<strong>` in `en.yml` instead?

JonahAragon: Yeah... That seems like a better plan, and I'm wondering if we should do that for all the strings we've broken up too 😳

djoate: This is done in https://github.com/privacytoolsIO/privacytools.io/pull/1457/commits/a007e37a5f79f0d72dad69e5448a898424bf63ac. (Of course, for those not knowledgeable about HTML, doing this means that we should to advise translators in the future about not translating `< ... >` contents)

Issue #1341, opened 2019-09-23
🌐 Website Issue | Hosting providers have no warning (unlike VPN/DNS)


Mikaela:
The VPN and DNS providers warn that we cannot know what they are doing and are trusting their word on it. The hosting providers have no such a warning and the page only says:

> Web Hosting
> Find a web hosting provider that won't track your visitors or give into government data requests.

blacklight447-ptio:
I agree, maybe even provide a notice that even when a service makes its server software open source, it has no way of proving that it really runs THAT software.

Issue #1196, opened 2019-08-22
✨ Feature Suggestion | Hosting Guidlines


ggg27:
**Description**: Discussion thread for adding "secure hosting provider" to [Software Criteria](https://github.com/privacytoolsIO/privacytools.io/blob/master/.github/CONTRIBUTING.md#software-criteria) 

**Why?**: There are a lot of web hosting providers promoting privacy (see #1811), the goal of this issue is to set criteria. Especially for issues like #1043 and #934.

**What do you recommend?**: 
The current VPN criteria is:

- [ ] Prioritize Products by privacy respecting nationality.
- [ ]    Cannot be based in USA or UK.
- [ ]    Must be acessable via free software (i.e OpenVPN, WireGuard)
- [ ]    Use Encryption
- [ ]    Accept Cryptocurrency
- [ ]    No logging policy

I think this is a good start. Although, I would like to propose expanding "Must be acessable via free software" to include free/libre cPanels like [VESTA](https://github.com/serghey-rodin/vesta) as well as using privacy respecting OSes (like Debian or OpenBSD/LibertyBSD) and up-to-date free/libre hypervisors.

OpenSSH is standard now, so basically everybody encrypts their connection to a VPS and web-hosting is generally done using HTTPS-client.

Mikaela:
Some things I would like to see:

* eco friendliness (see https://datacenterlight.ch/ for example)
* ethical (e.g. I don't want to see PTIO recommending providers that host child porn https://github.com/privacytoolsIO/privacytools.io/issues/1299#issuecomment-536671500)
    * This is not a call to delist Tor or other "darknets", as far as I am aware of, their developers have a stance against child porn, but cannot affect what users use them for regardless of whether it's bad. And if there was a backdoor against child abusers, there would be nothing to prevent it from being used against the next whistleblower or similar.
    * I am also not supporting that the providers should mass surveill what goes on their servers (I don't know whether the hosts of CyberBunker were aware of the illicit content on their services, but I hope that when someone reports malicious content on a server to the provider, the provider will confirm this claim and then take action).

blacklight447-ptio:
We will be making a solid set of criteria after our email section is finished.

Mikaela:
Comments from Linda at Matrix:

>> Prioritize Products by privacy respecting nationality.
> 
> What does that mean?

[...]

PR #1464, opened 2019-11-04
hosting-provider: add a quick warning

Mikaela:
## Description

Resolves: #1341

* Netlify preview for the mainly edited page: https://deploy-preview-1464--privacytools-io.netlify.com/providers/hosting/

### DIFF
--- a/_includes/sections/hosting-provider.html
+++ b/_includes/sections/hosting-provider.html
@@ -1,7 +1,7 @@
 <h1 id="host" class="anchor"><a href="#host"><i class="fas fa-link anchor-icon"></i></a> Secure Hosting Provider</h1>
 
 <div class="alert alert-warning" role="alert">
-  <strong>Note: We cannot know what the hosting providers are doing and have to take their word on it.</strong>
+  <strong>Note: We cannot know what the hosting providers are doing and have to take their word on it. Please verify their privacy policies.</strong>
 </div>
 
 {% include cardv2.html
--- a/_includes/sections/hosting-provider.html
+++ b/_includes/sections/hosting-provider.html
@@ -1,5 +1,9 @@
 <h1 id="host" class="anchor"><a href="#host"><i class="fas fa-link anchor-icon"></i></a> Secure Hosting Provider</h1>
 
+<div class="alert alert-warning" role="alert">
+  <strong>Note: We cannot know what the hosting providers are doing and have to take their word on it.</strong>
+</div>
+
 {% include cardv2.html
 title="Data Center: Bahnhof"
 image="/assets/img/provider/bahnhof.png"
--- a/_includes/sections/hosting-provider.html
+++ b/_includes/sections/hosting-provider.html
@@ -1,7 +1,7 @@
 <h1 id="host" class="anchor"><a href="#host"><i class="fas fa-link anchor-icon"></i></a> Secure Hosting Provider</h1>
 
 <div class="alert alert-warning" role="alert">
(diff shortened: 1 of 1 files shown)


Mikaela:
Linda:

> https://github.com/privacytoolsIO/privacytools.io/pull/1464
> The problem at https://github.com/privacytoolsIO/privacytools.io/issues/1341 is more about "won't track your visitors or give into government data requests" being dangerous to say. The quick warning is "fine", but it's not fine because... it's self-explanatory and not really a warning at all.
> It's best to question why such warning would be there in first place, as if PTIO didn't know what it was recommending or why.
> And honestly, I *don't* have an idea why PTIO is recommending any of those but Njalla.
> I know Bahnhof, but I don't know why it's on that list.
> Companies also have to keep records for tax purposes, etc.

Mikaela:
I don't know these providers well personally, but I would argue that a warning is necessary as we aren't all-knowing.

See also https://github.com/privacytoolsIO/privacytools.io/issues/1196

_includes/sections/hosting-provider.html
```diff
@@ -1,5 +1,9 @@
 <h1 id="host" class="anchor"><a href="#host"><i class="fas fa-link anchor-icon"></i></a> Secure Hosting Provider</h1>
 
+<div class="alert alert-warning" role="alert">
+  <strong>Note: We cannot know what the hosting providers are doing and have to take their word on it.</strong>
```
nitrohorse: What about adding an additional suggestion to verify their privacy policies?
```suggestion
  <strong>Note: We cannot know what the hosting providers are doing and have to take their word on it. Please verify their privacy policies.</strong>
```

Mikaela: I committed this, but would "read" make more sense than "verify" ?

nitrohorse: I meant verifying their claims on their main home pages, but yeah “read” probably is more clear.

_includes/sections/hosting-provider.html
```diff
@@ -1,5 +1,9 @@
 <h1 id="host" class="anchor"><a href="#host"><i class="fas fa-link anchor-icon"></i></a> Secure Hosting Provider</h1>
 
+<div class="alert alert-warning" role="alert">
+  <strong>Note: We cannot know what the hosting providers are doing and have to take their word on it. Please verify their privacy policies.</strong>
```
Mikaela:
```suggestion
  <strong>Note: We cannot know what the hosting providers are doing and have to take their word on it. Please read their privacy policies.</strong>
```

Lazy suggestion, so I don't have to terminal it.

_includes/sections/hosting-provider.html
Mikaela: Changed to read, please rereview

JonahAragon:
I do not think this is necessary, or we would have to add such a warning to every page on the site. VPN and DNS providers seem like special cases because they will receive your data when you are visiting third parties. Hosting providers will only see your data when you are directly accessing their servers.

If we want to add a generic warning, adding some text for clarification of our point and putting it on https://www.privacytools.io/providers/ would be a better solution.

Mikaela:
> Hosting providers will only see your data when you are directly accessing their servers. 

I don't understand how to interpret this. Hosting provider has root on your servers and thus see my data when I am using anything not-E2EE on a server hosted on them.

Mikaela:
I am unlikely to work on this any time soon.

PR #1484, opened 2019-11-10
Add warning to Cryptomator that its mobile apps are not open-source

nitrohorse:
## Description

Apologies for this very late PR.

Resolves: https://github.com/privacytoolsIO/privacytools.io/pull/1288/files

#### Check List 

- [x] I have read and understand [the contributing guidelines](https://github.com/privacytoolsIO/privacytools.io/blob/master/.github/CONTRIBUTING.md).

- [x] The project is [Free Libre](https://en.wikipedia.org/wiki/Free_software) and/or [Open Source](https://en.wikipedia.org/wiki/Open-source_software) Software

* Netlify preview for the mainly edited page: https://deploy-preview-1484--privacytools-io.netlify.com/software/encryption-tools/#encrypt

* Code repository of the project (if applicable): N/A

PR #1494, opened 2019-11-16
Warn Cryptomator mobile apps not open source in cloud storage

djoate:
See https://github.com/privacytoolsIO/privacytools.io/pull/1484

https://deploy-preview-1494--privacytools-io.netlify.com/providers/cloud-storage/


nitrohorse:
Ah good catch! Didn't realize we reference Cryptomator on several pages.

PR #1499, opened 2019-11-17
about: don't link to Keybase for my keys

Mikaela:
I am against listing Keybase as the PGP host, because they are a centralized service and may terminate accounts at any time. Also I cannot link keys to https://mikaela.info/PGP/0xB2F32B67.txt as it excludes existence of SSH and all the other key types (e.g. minisign?).

Of course now I am giving that power to gitea.blesmrt.net, but at least that instance is not for-profit business.

Later I may move the keys onto IPFS to be more sure that they won't be tampered with.


pages/about.html
```diff
@@ -96,8 +96,7 @@ <h2>Core Contributors</h2>
 nick="Mikaela"
 bio="I am interested in instant messengers. I review privacy tools, and participate in our forum, Matrix room, and issue tracker."
```
Mikaela: I am also somewhat unhappy about this, but to be done at a better time.

PR #1516, opened 2019-11-24
Expand CONTRIBUTING.md with information to help new contributors

dawidpotocki:
Installation, building, file/directory structure, submitting patches


JonahAragon:
I agree it doesn’t make sense to me to move repo files out of `.github`.

dawidpotocki:
> I agree it doesn’t make sense to me to move repo files out of
> `.github`

And for me it makes crap ton of sense. CONTRIBUTING.md is not GitHub
specific. Just like we have README.md and LICENSE.txt at the root of
repository, even though they can be placed in `.github`. It makes it
easier to find, because it was before in a hidden directory, for which
I'm never looking for when I clone the project and it will still be
detected by GitHub. I don't see a reason for it staying there.

freddy-m:
@privacytools/editorial I removed Dawid's assignment to this, should we still continue with it ourselves?

dngray:
> @privacytools/editorial I removed Dawid's assignment to this, should we still continue with it ourselves?

I guess so, I haven't looked into it in too much detail.

PR #1539, opened 2019-11-29
Add AzireVPN

dngray:
Don't merge yet, need to make sure the logo is up.


dngray:
> As mentioned before, I am not personally super interested in listing providers that have not undergone a security audit. IVPN and ProtonMail have both committed to doing so, and releasing their results in Q1 2020. And if they don't, they will be removed here. Unless AzireVPN meets that same guideline, "mentioning" them here still doesn't make sense to me.

They have said they are saving up for an comprehensive security audit of their network. I guess we can leave this open until they do.

I do agree though, we can't be seen to be giving special treatment.

> This is not the reason I'm blocking this PR though, if everyone else wants to list them, then _whatever_. I'm blocking this because links to their website should be removed. We are intentionally not linking to providers sites if they are not recommended.

They have been removed, that was my bad.

Mikaela:
https://github.com/azirevpn/azclient/pulse/monthly - how come their official client hasn't seen development for over an year?

dngray:
> https://github.com/azirevpn/azclient/pulse/monthly - how come their official client hasn't seen development for over an year?

Well it might be there aren't any issues with it. It is just after-all a wrapper around OpenVPN.

I do know they're developing an iOS client, and i think focusing on Wireshark at the moment.

nitrohorse:
> I do know they're developing an iOS client, and i think focusing on WireGuard at the moment.

Looks like it’s [out in beta today](https://www.azirevpn.com/blog/2019-12-23/azirevpn-on-ios) but no news of open-sourcing it.

dngray:
Going to push this to a separate branch, so that in the future we can resurrect this PR if we need to.

From here it can be pushed to master.

PR #1560, opened 2019-12-02
Remove wirecutter, affiliate driven site, talks about VPNs we don't recommend anyway

dngray:
Uses affiliate links, we don't want to be driving traffic there. Recommends VPNs that do not fit our criteria.


dngray:
I think we could use some of the information in a blog article. I don't feel we should link to a site that uses a tonne of affiliate links to VPNs we do not recommend, because they do not meet our criteria. Comes back to this doesn't it https://blog.privacytools.io/the-trouble-with-vpn-and-privacy-reviews/

PR #1569, opened 2019-12-05
Recommend "Tox" over "qTox"

dngray:
https://deploy-preview-1569--privacytools-io.netlify.com/software/real-time-communication/#peer-to-peer

### DIFF
--- a/_includes/sections/instant-messenger.html
+++ b/_includes/sections/instant-messenger.html
@@ -178,17 +178,22 @@
 
 {%
   include cardv2.html
-  title="qTox"
+  title="Tox"
   image="/assets/img/svg/3rd-party/tox.svg"
   image-dark="/assets/img/svg/3rd-party/tox-dark.svg"
   description='Encrypted instant messaging and video calling software. Uses its <a href="https://toktok.ltd/spec.html"> own encryption protocol</a> that has not yet been officially audited by cryptographers.'
   labels="warning:<a href=//tox.chat/download.html#warning>Experimental</a>:Encryption has not been audited by professional cryptographers|success:VoIP"
   website="https://tox.chat"
   forum="https://forum.privacytools.io/t/discussion-tox/2115"
-  github="https://github.com/qTox/qTox"
   windows="https://tox.chat/download.html#oses"
   mac="https://tox.chat/download.html#oses"
   linux="https://tox.chat/download.html#oses"
+  freebsd="https://tox.chat/download.html#oses"
+  openbsd="http://openports.se/search.php?so=tox"
+  netbsd="http://pkgsrc.se/search.php?so=tox"
+  fdroid="https://tox.chat/download.html#oses"
+  googleplay="https://tox.chat/download.html#oses"
+  ios="https://tox.chat/download.html#oses"
 %}
 
 <h4>Worth Mentioning</h4>


dngray:
> I wonder why the change of heart, but apparently Tox website is also promoting qTox

They promote qtox, utox, antidote (ios) and antox (android) on their downloads page. This PR makes it more consistent with the rest of the site like we've done with Matrix, XMPP etc.

_includes/sections/instant-messenger.html
```diff
@@ -173,22 +173,27 @@ <h3>Disadvantages</h3>
...
   include cardv2.html
-  title="qTox"
+  title="Tox"
   image="/assets/img/svg/3rd-party/tox.svg"
   image-dark="/assets/img/svg/3rd-party/tox-dark.svg"
   description='Encrypted instant messaging and video calling software. Uses its <a href="https://toktok.ltd/spec.html"> own encryption protocol</a> that has not yet been officially audited by cryptographers.'
   labels="warning:<a href=//tox.chat/download.html#warning>Experimental</a>:Encryption has not been audited by professional cryptographers|success:VoIP"
   website="https://tox.chat"
   forum="https://forum.privacytools.io/t/discussion-tox/2115"
-  github="https://github.com/qTox/qTox"
```
JonahAragon: https://github.com/Tox

_includes/sections/instant-messenger.html
```diff
@@ -173,22 +173,27 @@ <h3>Disadvantages</h3>
...
   windows="https://tox.chat/download.html#oses"
   mac="https://tox.chat/download.html#oses"
   linux="https://tox.chat/download.html#oses"
+  freebsd="https://tox.chat/download.html#oses"
+  openbsd="http://openports.se/search.php?so=tox"
+  netbsd="http://pkgsrc.se/search.php?so=tox"
```
JonahAragon: I'd probably prefer to remove these sites personally if they are not "official", anyone can search so it's hardly helpful. https://tox.chat/download.html does not reference these platforms at all.

dngray: Fair enough, I blame @dawidpotocki for his suggestion.

_includes/sections/instant-messenger.html
dngray: It seems qtox has their own organization though https://github.com/qTox should we link to both?

JonahAragon: Uh. I would say no since we’re no longer recommending qTox specifically?

dngray: Yes, but qtox and utox are both "official" clients on https://tox.chat/download.html

_includes/sections/instant-messenger.html
dngray: I have reconsidered, we do provide information about support for when Netbsd and Openbsd has packages for other things so I don't see why not here.

_includes/sections/instant-messenger.html
dngray: That is why i removed the Github link, because we're linking to the downloads page anyway.

nitrohorse: For Matrix’s GitHub link we list Riot for Web so I’m wondering if for consistency we stay with  linking qTox here?

dawidpotocki: > For Matrix’s GitHub link we list Riot for Web

But in Matrix case, we only recommend Riot, in Tox case, we don't only recommend qTox.

_includes/sections/instant-messenger.html
```diff
@@ -178,17 +178,22 @@ <h3>Disadvantages</h3>
...
```
Mikaela: http?

dngray: As @dawidpotocki said, it's not ideal, but there isn't really much of an alternative besides linking to CVS ie:

https://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/net/toxcore/
http://cvsweb.netbsd.org/bsdweb.cgi/pkgsrc/chat/toxcore/

Then if we do that we can't really link to more than one thing:

https://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/net/utox/
http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/net/toxic

Looks like OpenBSD doesn't have qtox packaged, guess we could link to utox, toxic is a command line client.

https://wip.pkgsrc.org/cgi-bin/gitweb.cgi?p=pkgsrc-wip.git;a=tree;f=uTox-git

PR #1570, opened 2019-12-05
Remote port forwarding badge for VPN providers

dngray:
https://deploy-preview-1570--privacytools-io.netlify.com/providers/vpn/

### DIFF
--- a/_includes/sections/vpn.html
+++ b/_includes/sections/vpn.html
@@ -88,7 +88,7 @@
     <h5><span class="badge badge-success">77+ Servers</span></h5>
     <p>IVPN has 77 servers in 31 countries at the time of writing this page. Typically the more servers a provider offers, the better. IVPN has a decent (but not exceptional) server count that will most likely provide adequate coverage to most users.</p>
     <h5><span class="badge badge-success">Remote port forwarding</span></h5>
-    <p>Remote <a href="https://en.wikipedia.org/wiki/Port_forwarding">port forwarding</a> is possible with a Pro plan. Port forwarding <a href="https://www.ivpn.net/knowledgebase/81/How-do-I-activate-port-forwarding.html">can be activated</a> via the client area.</p>
+    <p>Remote <a href="https://en.wikipedia.org/wiki/Port_forwarding">port forwarding</a> is possible with a Pro plan. Port forwarding <a href="https://www.ivpn.net/knowledgebase/81/How-do-I-activate-port-forwarding.html">can be activated</a> via the client area. Port forwarding is only available on IVPN when <a href="https://www.ivpn.net/knowledgebase/116/Port-forwarding-is-not-working-why.html">using OpenVPN and is disabled on US servers</a>.</p>
     <h5><span class="badge badge-success">Accepts Bitcoin</span></h5>
     <p>In addition to accepting credit/debit cards and PayPal, IVPN accepts <strong>Bitcoin</strong> and <strong>cash/local currency</strong> (on annual plans) as anonymous forms of payment.</p>
     <h5><span class="badge badge-success">Mobile Clients</span></h5>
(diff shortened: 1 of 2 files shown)


dngray:
> Reads as advertising, don't see how it is relevant (i.e. not related to our VPN criteria). I would move to the additional functionality section below.

Certainly comes under the technology section. Particularly when choosing a VPN for P2P purposes its nice to be able to open a port to get more peers. I think that information could be useful to people.

pages/providers/vpn.html
```diff
@@ -64,6 +64,7 @@ <h3><span class="badge badge-info">Technology</span></h3>
         <li>Killswitch with highly configurable options (enable/disable on certain networks, on boot, etc.)</li>
         <li>Easy-to-use VPN clients, especially open-source. Even better if the Android version is also <a href="https://www.f-droid.org/en/2019/05/05/trust-privacy-and-free-software.html">available in F-Droid</a>.</li>
         <li>Supports <a href="https://en.wikipedia.org/wiki/IPv6">IPv6</a>. We expect that servers will allow incoming connections via IPv6 and allow users to access services hosted on IPv6 addresses.</li>
+        <li>Capability of <a href="https://en.wikipedia.org/wiki/Port_forwarding#Remote_port_forwarding">remote port fowarding</a> assists in creating connections when using P2P (<a href="https://en.wikipedia.org/wiki/Peer-to-peer">Peer-to-Peer</a>) filesharing software, Freenet or hosting a server (eg Mumble).</li>
```
djoate:
```suggestion
        <li>Capability of <a href="https://en.wikipedia.org/wiki/Port_forwarding#Remote_port_forwarding">remote port fowarding</a> assists in creating connections when using P2P (<a href="https://en.wikipedia.org/wiki/Peer-to-peer">Peer-to-Peer</a>) filesharing software, Freenet or hosting a server (e.g., Mumble).</li>
```

_includes/sections/vpn.html
```diff
@@ -85,6 +87,8 @@ <h5><span class="badge badge-danger">No Security Audit</span></h5>
...
     <h5><span class="badge badge-success">77+ Servers</span></h5>
     <p>IVPN has 77 servers in 31 countries at the time of writing this page. Typically the more servers a provider offers, the better. IVPN has a decent (but not exceptional) server count that will most likely provide adequate coverage to most users.</p>
+    <h5><span class="badge badge-success">Remote port forwarding</span></h5>
+    <p>Remote <a href="https://en.wikipedia.org/wiki/Port_forwarding">port forwarding</a> is possible with pro accounts. Port forwarding <a href="https://www.ivpn.net/knowledgebase/81/How-do-I-activate-port-forwarding.html">can be activated</a> via the client area.</p>
```
djoate: Maybe we should mention that port forwarding is only available for OpenVPN and non-US servers (https://www.ivpn.net/knowledgebase/116/Port-forwarding-is-not-working-why.html)

Also, it seems that Pro is used as a proper noun (for example, they say it as "IVPN Pro plan" at https://www.ivpn.net/knowledgebase/20/How-many-devices-or-simultaneous-connections-can-I-have-to-IVPN.html)

```suggestion
    <p>Remote <a href="https://en.wikipedia.org/wiki/Port_forwarding">port forwarding</a> is possible with a Pro plan. Port forwarding <a href="https://www.ivpn.net/knowledgebase/81/How-do-I-activate-port-forwarding.html">can be activated</a> via the client area.</p>
```

dngray: > Maybe we should mention that port forwarding is only available for OpenVPN and non-US servers (https://www.ivpn.net/knowledgebase/116/Port-forwarding-is-not-working-why.html)

The reason for that is they don't want people doing P2P on US servers, likely because of DMCA notices. I have noticed that with a few providers.

They are also doing Wireguard, so that might not be strictly true about it only being available with OpenVPN, in the future.

dngray: Looking at https://mullvad.net/nl/help/port-forwarding-and-mullvad/ it seems that port forwarding over wireguard is certainly possible, so I think that's a fair point.

PR #1575, opened 2019-12-06
Mention F-Droid, seeing as we it in the VPN section and in other places such as software recommendations.

dngray:
https://deploy-preview-1575--privacytools-io.netlify.com/operating-systems/#mobile_os


Mikaela:
What was the reason for closure?

Issue #1532, opened 2019-11-27
Firefox about:config changes in 70


Mikaela:
@jasonbrown1965 commented in https://github.com/privacytoolsIO/privacytools.io/issues/1328#issuecomment-558900201 :

> Minor update for add-ons page:
>
> Step 2 under about:config Preparations currently reads:
> 2. Press the button "I'll be careful, I promise!"
> 
> In FF 70.01, the text now reads:
> "I accept the risk!"

I guess we should have both versions listed and find out when the change arrives at ESR, so we know when to remove it.

Issue #1602, opened 2019-12-20
Update button caption of about:config in Firefox


igoose1:
## Description:

I think with an update long time ago button had been changed to "Accept the Risk and Continue" so it'd be better to update that in `_includes/sections/browser-tweaks.html` on line 9.

https://github.com/privacytoolsIO/privacytools.io/blob/34c099ecaba0c4b196b9f2c74b28f04a965ac410/_includes/sections/browser-tweaks.html#L9

## Screenshot:

There is a screenshot to show what I see when visiting about:config.
![image](https://user-images.githubusercontent.com/28492051/71249701-dce9d080-232e-11ea-8e38-d537e771a549.png)

Mikaela:
Duplicate of https://github.com/privacytoolsIO/privacytools.io/issues/1532.

PR #1603, opened 2019-12-20
Change a warning of Firefox about:config entering

igoose1:
## Description

Button had been changed to "Accept the Risk and Continue" so I've updated that. More information at #1602

Resolves: https://github.com/privacytoolsIO/privacytools.io/issues/1532

#### Check List 

### DIFF
--- a/_includes/sections/browser-tweaks.html
+++ b/_includes/sections/browser-tweaks.html
@@ -6,7 +6,7 @@
 
 <ol>
   <li>Enter "about:config" in the firefox address bar and press enter.</li>
-  <li>Press the button "Accept the Risk and Continue"</li>
+  <li>Press the button "I'll be careful, I promise!" or "Accept the Risk and Continue"</li>
   <li>Follow the instructions below...</li>
 </ol>
 
--- a/_includes/sections/browser-tweaks.html
+++ b/_includes/sections/browser-tweaks.html
@@ -6,7 +6,7 @@
 
 <ol>
   <li>Enter "about:config" in the firefox address bar and press enter.</li>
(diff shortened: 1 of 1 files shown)


_includes/sections/browser-tweaks.html
```diff
@@ -6,7 +6,7 @@ <h3>Preparation:</h3>
 
 <ol>
   <li>Enter "about:config" in the firefox address bar and press enter.</li>
-  <li>Press the button "I'll be careful, I promise!"</li>
+  <li>Press the button "Accept the Risk and Continue"</li>
```
Mikaela: Has it also changed in Firefox ESR?

igoose1: > Has it also changed in Firefox ESR?

I cannot check that on my system :woman_shrugging:. I'll try to look at someone's ESR version but I cannot promise that. Right now I can say it is, at least, on 71.0, but that's a "standard" version, not ESR.

igoose1:
I added "or" with both wordings but it must be better to check if ESR really hasn't changed the sentence.

Thorin-Oakenpants:
ESR 68.0.2, FF70 and lower (tested back to FF60)
- `I accept the risk`

FF71+
- `Accept the Risk and Continue`

I only checked on Windows, but I doubt the OS has any affect here.

---

No idea where you're getting `I'll be careful, I promise!` from but it looks vaguely familiar - maybe that's an even older wording?

Mikaela:
@igoose1 Would you be open to committing based on Thorin-Oakenpants's comment?

igoose1:
@Mikaela, here is. Thank you, @Thorin-Oakenpants!

PR #1654, opened 2020-01-16
Update Edward Snowden quotes from Permanent Record

sgraaf:
## Description

Resolves: #none 

#### Check List 

* Netlify preview for the mainly edited page: https://deploy-preview-1654--privacytools-io.netlify.com/

### DIFF
--- a/_includes/sections/quotes.html
+++ b/_includes/sections/quotes.html
@@ -14,6 +14,11 @@
    <footer class="blockquote-footer">Joshua in <cite title="The Crypto Paper"><a href="https://github.com/cryptoseb/CryptoPaper#let-me-explain-further">The Crypto Paper</a></cite></footer>
 </blockquote>
 
+<blockquote class="blockquote">
+  <p>[...] But saying that you don't need or want privacy because you have nothing to hide is to assume that no one should have, or could have, to hide anything -- including their immigration status, unemployment history, financial history, and health records. You're assuming that no one, including yourself, might object to revealing to anyone information about their religious beliefs, political affiliations, and sexual activities, as casually as some choose to reveal their movie and music tastes and reading preferences.</p>
+  <footer class="blockquote-footer">Edward Snowden in <cite title="Permanent Record"><a href="https://www.amazon.com/Permanent-Record-Edward-Snowden/dp/1250237238">Permanent Record</a></cite></footer>
+</blockquote>
+
 <h4>Read also:</h4>
 
 <ul>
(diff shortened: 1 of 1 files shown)


sgraaf:
> 
> 
> Thank you for this PR, I checked that the quotes are correct, but the only thing I am not certain of is linking to Amazon.
> 
> As I didn't find any official website for the book, what would you think about linking to Wikipedia (or would it give spoilers) or Goodreads (I am not sure if this one is justifiable) instead?
> 
>     * https://en.wikipedia.org/wiki/Permanent_Record_(autobiography)
> 
>     * https://www.goodreads.com/book/show/46223297-permanent-record
> 
> 
> If everyone else is fine with linking to Amazon, could the link point at least to smile.amazon.com so Amazon will donate some to charities [including Tor (see question 26)](https://donate.torproject.org/donor-faq)?

I was unsure about linking to Amazon myself. Linking to Goodreads seems like a perfectly fine substitute.

danarel:
> I was unsure about linking to Amazon myself. Linking to Goodreads seems like a perfectly fine substitute.

Amazon own's Goodreads.

I would recommend linking directly to the publisher and letting the user decide where to buy:

https://us.macmillan.com/books/9781250237231

Mikaela:
> Although I’m inclined to recommend just downloading a copy online due to this:

I fear that would get copyright people after us, while it would be interesting how much information there can be about that.

Unless you just said (note: Edward Snowden doesn't receive money from the book)?

dngray:
We probably can't link to a pirated copy of the book no. But we certainly shouldn't link to a shopfront of any kind given that all money is being confiscated.

nitrohorse:
@sgraaf, would you be able to update the link to Wikipedia? Then we could merge I think.

danarel:
Is there a reason why we are choosing Wikipedia over the publishers site?

nitrohorse:
I’d think Wikipedia could be a bit more informative for users due to it containing info on the civil lawsuit, specifically that the US government “aims to capture the proceeds Snowden would be earning from [the book].”

danarel:
Fair enough

sgraaf:
> 
> 
> @sgraaf, would you be able to update the link to Wikipedia? Then we could merge I think.

Yes. Will do so tomorrow.

sgraaf:
> 
> 
> @sgraaf, would you be able to update the link to Wikipedia? Then we could merge I think.

Done.

Issue #1648, opened 2020-01-10
🆕 Software Suggestion | Pass


FishCoding:
## Basic Information

**Name:** Pass
**Category:** Password Manager
**URL:** https://www.passwordstore.org/ 

## Description
It is a password manager that basically uses GPG keys as a method of encryption and decryption and contain all the password in a git repository (password storage).

adalricus:
I'm surprised pass was not listed here. @Mikaela I could help with research on this. It's a simple password manager which uses existing tools, each entry is a gpg encrypted file store in pass root directory ($HOME/.passwordstore is default).

Example tree:
```
Password Store
├── service1.org
│   ├── accounts
│   └── email
├── service2.com
│   └── www
├── Bank
│   └── bank1
```
Other users having access to pass root directory (which they shouldn't have, it should be rwx------) can read services names.

FishCoding:
It also has several plugins to for example be less permissive with the folder hierarchy encrypting the folder tree and more. Other plugins allow generating 2FA codes. 

In the overall, it's not more than a bash script for encrypting and decrypting the passwords with the gpg pair keys.

Mikaela:
Would any of you mind making a PR?

@privacytoolsIO/editorial and would any of you mind researching this? I imagine it's encryption should be OK if it really does use GPG for that.

dngray:
I actually use pass already, for mbsync, msmtp, vdirsyncer It's really quite good.

PR #1676, opened 2020-01-30
Added Pass to the PasswordManagers page

FishCoding:
## Description

Resolves: #1648  

#### Check List 

* Netlify preview for the mainly edited page: https://deploy-preview-1676--privacytools-io.netlify.com/software/passwords/
    * I have edited the Password Managers page adding the information about the program named in the Issue #1648. Also I have add a svg that is pretty accurate for the program because It does not have one.


_includes/sections/password-managers.html
```diff
@@ -67,6 +67,23 @@ <h1 id="pw" class="anchor"><a href="#pw"><i class="fas fa-link anchor-icon"></i>
   googleplay="https://play.google.com/store/apps/details?id=com.lesspass.android"
 %}
 
+{%
+  include cardv2.html
+  title="Pass - Self-Host"
+  image="/assets/img/svg/3rd-party/pass.svg"
+  description="<strong>Pass</strong> is a very simple password store that keeps passwords using gpg2 encrypted files inside a simple directory tree residing at ~/.password-store. It has a simple terminal interface where the user can perform the usual actions, and some plugins can be installed for extra functionalities."
+  website="https://www.passwordstore.org/"
+  github="hhttps://git.zx2c4.com/password-store"
```
Mikaela:
```suggestion
  git="https://git.zx2c4.com/password-store"
```

I think we have a separate photo for non-GitHub repos when `git` is used. Also a typo.

_includes/sections/password-managers.html
```diff
@@ -67,6 +67,23 @@ <h1 id="pw" class="anchor"><a href="#pw"><i class="fas fa-link anchor-icon"></i>
   googleplay="https://play.google.com/store/apps/details?id=com.lesspass.android"
 %}
 
+{%
+  include cardv2.html
+  title="Pass - Self-Host"
+  image="/assets/img/svg/3rd-party/pass.svg"
+  description="<strong>Pass</strong> is a very simple password store that keeps passwords using gpg2 encrypted files inside a simple directory tree residing at ~/.password-store. It has a simple terminal interface where the user can perform the usual actions, and some plugins can be installed for extra functionalities."
```
Mikaela:
```suggestion
  description="<strong>Pass</strong> is a very simple password store that keeps passwords using gpg2 encrypted files inside a simple directory tree residing at <code>~/.password-store</code>. It has a simple terminal interface where the user can perform the usual actions, and some plugins can be installed for extra functionalities."
```

I am not entirely sure on this description though, are Windows users likely to be familiar with `~` ?

_includes/sections/password-managers.html
```diff
@@ -67,6 +67,23 @@ <h1 id="pw" class="anchor"><a href="#pw"><i class="fas fa-link anchor-icon"></i>
...
+  website="https://www.passwordstore.org/"
+  github="hhttps://git.zx2c4.com/password-store"
+  windows="https://www.passwordstore.org/"
+  mac="https://www.passwordstore.org/"
+  linux="https://www.passwordstore.org/"
+  firefox="https://addons.mozilla.org/en-US/firefox/addon/passff/"
+  chrome="https://chrome.google.com/webstore/detail/browserpass/naepdomgkenhinolocfifgehidddafch"
+  fdroid="https://f-droid.org/packages/com.zeapo.pwdstore"
+  googleplay="https://play.google.com/store/apps/details?id=dev.msfjarvis.aps"
+  ios="https://mssun.github.io/passforios/"
```
Mikaela: I feel a bit uncomfortable with these as they aren't official apps, while they seem to be open source.

assets/img/svg/3rd-party/pass.svg
```diff
@@ -0,0 +1,46 @@
...
+</g>
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
+</svg>
```
Mikaela: Do SVGs usually look like this?

FishCoding: It can be seen in the preview, but yee it's quite strange

FishCoding:
For the review itself, the program is no more than a bash script that encrypts the password and decrypts it. 

I currently use the Android App that is in the froid store, but concerning the other applications none can ensure that are working well or are trustworthy so that I can delete the links. 

And least I agree with you to move it to the worth mentioning instead in the general section due to the "easy to use" point, I will add the the phrase you have said.

_includes/sections/password-managers.html
dngray: I agree, i think we could add pass as a worth mentioning app but not with it's own card.

The main power of `pass` is that it can store secrets which can be used in scripts without having the actual password in the script. This isn't a feature however a lot of users are going to use, and I would certainly say the 3 mentioned Bitwarden, KeePassXC, and LessPass are better choices for most users.

FishCoding: Moved to the worth mentioning section :)

_includes/sections/password-managers.html
```diff
@@ -81,4 +81,7 @@ <h3>Worth Mentioning</h3>
   <li>
     <a href="https://pwsafe.org/">Password Safe</a> - Whether the answer is one or hundreds, Password Safe allows you to safely and easily create a secured and encrypted username/password list. With Password Safe all you have to do is create and remember a single "Master Password" of your choice in order to unlock and access your entire username/password list.
   </li>
+  <li>
+    <a href="https://www.passwordstore.org/"> Pass</a> - Pass is a very simple password store that keeps passwords using gpg2 encrypted files inside a simple directory tree residing at ~/.password-store. It has a simple terminal interface where the user can perform the usual actions, and some plugins can be installed for extra functionalities. It can be used in scripts without having to put the actual password in plain text.
```
Mikaela:
```suggestion
    <a href="https://www.passwordstore.org/">Pass</a> - Pass is a bare-bones password store that keeps passwords using gpg2 encrypted files inside a simple directory tree residing at <code>~/.password-store</code>. It has a simple terminal interface where the user can perform the usual actions, and it's functionality can be extended by plugins. It can also be used in scripts without having to input the actual password in plain text.
```

Mikaela: * The other suggestions don't have a space
* @cradamy nearby suggested bare-bones as very simple means easy to use (which happens to also be one of the reasons we requested moving it as worth mentioning)
* code tags were previously suggested and approved
* the rest is just my gut feeling on what would seem better

PR #1677, opened 2020-01-31
EARN IT ACT, a workaround for banning E2EE

dngray:
I guess AG Barr is back from Christmas holidays!

Latest news on the war on end to end encryption.

### DIFF
--- a/pages/software/real-time-communication.html
+++ b/pages/software/real-time-communication.html
@@ -14,6 +14,11 @@ description: "Discover secure and private ways to communicate with others online
     Recent news about breaking E2EE on centralized instant messengers
 </h3>
 
+<h5>January 2020</h5>
+<ul>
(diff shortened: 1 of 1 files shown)


dngray:
The TLDR of it is that, companies like Facebook, Twitter etc have "safe harbor" meaning if someone uploads something illegal they are not held responsible for the possession of the content on their servers. They obviously *do* remove such content and report to law enforcement.

In this case they're talking about CSAM, (child pornography). The issue being here is they want to set up a "commission" that basically makes "best practices" in dealing with the content. The thing is if the company doesn't follow the "best practices" they lose their safe harbor protection.

The commission only requires 10 of the 15 members to agree, and none of them are civil liberties representatives. Language is extremely weak, for example they "shall consider" in "interests in privacy, data security, and product quality", which basically means nothing, they can consider it for 5 seconds and then go "nope".

They also say that the officers must "certify" regardless of if they know what they are saying is a lie. In the end all the "recommendations" are signed off by the AG unilaterally, which is currently William Barr.

So it will end up being very much law enforcement saying, ban all E2EE because child pornographers could use it, so it's a best practice that nobody has it. That document also says it won't really help with that because there is absolutely nothing stopping such criminals from either using self contained networks, (Tor, I2P, Freenet etc) or encrypting the files before they upload them.

Issue #1233, opened 2019-08-28
🆕 Software Suggestion | LuFi (Let's upload that file)


Mikaela:
## Basic Information

**Name:** LuFi
**Category:** File sharing
**URL:** https://framagit.org/fiat-tux/hat-softwares/lufi

## Description

A simple open source filesharing service with e2ee working directly in the browser without the need of installing anything.

I closed https://github.com/privacytoolsIO/privacytools.io/issues/1227 in favour of it, so I guess we should talk about listing it.

Demo instance: https://demo.lufi.io/

Production instances:

* https://framadrop.org/
* https://upload.disroot.org/

Mikaela:
I am missing discussion about this with #1684 / #1686 .

lrq3000:
@Mikaela Thanks for the ping, I forgot about that tool, it's a nice alternative to Firefox Send because there is apparently no limit of size (but then files > 500MB are kept only 24h). However, they also [log IP addresses, and they do not specify how long](https://framasoft.org/fr/cgu/).

lrq3000:
The relevant part (in french):

> * Tout comme d’autres services en ligne, Framasoft enregistre automatiquement certaines informations concernant votre utilisation du service telles que l’activité du compte (exemple : espace de stockage occupé, nombre d’entrées, mesures prises), les données affichées ou cliquées (exemple : liens, éléments de l’interface utilisateur), et d’autres informations pour vous identifier **(exemple : type de navigateur, adresse IP, date et heure de l’accès, URL de référence)**.
> * Nous utilisons ces informations en interne pour vous améliorer l’interface utilisateur des services Framasoft et maintenir une expérience utilisateur cohérente et fiable.
> * Ces données ne sont ni vendues, ni transmises à des tiers.

lrq3000:
I have tested and can confirm it's possible to upload arbitrary sized files (there is probably a technical limit somewhere but it's not specified and I could upload/download 2GB without registration). I think it's worth mentioning on PTIO.

Mikaela:
I am not sure, but I think IP address storage depends on instance and the file size is defined by an instance also, I have a faint idea of once having gone through exaple config file.

lrq3000:
Yes of course but that's why in my PR i made the entry about FramaDrop, the
official instance of LuFi (same devs, git repo hosted on framasoft), just
like PRIO already links to firefox sync instance and not just the code
repository to self host, i guess for user convenience (because else what
[...]

PR #1686, opened 2020-02-03
Updates for File Sharing

### LINKED ISSUE #1684: ✨ Feature Suggestion | Upgrade OnionShare as top 1 file sharing option, demote Firefox Send to 2nd
## Description

I suggest Firefox Send to be demoted from 1st to 2nd place in the File Sharing category, for the following reasons:

1. Firefox Send retains IP addresses in logs [for 90 days and could be used to identify an account](https://send.firefox.com/legal). They are clear about this, so it's not a malpractice in any way, it's just a limitation of their system that needs to be accounted for.
2. Files are stored (although encrypted) on a server. Mozilla is trustable apriori, but it's just less private than direct transfer between 2 parties.
3. Filesize limited to 1 GB if (pseudo-)anonymous.

Compared to OnionShare, the only advantage of Firefox Send is that it does not require to install a software. But given the downsides, and the fact that OnionShare v2 now has a nice GUI that greatly eases file sharing (I would argue on par with Firefox Send interface: only one party needs to setup the software, the other can just use Tor Browser and access the given URL, maybe even access with a non-Tor browser through a Tor proxy but I did not test yet), it seems to me that OnionShare should be proposed as the top 1 suggestion on PTIO.

By comparison, Magic Wormhole vs Firefox Send would for me be a tie, because although Magic Wormhole allows for any file size and it removes the third-party server (and it works very well even behind corporate firewalls in my tests), the fact it requires to install Python and then launch a commandline on both the sender and receiver (in other words the low usability for the lambda user) counterbalances its advantages compared to Firefox Send great UX.

lrq3000:
## Description

Resolves: #1684 
Resolves: #1233
Resolves: #1692

* ~~Upgrade OnionShare as top 1 file sharing option~~
* ~~Demote Firefox Send to 2nd~~
* Add warning on Firefox Send about logging
* Add "or send" in OnionShare description to signal that there is a receiver mode too~
* Add FramaDrop/LuFi
* Extend description of Croc (requires no dependencies, similar to Magic-Wormhole [per author](https://schollz.com/software/sending-a-file/)):

> Essentially it is the same as wormhole but it has no dependencies, you can just download it and run.

* Change "Jabber server" to "XMPP server" in FreedomBox description.

#### Check List 

* Netlify preview for the mainly edited page: https://deploy-preview-1686--privacytools-io.netlify.com/software/file-sharing/

* Code repository of the project (if applicable):


lrq3000:
Following the various voiced opinions, I removed the promotion of OnionShare as top tool. (I still think it would be better though).

I have instead refocused this PR to update the descriptions, warnings about logging, and add FramaDrop/LuFi.

_includes/sections/file-sharing.html
```diff
@@ -16,7 +17,7 @@ <h1 id="filesharing" class="anchor"><a href="#filesharing"><i class="fas fa-link
 image="/assets/img/svg/3rd-party/onionshare.svg"
 website="https://onionshare.org/"
 tor="http://lldan5gahapx5k7iafb3s4ikijc4ni7gx5iywdflkba5y2ezyg6sjgyd.onion/"
-description="OnionShare is an open-source tool that lets you securely and anonymously share a file of any size. It works by starting a web server, making it accessible as a Tor onion service, and generating an unguessable URL for you to share so that the recipients can access and download the files."
+description="OnionShare is an open-source tool that lets you securely and anonymously share a file of any size. It works by starting a web server, making it accessible as a Tor onion service, and generating an unguessable URL for you to share so that the recipients can access and download or send the files."
```
Mikaela: Something looks funny with the language here, but I cannot point what.

_includes/sections/file-sharing.html
```diff
@@ -44,6 +45,7 @@ <h1 id="filesharing" class="anchor"><a href="#filesharing"><i class="fas fa-link
 <h3>Worth Mentioning</h3>
 
 <ul>
-  <li><a href="https://github.com/schollz/croc">croc</a> - Easily and securely send things from one computer to another.</li>
+  <li><a href="https://framadrop.org/">FramaDrop</a> - Stores a file of any size for 24h. Data is end-to-end encrypted from your browser, powered by <a href="https://framagit.org/fiat-tux/hat-softwares/lufi">LuFi</a>. <span class="badge badge-warning" data-toggle="tooltip" title="FramaDrop logs IP addresses and fingerprints the browser for an unclear amount of time."><a href="https://github.com/privacytoolsIO/privacytools.io/issues/1233#issuecomment-582211312"><i class="fas fa-exclamation-triangle"></i></a></span></li>
```
Mikaela: I agree putting it to worth mentioning, but I think there is a bit of confusion on what is a sofware and what a service as Firefox Send seems to be both (or can it be self-hosted?),

Also I think it would be better to link to some upstream source rather than back to us whenever possible.

I wonder if there would be some list of LuFI instances that could be used instead.

_includes/sections/file-sharing.html
lrq3000: Ok I tried to simplify by changing to the following:

"OnionShare is an open-source tool that lets you securely and anonymously share a file of any size. It works by starting a web server accessible as a Tor onion service, with an unguessable URL that you can share with the recipients to download or send files."

_includes/sections/file-sharing.html
lrq3000: Firefox Send and Lufi/Framadrop work the same, they both are instances and self-hostable opensource softwares.

I am not sure there are any other instances of both than the official ones that are linked in this PR. Lufi has a [demo website](https://demo.lufi.io/) but it is much more limited (100MB max filesize) and explicitly not meant for real usage, whereas Framadrop is the real instance, just like send.firefox.com is the real instance of Firefox Send.

Mikaela: There is at least https://upload.disroot.org/ which is why I know of Lufi and I think I have seen at least one other one, but I cannot find it.

nitrohorse:
Thanks for the PR!

lrq3000:
Ok found an [english version](https://framasoft.org/en/cgu/) of their legal statement, I have now replaced the PTIO link by this one, thanks for reviewing!

_includes/sections/file-sharing.html
lrq3000: Still more limited (1GB max) than framasoft, but yes it would be nice to have a list of instances, but I did not find one?

Issue #1727, opened 2020-02-20
Add a warning that Jami is partially centralized


lrq3000:
## Description

Although it is described as a P2P client, Jami uses 5 different central servers to provide some functionalities, as described [here](https://git.jami.net/savoirfairelinux/ring-project/issues/765) and [here](https://jami.net/why-is-jami-truly-distributed/). Hence, without self-hosting, these servers are points of failure where meta-data can leak (see also #1357).

This by the way goes counter to the [cited advantage of P2P clients](https://www.privacytools.io/software/real-time-communication/#peer-to-peer): "There are no servers that could potentially intercept and decrypt your transmissions, unlike centralized and federated models."

I suggest a warning is added about the possibility to disable/self-host these servers.

blacklight447-ptio:
I think small note is would be a proper solution to this. this issue is one of the grey areas which are always difficult to define.

lrq3000:
@blacklight447-ptio yes I'll do it :-)

PR #1752, opened 2020-03-02
add partial centralization warning for Jami

lrq3000:
#### Description

Resolves: #1727 

Changes:
* Add label warning that Jami is partially centralized
* Extend description to mention all communications are end-to-end encrypted, including through TURN servers if used.

#### Check List 

* Netlify preview for the mainly edited page: https://deploy-preview-1752--privacytools-io.netlify.com/software/real-time-communication/#peer-to-peer

### DIFF
--- a/_includes/sections/instant-messenger.html
+++ b/_includes/sections/instant-messenger.html
@@ -164,7 +164,7 @@
   title="Jami"
   image="/assets/img/svg/3rd-party/jami.svg"
   description="Encrypted instant messaging and video calling software. Uses <a href=//jami.net/improving-performance-and-security-with-tls-1-3/>TLS 1.3</a> for encryption."
-  labels="success:VoIP"
+  labels="warning:<a href=//git.jami.net/savoirfairelinux/ring-project/issues/765>Warning</a>:This software is partially centralized but can be self-hosted.|success:VoIP"
   website="https://jami.net/"
   forum="https://forum.privacytools.io/t/discussion-jami/2116"
   gitlab="https://git.jami.net/savoirfairelinux"
--- a/_includes/sections/instant-messenger.html
+++ b/_includes/sections/instant-messenger.html
@@ -163,7 +163,7 @@
   include cardv2.html
   title="Jami"
   image="/assets/img/svg/3rd-party/jami.svg"
-  description="Encrypted instant messaging and video calling software. All communications are fully end-to-end encrypted using <a href=//jami.net/improving-performance-and-security-with-tls-1-3/>TLS 1.3</a> and never stored elsewhere than on user's devices, even when <a href=//jami.net/why-is-jami-truly-distributed/>TURN servers are used</a>."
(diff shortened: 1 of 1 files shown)


lrq3000:
Sorry I didn't know the syntax for links in a label, now it's fixed :-)

blacklight447-ptio:
The preview looks great so far, ill leave a review tommorow!

lrq3000:
@Mikaela do you mean something like [this](https://jami.net/why-is-jami-truly-distributed/)? This is linked by the first reply in the [issue](https://git.jami.net/savoirfairelinux/ring-project/issues/765) I linked in the warning, so from the issue, the user can easily go there. I chose the issue because it's a lot less verbose and it specifically addresses how to workaround this issue (ie, how to self-host), whereas the blog post while being maybe cleaner only describe the issue without providing any solution.

lrq3000:
Also I just noticed they say in the blog post that:

> Furthermore, the data that users choose to share is always transmitted through entirely peer-to-peer connections, except when a TURN server is necessary. Even then, it is fully end-to-end encrypted and never stored elsewhere than on the end-users’ devices.

Maybe we should add that in the description (that it's always end-to-end encrypted)?

blacklight447-ptio:
> 
> 
> Also I just noticed they say in the blog post that:
> 
> > Furthermore, the data that users choose to share is always transmitted through entirely peer-to-peer connections, except when a TURN server is necessary. Even then, it is fully end-to-end encrypted and never stored elsewhere than on the end-users’ devices.
> 
> Maybe we should add that in the description (that it's always end-to-end encrypted)?

Seems like a good idea to me.

Mikaela:
Would it be possible to say in the warning what exactly is not decentralized?

lrq3000:
@Mikaela Here is the list:

* push notifications
* the OpenDHT proxy
* bootstrap
* name server
* TURN

If you think it's not too long, I can add them.

Mikaela:
Oh, I see, yes, that is too long. What would be the equivalent decentralized list?

lrq3000:
Oh, tough question, I think that's still a topic of research. I think Session (Loki Messenger) implements workarounds for most of these:

* push notifications -> no real way around having a central server if phones aren't rooted, but what Session does for example is that the server is only used to regularly ping the phone and it's the phone which connects to the network to check if there are any new messages, the server has no clue (info from one of their blog post IIRC).
* the OpenDHT proxy -> federated servers (Riot) or onion-like nodes (future Session) or onion-like proxy nodes (current Session)
* bootstrap -> no clue, but I guess fixing the previous may also fix the need for a bootstrap.
* name server -> [Alpenhorn](https://github.com/vuvuzela/alpenhorn) (Vuvuzela bootstrapping and name resolving system). Also probably interesting to look at what some blockchains are doing to associate names with a public key.
* TURN -> no real alternative to my knowledge, or have a list of federated TURN servers, that's all. Or don't use them (it's optional in Jami BTW).

Issue #1522, opened 2019-11-25
🌐 Website Issue | Snowflake browser browser addon does not work with WebRTC set to 'fasle'


Bosch-0:
The Tor project Snowflake browser add-on does not work when you disable WebRTC in Firefox. I thought this would be worth adding as a note in the Web Browser -> Recommended Browser Add-ons -> Snowflake section of the site.

hugoncosta:
Want to revive this issue. Snowflake isn't necessarily a tool that would help with an individual's privacy, it can even be construed as a redflag, since it's basically like creating a bridge on the person's computer for others to benefit. And as mentioned, it does that using WebRTC, a technology we tell people to disable. Thus I'd propose we remove it.

blacklight447-ptio:
We actually aimed to split snowflake into its own little optional addon category, explaining that you will help people in oppresive countries connect to the internet anonymously, while explaining the personal privacy tradeoff.

Issue #1483, opened 2019-11-10
Add steps to disable WebRTC in Safari


vol24pl:
Please update the webRTC section about the Safari support. Current version of Safari just started working with webRTC.

EDIT: Sorry for the bad labeling, but you don’t have a good label for outdated info on the site, and this seemed like a close enough label 😅

nitrohorse:
Thanks for bringing this up @vol24pl :+1: For team reference, looks like these are the steps to disable WebRTC on Safari: https://www.privateproxyguide.com/how-to-disable-webrtc/#Safari

rex4539:
This is how the latest Safari version Experimental Features debug menu looks like.

<img width="653" alt="Screen Shot 2019-12-01 at 13 24 30" src="https://user-images.githubusercontent.com/227442/69913238-043f3300-143e-11ea-88c5-ce17a0d6545a.png">

Issue #298, opened 2017-07-23
Suggestion: WebRTC 


Jasper-Ben:
Since the Website is also directed to users who lack in depth knowledge about the technical details I would suggest adding some more context to the WebRTC section. Otherwise those people following the guide might disable it and wonder why e.g. hangout, jitsi or similar isn't working.

Also followup question: The chrome section mentions that there only is a plugin that is easily circumvented. Is this only the case with this specific plugin or is the same problem present in other (firefox) plugins? So far I've been using https://github.com/ChrisAntaki/disable-webrtc-firefox for webRTC control.

C-O-M-P-A-R-T-M-E-N-T-A-L-I-Z-A-T-I-O-N:
IIRC (tested one year ago) uBlock has that same problem in Firefox.

Shifterovich:
Can't WebRTC be disabled completely in about:config?

Jasper-Ben:
@Shifterovich It can. But from time to time I want to use WebRTC (e.g. for my selfhosted jitsi instance) without going through the hassle of enabling it in the about:config. The plugin mentioned above gives the possibility to enable/disable it with one click. So I was wondering if this is a widespread problem affecting multiple plugins.

Shifterovich:
Ah. Well, we should definitely do some research as the claim about the Chrome extension may not be true anymore.

Atavic:
Firefox and its forks can block WebTRC into [about:config](https://privacytoolsio.github.io/privacytools.io/#webrtc) or with addons.

Google Chrome has WebRTC built into the application. Alternatives:

Ungoogled Chromium modifies Chromium to eliminate these APIs, see:
https://github.com/Eloston/ungoogled-chromium/issues/179

[Unofficial Google Chromium](http://chromium.woolyss.com/) builds with No Sync • No WebRTC • No Widevine

Datenschutz:
uBlock origin also has a setting that disables the IP leak created by WebRTC if that's what you're looking for

CHEF-KOCH:
Chrome will get a feature to block insecure channels or the original stun to avoid any leaks even if you use a build with webrtc or in case you use official builds.

![webrtc](https://user-images.githubusercontent.com/727123/38247181-93ae4cd4-3744-11e8-950e-83d184caffe4.png)

So the options are as following:
* Don't use A Browser with WebRTC 
* Use a Browser with an option to disable WebRTC like Opera (Chromium tweaked Browser)
[...]

PR #1773, opened 2020-03-07
Solve issues related to WebRTC

zenithium:
## Description

Closes: #1483 
Closes: #298 
Closes: #1522

* Netlify preview for the mainly edited page: https://deploy-preview-1773--privacytools-io.netlify.com/browsers/#webrtc

### DIFF
--- a/_includes/sections/browser-webrtc.html
+++ b/_includes/sections/browser-webrtc.html
@@ -55,4 +55,4 @@
 
 <h3>What about other browsers?</h3>
 
-<p>Chrome on iOS and Internet Explorer do not implement WebRTC yet. <a href="#browser"><i class="fas fa-link"></i> But we recommend using Firefox on all devices.</a></p>
+<p>Chrome on macOS and Internet Explorer do not implement WebRTC yet. <a href="#browser"><i class="fas fa-link"></i> But we recommend using Firefox on all devices.</a></p>
--- a/_includes/sections/browser-webrtc.html
+++ b/_includes/sections/browser-webrtc.html
@@ -4,7 +4,7 @@
   <strong>WebRTC is a new communication protocol that relies on JavaScript that can leak your actual IP address from behind your VPN.</strong>
 </div>
 
-<p>While software like NoScript prevents this, it's probably a good idea to block this protocol directly as well, just to be safe.</p>
+<p>While software like NoScript prevents this, it's probably a good idea to block this protocol directly as well, just to be safe. <b>Note:</b> This disables browser-based call functionality that is used for webapps like Discord, Hangouts, Jitsi, etc.</p>
 
 <p>
 <a class="btn btn-warning" target="_blank" rel="noopener noreferrer" href="https://ipleak.net">Test your Browser now</a>
--- a/_includes/sections/browser-webrtc.html
+++ b/_includes/sections/browser-webrtc.html
@@ -55,4 +55,4 @@
 
 <h3>What about other browsers?</h3>
 
-<p>Chrome on iOS and Internet Explorer does not implement WebRTC yet. <a href="#browser"><i class="fas fa-link"></i> But we recommend using Firefox on all devices.</a></p>
+<p>Chrome on iOS and Internet Explorer do not implement WebRTC yet. <a href="#browser"><i class="fas fa-link"></i> But we recommend using Firefox on all devices.</a></p>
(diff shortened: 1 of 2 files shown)


zenithium:
Now that I think about it, having two buttons for testing again doesn't look good. Perhaps they should both be removed... The test is at the top and in step 5 users should just go to that button so having additional buttons might be unnecessary. Thoughts?

zenithium:
There seems to be conflicting information for the interface for doing this. @JonahAragon can you check that these steps are correct? Don't know anyone else who's an Apple user lol

_includes/sections/browser-webrtc.html
```diff
@@ -38,10 +38,24 @@ <h3>How to disable WebRTC in Firefox?</h3>
...
-<h3>How to fix the WebRTC Leak in Google Chrome?</h3>
+<h3>How to disable WebRTC in Safari?</h3>
+
+<ol class="long-string-list">
+  <li>Open the browser and click on the Safari tab in the menu tab.</li>
+  <li>On the Advanced tab, check the box at the bottom that says "Show the Develop menu in menu bar".</li>
+  <li>Come back to the main menu and go to the Develop tab.</li>
+  <li>In the drop-down menu, go to Experimental Features and check "Remove Legacy WebRTC API"</li>
+  <li>Done. Do the WebRTC leak test again.</li>
+</ol>
```
JonahAragon: The menu option you've mentioned here (Remove Legacy WebRTC API) does not exist in Safari. There is a menu option "_Enable_ Legacy WebRTC API" in the WebRTC menu:

image

...but it appears to be disabled by default in Catalina. I've run some tests and I can't see any evidence that WebRTC implemented in Safari leaks your IP address. I would change your text to:

```suggestion
<h3>How to disable WebRTC in Safari?</h3>

<p>Safari is far stricter with WebRTC than other major browsers, and it does not leak your IP address in its default configuration. If you'd like, you can follow these steps just to double-check your browser:</p>

<ol class="long-string-list">
  <li>Choose "Preferences" under the Safari menu in your menu bar.</li>
  <li>Select the Advanced tab and check the "Show Develop menu in menu bar" box.</li>
  <li>Exit Preferences and open the Develop menu in your menu bar.</li>
  <li>In the drop-down menu, open the "WebRTC" submenu and ensure "Enable Legacy WebRTC API" is unchecked. If it's grayed out, even better.</li>
</ol>
```

Also, I'm not sure if this PR *actually* closes #298?

zenithium: Thanks for the assist, they don't call you iJonah for nothing :) The problem was that most sources for this WebRTC implementation were from 2018-2019, and posts from 2020 were just reposting the old stuff -_-

_includes/sections/browser-webrtc.html
```diff
@@ -38,10 +38,21 @@ <h3>How to disable WebRTC in Firefox?</h3>
...
+
+<h3>How to disable WebRTC in Google Chrome?</h3>
 
 <p>WebRTC cannot be fully disabled in Chrome; however, it is possible to change its routing settings (and prevent leaks) using an extension. Two open-source solutions include <a href="https://chrome.google.com/webstore/detail/webrtc-leak-prevent/eiadekoaikejlgdbkbdfeijglgfdalml">WebRTC Leak Prevent</a> (options may need to be changed depending on the scenario), and <a href="https://chrome.google.com/webstore/detail/ublock-origin/cjpalhdlnbpafiamejdnhcphjbkeiagm">uBlock Origin</a> (select "Prevent WebRTC from leaking local IP addresses" in Settings).</p>
 
 <h3>What about other browsers?</h3>
 
-<p>Chrome on iOS, Internet Explorer and Safari does not implement WebRTC yet. <a href="#browser"><i class="fas fa-link"></i> But we recommend using Firefox on all devices.</a></p>
+<p>Chrome on iOS and Internet Explorer does not implement WebRTC yet. <a href="#browser"><i class="fas fa-link"></i> But we recommend using Firefox on all devices.</a></p>
```
JonahAragon:
```suggestion
<p>Chrome on iOS and Internet Explorer do not implement WebRTC yet. <a href="#browser"><i class="fas fa-link"></i> But we recommend using Firefox on all devices.</a></p>
```

> Chrome on iOS

[citation needed]? This may be the same as Safari now (in that WebRTC is supported, because I believe it is supported on Safari on iOS) -- Alternatively *if* Chrome on iOS doesn't implement WebRTC, we should check if Firefox on iOS and Safari on iOS does implement it in the first place, because they would be worth adding to this list if not. I'll double-check later...

zenithium: IIRC I saw the same info somewhere...

PR #1778, opened 2020-03-11
Update nitrohorse's key

nitrohorse:
## Description

Updates my public key.

> gpg --no-armor --export-options export-minimal --export [30CE697C77678A9A6B0A5D5C6F3175557E766CBF](https://keybase.io/513/pgp_keys.asc?fingerprint=30ce697c77678a9a6b0a5d5c6f3175557e766cbf) > [privacytools.io/.well-known/openpgpkey/hu/zt71pstaerw1jpa5ru3borqmfsnbwnf6](https://github.com/privacytoolsIO/privacytools.io/blob/master/.github/CODEOWNERS#L31)


nitrohorse:
Some teammates (including you 🙂) signed my key.

Mikaela:
`--export-minimal` removes all non-self-signatures and when importing keys from wkd, they are again stripped out as are the uids that weren't requested. Thus my comment

> In case the signatures are desirable, I think we should have a different repository for that, talk about how to handle it or wait for [keys.openpgp.org to hopefully support publishing cross-signatures](https://gitlab.com/hagrid-keyserver/hagrid/issues/89#note_268528260).

in #1718.

nitrohorse:
Ah thanks for clarifying that Mikaela! Blacklight already merged but it sounds like this could have been closed then. TIL

Mikaela:
Something did change though, I just have no idea what that diff ` BIN +1.86 KB (180%)` was. Did you add uids or change expiry date or add subkeys or do anything like that?

PR #1781, opened 2020-03-12
Correct mistake from VPN page

dngray:
https://deploy-preview-1781--privacytools-io.netlify.com/providers/email/#criteria


dngray:
> Only problem is that your PR title talks about VPN page while the diff is on email page

Yeah, because it was copied from the VPN page when the new email section was done. That's why it says "internet data" in the context of VPNs. We used the same text (and forgot to change to email).

PR #1804, opened 2020-03-27
privacy-resources: delist The Great Cloudwall

Mikaela:
* https://deploy-preview-1804--privacytools-io.netlify.com/#resources

### DIFF
--- a/_includes/sections/privacy-resources.html
+++ b/_includes/sections/privacy-resources.html
@@ -22,7 +22,6 @@
   <li><a href="https://www.grc.com/securitynow.htm"><strong>Security Now!</strong></a> - Weekly Internet Security Podcast by Steve Gibson and Leo Laporte.</li>
   <li><a href="https://www.jupiterbroadcasting.com/show/techsnap/"><strong>TechSNAP</strong></a> - Weekly Systems, Network, and Administration Podcast. Every week TechSNAP covers the stories that impact those of us in the tech industry.</li>
   <li><a href="https://tosdr.org/"><strong>Terms of Service; Didn't Read</strong></a> - "I have read and agree to the Terms" is the biggest lie on the web. We aim to fix that.</li>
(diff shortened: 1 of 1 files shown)


dngray:
I think in order to approve this, I would have to see some of the fluff meme stuff that does not look professional at all removed.

While it does raise some good points, I think in general it should be more succinct. Maybe we can replace it entirely with a blog post or something.

dngray:
After looking at this further I think we should remove it. It diverges into tirades that don't seem relevant. The meme rubbish sounds unprofessional and it does reflect on us by linking to it.

> And their DNS service, [1.1.1.1](https://1.1.1.1/), is also filtering out users from visiting the website by returning [fake IP](https://trac.torproject.org/projects/tor/ticket/32915) address owned by Cloudflare, localhost IP such as “127.0.0.x”, or just return nothing. Cloudflare DNS also [break](https://nitter.net/bowranger/status/1213031783576428550) [online](https://nitter.net/jb510/status/1212521533907668992) [software](https://nitter.net/No_Style/status/1201525422795710466) [from](https://nitter.net/daemuth/status/1187758306535903233) [smartphone](https://nitter.net/gregortorrence/status/1183102089439805441) [app to computer game because of their fake DNS answer](https://git.nixnet.xyz/you/stop_cloudflare/src/branch/master/PEOPLE.md).

Also that I don't believe is factually correct. There can be other reasons why this happens, Eg. https://news.ycombinator.com/item?id=19828702

dngray:
This was also somewhat related https://community.cloudflare.com/t/what-data-does-cloudflare-actually-see/28660

Comes down to threat model. It is good enough to be secure from outside adversaries... but no don't rely on CloudFlare to help you be the next Edward Snowden.

For that you want true E2EE. Double Ratchet, OTR, OpenPGP etc.

Issue #1835, opened 2020-04-17
🆕❌ Removal/cleanup 2.0 Real-Time Communication


dngray:
This issue is documenting some discussion in Matrix. The community may have noticed over the past year or so we've been really trying to improve the quality of our recommendations.

This is because we cannot in good consequence recommend experimental software.

This issue documents a number of changes which I plan to put in a PR:

1. It is finally time to de-list Tox. We've been monitoring this issue https://github.com/TokTok/c-toxcore/issues/426 and others and it is just hasn't progressed sufficiently enough.

One of the things we've observed with all our recommendations, is that documentation of the protocol is necessary for a robust design. When things are scary and undocumented developers aren't interested in joining a project. This hurts the momentum of the project and people lose interest.

2. Removal of Bitmessage

While commit logs seem to show some regular activity by the main project developer, the site indicates a stable release has not occurred since 2016. Also looks as if it depends on python2.

3. Elevating status.im and corrections.

It appears now that status.im does now no longer use the Whisper protocol, but rather a fork, called Waku, which is apparently more efficient. https://github.com/status-im/nim-eth/issues/123 there was also this article https://decrypt.co/25629/status-wants-to-be-the-worlds-most-private-instant-messaging-app it also looks as if they're going to publish in f-droid which would be good https://github.com/status-im/status-react/issues/8512

A previous issue the idea was not of high importance as it was considered too alpha-quality: https://github.com/status-im/status-react/issues/1125#issuecomment-298214111

It also looks as if it now no longer depends on Firebase https://github.com/status-im/status-react/issues/8742

They also seem they are interested in deterministic building https://github.com/status-im/status-react/issues/5587

Would require some testing.

5a384507-18ce-417c-bb55-d4dfcc8883fe:
I know there's already another issue regarding Session, but since this one is about re-structuring the entire IM section I think this can be discussed here, too.

So far the conclusion around it on #1678 was that it wasn't mature enough on stability terms and that its "onion" network may not be robust enough, from comments of the developer made on that same issue, and from testing it heavily for a month now, I can say that it functions properly in both aspects, even more than other recommendations that are listed (Jami can't even send/receive text messages properly, for instance).

[...]

PR #1836, opened 2020-04-18
Cleanup 2.0 instant messenger page

dngray:
Closes: https://github.com/privacytoolsIO/privacytools.io/issues/1835

Some interesting articles about it (just for future reference, not linked on the site).
- https://decrypt.co/25629/status-wants-to-be-the-worlds-most-private-instant-messaging-app
- https://decrypt.co/24767/blockchain-privacy-app-waku-whisper

https://deploy-preview-1836--privacytools-io.netlify.app/software/real-time-communication/


_includes/sections/instant-messenger.html
```diff
@@ -111,40 +120,44 @@ <h3>Disadvantages</h3>
...
-        </ul>
-        <li><a href="https://www.kontalk.org">Kontalk</a> is a community-driven instant messaging network based on XMPP.</li>
+      <li><a href="https://gajim.org/">Gajim</a></li>
+      <li><a href="https://conversations.im">Conversations</a></li>
+      <li><a href="https://siskin.im/">Siskin</a></li>
+      <li>Other <a href="https://omemo.top">OMEMO</a> capable clients for XMPP.</li>
     </ul>
+    <li><a href="https://www.kontalk.org">Kontalk</a> is a community-driven instant messaging network based on XMPP.</li>
+    <li><a href="https://status.im">Status.im</a> - Encrypted instant messenger with an integrated <a href="https://en.wikipedia.org/wiki/Ethereum">Ethereum</a> wallet (cryptocurrency) that also includes support for <a href="https://our.status.im/tag/dapps">DApps (decentralized apps)</a> (web apps in a curated store). Uses the <a href="https://our.status.im/status-launches-private-peer-to-peer-messaging-protocol/">Waku protocol (a fork of Whisper)</a> for P2P communication. Only available for iOS and Android.</li>
+  </ul>
```
JonahAragon, line 136: I wonder if this should be alphabetized.

dngray: - Users advanced enough to want to use another matrix client can find it themselves.
- Kontalk, a fork of XMPP, yay more fragmentation. Does anyone actually use this?
- XMPP, thinking of removing this when https://github.com/vector-im/riot-web/issues/6779 finally closes. We should in future not recommend clients/protocols that do not have E2EE on by default for private chat.

dngray: Taking another look at Kontalk, it looks like it requires phone numbers, and additionally uses [openpgp for group chat](https://github.com/kontalk/specs/blob/master/group.md#encryption) so that would indicate no PFS.

dngray: Removing XMPP recommendations, as all future clients must support E2EE by default. This is something we've discussed in the past thoroughly.

While Matrix does not *at this moment*, https://github.com/vector-im/riot-web/issues/6779#issuecomment-614822531 is imminent, so we make an exception for that.

_includes/sections/instant-messenger.html
```diff
@@ -111,40 +120,35 @@ <h3>Disadvantages</h3>
 %}
 
 <div>
-    <h4>Worth Mentioning</h4>
+  <h4>Worth Mentioning</h4>
 
-    <ul>
-        <li>Other <a href="https://matrix.org/clients">Matrix</a> clients, that may however be less feature complete than Riot.im.</li>
-        <li><a href="https://xmpp.org/about">XMPP</a> (Extensible Messaging and Presence Protocol) is an open-source communications protocol that began development in 1999. Since then, XMPP has been extended by the publishing of XEPs (XMPP Extension Protocols). <a href="https://conversations.im/omemo/">OMEMO</a> is the most popular XEP (XMPP extension) for E2EE. Clients are developed by the community and not by the XSF (XMPP Standards Foundation). <span class="badge badge-warning" data-toggle="tooltip" title="VoIP and file transfers/names may not be end-to-end encrypted.">Inconsistent E2EE</span></li>
```
Mikaela, line 118: I think delisting XMPP from federated protocols is a disservice.

_includes/sections/instant-messenger.html
```diff
@@ -111,40 +120,35 @@ <h3>Disadvantages</h3>
...
-        <ul>
-            <li><a href="https://gajim.org/">Gajim</a></li>
-            <li><a href="https://conversations.im">Conversations</a></li>
```
Mikaela, line 121: Has OMEMO enabled by default, but I am not willing to clear data on my setup to confirm.

_includes/sections/instant-messenger.html
```diff
@@ -176,30 +180,8 @@ <h3>Disadvantages</h3>
...
-    <li><a href="https://retroshare.cc">Retroshare</a> - Encrypted instant messaging and voice/video call client. RetroShare supports both <a href="https://www.torproject.org/">Tor</a> and <a href="https://geti2p.net">I2P</a>.</li>
-    <li><a href="https://bitmessage.org">Bitmessage</a> is a decentralized, encrypted, peer-to-peer, trustless communications protocol that can be used by one person to send encrypted messages to another person, or to multiple subscribers.</li>
+  <li><a href="https://retroshare.cc">Retroshare</a> - Encrypted instant messaging and voice/video call client. RetroShare supports both <a href="https://www.torproject.org/">Tor</a> and <a href="https://geti2p.net">I2P</a>.</li>
```
Mikaela, line 186: Is anyone using it?

dngray: The reason I left retroshare was because it appears to have continuous development. To be honest it looks more like a collaboration platform. It could very well be removed.

_includes/sections/instant-messenger.html
dngray: The issue is one client that does this isn't really enough.

We want to in the future make a criteria for the instant messenger page that all recommendations must have E2EE on by default for private chat. Allowing XMPP to remain means we can never do that.

_includes/sections/instant-messenger.html
dngray: It does hurt to de-list things we once recommended, really we owe it to our readers to make the best and most succinct choices. I think we must ensure that we keep in mind with our recommendations:

- security of the product, eg auditing, documentation of crypto etc
- user experience, (nobody is going to use air-gapped openpgp and pasting text files into pastebins, even though that would be "most secure")
- consistency, typically a criteria helps with this.

_includes/sections/instant-messenger.html
dngray: So I have checked, Retroshare is still listed https://github.com/privacytoolsIO/privacytools.io/blob/master/_includes/sections/self-contained-networks.html#L65 as it is appropriately a self contained network of its own, instant-messaging seems like it always a secondary functionality.

_includes/sections/instant-messenger.html
muppeth: Wait, so you will delist xmp (a protocol) because not all clients have it enabled by default, but leave matrix because atm pretty much only one client has e2ee and all the rest does not even have a support for it not to mention having it by default?

_includes/sections/instant-messenger.html
muppeth: @dngray But XMPP  is a protocol and you are talking about client feature. If you are listing only clients, then of course xmpp should not be there as its a protocol, but otherwise, why not. To my knowledge there is more xmpp clients supporting OMEMO then for example matrix clients with e2ee support.

dngray: > XMPP is a protocol and you are talking about client feature.

I am referring to the ecosystem of specifications. The issue is that we don't just recommend a protocol (that would be like recommending http, and not a web browser).

> To my knowledge there is more xmpp clients supporting OMEMO then for example matrix clients with e2ee support.

This may be the case, the issue however is that not all streams of transmission are actually E2EE (eg voice/video) and we're looking at implementing a rule (we have been for a while) that all instant messenger recommendations *must* do E2EE by default for private communications. We wanted to avoid the footgun of "oops that particular action was not E2EE, sorry you didn't know about that" to our readers.

Incidentally this PR only recommends Riot, currently that is the most fully featured. Fortunately for any other client (which lets face it is an advanced user at this point) can make use of pantalaimon. (We don't mention that though).

_includes/sections/instant-messenger.html
dngray: > Wait, so you will delist xmpp (a protocol) because not all clients have it enabled by default, but leave matrix because atm pretty much only one client has e2ee and all the rest does not even have a support for it not to mention having it by default?

We are referring to Riot specifically because it will very shortly have E2EE on by default.

> At the same time keeping and promoting matrix with all it's metadata stored indefinatelly in the database?

Individual [XMPP servers also store metadata](https://infosec-handbook.eu/blog/xmpp-aitm/) (or can). High security environments where that is an issue will operate non-federating Matrix and XMPP servers.

_includes/sections/instant-messenger.html
muppeth: Does Riot do e2ee by default. Does it do e2ee for voip too?

_includes/sections/instant-messenger.html
muppeth: > Individual XMPP servers also store metadata (or can).

Though on matrix they just DO regardless of whether you want it or not. And while others do it in logs, matrix does it in database. 

>We are referring to Riot specifically because it will very shortly have E2EE on by default.

Conversations already does have it on by default so not sure whats the logic behind it.

_includes/sections/instant-messenger.html
dngray: > Does Riot do e2ee by default. Does it do e2ee for voip too?

Yes, it does.

Not for group meetings but nobody has that at the moment. We can't have a requirement for something that doesn't exist. Jitsi is working on that though https://jitsi.org/blog/e2ee/ (group meetings in Riot use Jitsi).

_includes/sections/instant-messenger.html
dngray: > Though on matrix they just DO regardless of whether you want it or not. And while others do it in logs, matrix does it in database.

The logic is if something can, then we assume it does. Particularly in a federated network. Better to assume that it does than pretend like it might not.

Unless you have a non-federating server there's really no way to know what remote servers do.

> Conversations already does have it on by default so not sure whats the logic behind it.

Yes it does, but the issue is a lack of other high quality clients like it for other platforms.

Future discussion about XMPP should be in our issue https://github.com/privacytoolsIO/privacytools.io/issues/1838

5a384507-18ce-417c-bb55-d4dfcc8883fe:
No Session, then?

dngray:
> No Session, then?

Out of scope for this issue see https://github.com/privacytoolsIO/privacytools.io/issues/1678

PR #1924, opened 2020-05-17
Update other old badges

dngray:


dngray:
Last spot is the legal document, but that needs some work in `badge.html` I think.

dngray:
We'll fix the legal page in a separate PR.

PR #1926, opened 2020-05-18
Updated Email Criteria According to the Website for The GitHub Contributing Section

FreddyMarsden:
## Description

Resolves: #none 

#### Check List


dngray:
I think we should remove the requirement out of there entirely, so we don't have to make sure the two files are in sync.

FreddyMarsden:
@dngray good idea.

PR #1936, opened 2020-05-27
Use cdn.privacytools.io for static resources

jonaharagon:
You'd have to build locally with `JEKYLL_ENV=production bundle exec jekyll serve` or `JEKYLL_ENV=production bundle exec jekyll build` to see the changes yourself, since the CDN URL should not be used in a development environment like Netlify


nitrohorse:
Nice, could we also update the Readme build instructions to include this PR’s description? Helpful for future reference.

blacklight447-ptio:
Why would we need a CDN though?

jonaharagon:
> could we also update the Readme build instructions to include this PR’s description?

We *could*, although it doesn't make sense to load from the CDN in development. Because if you add a new image or whatever to your local fork, the CDN won't display it obviously, it will only load assets on the webserver. So really you'd only set `JEKYLL_ENV=production` when building the live website, or if you just wanted to see the CDN in action for some reason (like you might with this specific PR). But otherwise I'm not sure it's particularly helpful information.

nitrohorse:
Ah that makes sense, okay, I was mistaken; we can leave the Readme as-is.

Issue #1707, opened 2020-02-13
🌐 Website Issue | New email clients page


dngray:
After https://github.com/privacytoolsIO/privacytools.io/pull/1672 is complete I want to improve the email clients page.

I think for a start we should split the recommendations by platform. A criteria should exist that only allows email clients that support OpenPGP.

Closes: https://github.com/privacytoolsIO/privacytools.io/issues/1643
Closes: https://github.com/privacytools/privacytools.io/issues/2072

blacklight447-ptio:
So it should be similar to our browser section?

dngray:
I was more thinking of a tidyup, eg removing Torbirdy as it's no longer supported etc.

Splitting the "Privacy Email Tools" so testing tools are under "Testing tools". Maybe we could do our own OpenPGP tutorial, ie https://github.com/privacytoolsIO/privacytools.io/issues/1725 first.

blacklight447-ptio:
We could do both, one section with clients for all platforms like the browser section, and below that we list email testing tools and maybe link to a blog post with a openpgp tutorial.

5a384507-18ce-417c-bb55-d4dfcc8883fe:
Enigmail should be delisted too as Thunderbird will add its own support for OpenPGP and it's going to be discontinued this year.

https://blog.thunderbird.net/2019/10/thunderbird-enigmail-and-openpgp/

>The Enigmail Add-on has made it possible to use Thunderbird with external GnuPG software for OpenPGP messaging. Because the types of add-ons supported in Thunderbird will change with version 78, the current Thunderbird 68.x branch (maintained until Fall 2020) will be the last that can be used with Enigmail.

Mikaela:
I would wait for the Thunderbird version with it's own OpenPGP implementation to be released and at least the biggest distributions to remove Enigmail from their repositories in favour of it first.

FreddyMarsden:
[Mailpile](https://www.mailpile.is/) is currently under worth mentioning, though I feel as if it should be given its own card. Heres what the new layout could look like:

### Desktop Email Clients:

- Thunderbird
- Mailpile
- Claws-Mail

### Mobile Email Clients:

- K-9 Mail
- FariMail

Other clients to think about could be stuff like [notmuchmail](https://notmuchmail.org/) or [aerc](https://aerc-mail.org/) though they don't necessarily focus on privacy.

dngray:
[...]

Issue #1643, opened 2020-01-09
🆕 Software Suggestion | Evolution Mail


cvandesande:
## Basic Information

Name: Evolution Mail
Category: Email Software
URL: https://wiki.gnome.org/Apps/Evolution

## Description
One of the oldest and still regularly developed email clients for Linux. Feels like this should have been a recommend app for a long time.

dngray:
The only issue with adding this, is we really then should also add [KMail](https://userbase.kde.org/KMail) as well.

Unfortunately these things aren't really portable to Windows, MacOS etc.

blacklight447-ptio:
well we currently don't really require cross platform do we?

PR #1990, opened 2020-07-14
That email clients cleanup

freddy-m:
## Description

Resolves: #1707 
Resolves: #1643
Resolves: #2112
Resolves: https://github.com/privacytools/privacytools.io/issues/2072
Resolves: https://github.com/privacytools/privacytools.io/issues/2136
Resolves: https://github.com/privacytools/privacytools.io/pull/2137
Resolves: https://github.com/privacytools/privacytools.io/issues/2141
Resolves: https://github.com/privacytools/privacytools.io/issues/1248
Resolves: https://github.com/privacytools/privacytools.io/issues/2133 - lets close this for now, as they're working on it.

#### Check List 

* Netlify preview for the mainly edited page: https://deploy-preview-1990--privacytools-io.netlify.app/software/email/


dngray:
I'm thinking we might have a sub category for Mobile clients.. Android and iOS, because on mobile platform things like this are never multi platform.

We should also look into some iOS clients that support PGP. Realistically we should be only recommending clients that can do PGP. Tutanota being the only exception, but that's not really a client, it's a service (so I wouldn't be recommending their app on this page).

dngray:
I'm thinking Letterbox is probably a bit too experimental at the moment to be added. I'd probably add https://canarymail.io though, and yes I'm aware it's closed source, but you have a lot less options on iOS anyway.

FreddyMarsden:
I think swapping letterbox for Canary Mail works, and maybe letterbox could be put under worth mentioning (unless we are trying to get rid of that)?

dngray:
> I think swapping letterbox for Canary Mail works, and maybe letterbox could be put under worth mentioning (unless we are trying to get rid of that)?

This might be a good idea, we may make an exception here to the worth mentioning rule as it's something we want to keep an eye on.

FreddyMarsden:
Do we want to create a "Testing tools" section as mentioned in the issue? Or should that be something for the wiki/blog?

dngray:
> Do we want to create a "Testing tools" section as mentioned in the issue? Or should that be something for the wiki/blog?

We could have a heading maybe instead of "worth mentioning" something like "Future items to be considered:" or something to that effect.

FreddyMarsden:
Hmmm, but should we try and distinguish between email clients that could be consdered, like tools such as [Email Privacy Tester](https://www.emailprivacytester.com/) or some of the other things currently under Worth Mentioning?

dngray:
We don't have to mention Enigmail anymore https://blog.thunderbird.net/2020/07/whats-new-in-thunderbird-78/

koenvervloesem:
> We don't have to mention Enigmail anymore

I would still mention it, because OpenPGP support is not there yet: "Thunderbird 78.2, due out in the coming months, will offer a new feature that allows you to end-to-end encrypt your email messages via OpenPGP."

dngray:
> I would still mention it, because OpenPGP support is not there yet: "Thunderbird 78.2, due out in the coming months, will offer a new feature that allows you to end-to-end encrypt your email messages via OpenPGP."

ah this is true, and you need that for card reader support.

FreddyMarsden:
[Mailvelope](https://www.mailvelope.com/en) doesn't really deserve to be delisted I don't think. Could be but under "worth mentioning" or the proposed "Future items to be considered:" area.

dngray:
> I don't think [Mailvelope](https://www.mailvelope.com/en) deserves to be delisted. Could be but under "worth mentioning" or the proposed "Future items to be considered:" area perhaps?

I'd suggest it under the desktop section actually. It underwent an audit a while back https://www.mailvelope.com/en/blog/mailvelope-4.0

https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/Studies/Mailvelope_Extensions/Mailvelope_Extensions_pdf.pdf?__blob=publicationFile&v=3

FreddyMarsden:
Something about having Mailvelope amongst the other email clients seems odd, perhaps we could create a sepeate section for extensions? Seeing as it isn't an email client, it feels weird that it is in with email clients, or maybe thats just me.

dngray:
I'm also not opposed to adding https://neomutt.org down the bottom under worth mentioning maybe.

It is a multi platform, and I'm not against TUI clients like https://aerc-mail.org or https://meli.delivery in the future, if they gain some stability.

freddy-m:
I'd be ok with that. notmuchmail could also be worth considering.

freddy-m:
We should probably add a criteria, most likely just a copypasta of the email one but adapted for clients.

dngray:
I think this is looking pretty good now.

freddy-m:
@dngray same, however it would be good to create a criteria surely?

dngray:
> @dngray same, however it would be good to create a criteria surely?

The issue I have with a criteria is, we test these products. Generally I don't like recommending closed source products, but we sometimes make an exception ie with CanaryMail for example.

I think the main thing we would require is that the client have support for PGP encryption, I think that's really all we can put there.

freddy-m:
@dngray perhaps we should include this somewhere on the page, otherwise I think we're pretty much done here!

LongJohn-Silver:
> @dngray perhaps we should include this somewhere on the page, otherwise I think we're pretty much done here!
> 
> Also, I think we could put Enigmail under worth mentioning for desktop?

I'm pretty sure that Enigmail is going to reach EOL somewhere this year since Thunderbird is adding its built in PGP support.

dngray:
> I'm pretty sure that Enigmail is going to reach EOL somewhere this year since Thunderbird is adding its built in PGP support.

Redirected from [Will OpenPGP cards be supported for private key storage ?](https://wiki.mozilla.org/Thunderbird:OpenPGP:2020#Will_OpenPGP_cards_be_supported_for_private_key_storage_.3F):

[How to use Thunderbird 78 with smartcards](https://wiki.mozilla.org/Thunderbird:OpenPGP:Smartcards)

> Thunderbird 78 does NOT use GnuPG by default. Instead, it uses the RNP library. The RNP library does NOT yet support the use of smartcards for secret keys.

dngray:
> We should verify this works first.

I've verified this with 78.5 and this works quite nicely. There is a wizard that helps with conversion.

These releases included OpenPGP fixes, so it's being made 1st class:


- https://www.thunderbird.net/en-US/thunderbird/78.1.1/releasenotes/
- https://www.thunderbird.net/en-US/thunderbird/78.2.0/releasenotes/
- https://www.thunderbird.net/en-US/thunderbird/78.2.1/releasenotes/
- https://www.thunderbird.net/en-US/thunderbird/78.2.2/releasenotes/
- https://www.thunderbird.net/en-US/thunderbird/78.3.0/releasenotes/
- https://www.thunderbird.net/en-US/thunderbird/78.3.2/releasenotes/
- https://www.thunderbird.net/en-US/thunderbird/78.3.3/releasenotes/
- https://www.thunderbird.net/en-US/thunderbird/78.5.0/releasenotes/

dngray:
Okay, so it looks like you've removed the repo that this PR merges from.

We will need another PR later to, so I'll unlink these issues from this PR.

- Criteria, which really at this point can only contain OpenPGP support. Ideally I'd like to make sure things are cross-platform and open source, but that narrows the page down significantly further than what we've got. CanaryMail is closed source and MacOS only, but on Apple/iOS devices it is likely to provide the most native experience. Therefore I'm thinking we won't add a criteria as such, as it is a general guide.
- We should evaluate https://github.com/privacytools/privacytools.io/issues/1248 as a possible alternative to k-9 mail, I've not used this.
- https://github.com/privacytools/privacytools.io/issues/2120
- https://github.com/privacytools/privacytools.io/issues/2133

dngray (approved):

dngray:
For some reason this didn't end up merging with master, even though the status says it did, https://github.com/privacytools/privacytools.io/pull/2142 should fix that.

PR #1995, opened 2020-07-17
Change PC Operating Systems to Introductory Operating Systems

dngray:
https://deploy-preview-1995--privacytools-io.netlify.app/operating-systems/#os

### DIFF
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -86,7 +86,7 @@
             </span>
           </summary>
           <span class="nav-dropdown">
-            <a class="dropdown-item" href="/operating-systems/#os"><span class="fas fa-th-large fa-fw"></span> PC OS</a>
+            <a class="dropdown-item" href="/operating-systems"><span class="fas fa-th-large fa-fw"></span> PC OS</a>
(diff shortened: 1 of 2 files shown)


jonaharagon:
We should change the navbar menu and add an entry there for the advanced section as well so it matches the page.

Mikaela:
What does Introductory mean in this context?

dngray:
> What does Introductory mean in this context?

A place where you should start.

dngray:
> We should change the navbar menu and add an entry there for the advanced section as well so it matches the page.

I decided we don't really need to link to introductory/advanced as they're both "PC OS", so just link to the main page.

If we did add those two separate subcategories to the nav.html, we'd need to think of suitable icons for each.

_includes/nav.html
```diff
@@ -86,7 +86,7 @@
             </span>
           </summary>
           <span class="nav-dropdown">
-            <a class="dropdown-item" href="/operating-systems/#os"><span class="fas fa-th-large fa-fw"></span> PC OS</a>
+            <a class="dropdown-item" href="/operating-systems"><span class="fas fa-th-large fa-fw"></span> PC OS</a>
```
jonaharagon, line 89: What is the reason for this change?

PR #2004, opened 2020-07-27
Owned by domain squatters

dngray:


Mikaela:
I think the commit and GitHub PR topic should have said what link exactly was owned by domain squatters without needing to open commit details for seeing it.

PR #2007, opened 2020-08-01
Replace Alt-Svc with Onion-Location header

dawidpotocki:
Onion-Location is a preferred way by Tor Project to indicate an
available onion service. It allows users to make Tor Browser always use
onion version or to keep using clearnet in a seamless way.

https://community.torproject.org/onion-services/advanced/onion-location/

---

Example sites:
* https://torproject.org
* https://dawidpotocki.com

### DIFF
--- a/nginx/010-headers.conf
+++ b/nginx/010-headers.conf
@@ -4,7 +4,7 @@ add_header X-Content-Type-Options nosniff always;
 add_header Content-Security-Policy "default-src 'none'; script-src 'self' https://stats.privacytools.io; style-src 'self'; img-src 'self' data: https://*.privacytools.io; object-src 'none'; frame-src https://stats.privacytools.io; font-src 'self'; base-uri 'none'; form-action 'self' https://search.privacytools.io; frame-ancestors 'none'; manifest-src 'self';" always;
 add_header 'Access-Control-Allow-Origin' '*';
 add_header Strict-Transport-Security "max-age=31557600; includeSubDomains; preload";
-add_header Alt-Svc 'h2="privacy2zbidut4m4jyj3ksdqidzkw3uoip2vhvhbvwxbqux5xy5obyd.onion:443"; ma=86400; persist=1';
+add_header Onion-Location http://privacy2zbidut4m4jyj3ksdqidzkw3uoip2vhvhbvwxbqux5xy5obyd.onion$request_uri always;
 add_header Expect-CT 'max-age=86400, enforce';
 add_header Referrer-Policy "strict-origin";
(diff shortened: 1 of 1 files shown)


nginx/010-headers.conf
```diff
@@ -4,7 +4,7 @@ add_header X-Content-Type-Options nosniff always;
 add_header Content-Security-Policy "default-src 'none'; script-src 'self' https://stats.privacytools.io; style-src 'self'; img-src 'self' data: https://*.privacytools.io; object-src 'none'; frame-src https://stats.privacytools.io; font-src 'self'; base-uri 'none'; form-action 'self' https://search.privacytools.io; frame-ancestors 'none'; manifest-src 'self';" always;
 add_header 'Access-Control-Allow-Origin' '*';
 add_header Strict-Transport-Security "max-age=31557600; includeSubDomains; preload";
-add_header Alt-Svc 'h2="privacy2zbidut4m4jyj3ksdqidzkw3uoip2vhvhbvwxbqux5xy5obyd.onion:443"; ma=86400; persist=1';
+add_header Onion-Location http://privacy2zbidut4m4jyj3ksdqidzkw3uoip2vhvhbvwxbqux5xy5obyd.onion$request_uri always;
```
Mikaela, line 7: https://community.torproject.org/onion-services/advanced/onion-location/#nginx doesn't seem to have the ` always` part?

dawidpotocki: > https://community.torproject.org/onion-services/advanced/onion-location/#nginx doesn't seem to have the ` always` part?

Yes you are right, it doesn't have `always`.
WONTFIX 🙃

jonaharagon:
Are we sure Onion-Location replaces alt-svc entirely? It seems like they serve different purposes to me, so it probably makes more sense to serve both.

Mikaela:
Is Alt-Svc documented anywhere to show it's supported at all? Did it have a spec?

Personally I did leave it present in the only place I used it

jonaharagon:
> Did it have a spec?

It isn't a _Tor_ specification, `alt-svc` is an established header that has been used for many different purposes in the past. (Silent HTTPS upgrades, SPDY support, and now this). Its specification was [RFC 7838](https://tools.ietf.org/html/rfc7838).

Support for it with .onion addresses was added to Tor Browser in Sept 2018.

Mikaela:
:+1: I was thinking of a Tor specification like the [100-onion-location-header.txt](https://gitweb.torproject.org/tor-browser-spec.git/tree/proposals/100-onion-location-header.txt) via the original link, which to be honest appears to be a proposal rather than actual accepted spec, but I guess it's wide adoption means it's becoming one soon.

dawidpotocki:
> Are we sure Onion-Location replaces alt-svc entirely? It seems like they serve different purposes to me, so it probably makes more sense to serve both.

Well, Alt-Svc kind of removes the option to keep using clearnet… no?

PR #2011, opened 2020-08-04
Hated one Video, Tor vs VPN

dngray:


jonaharagon (changes requested):
bad thumbnail, bad layout

PR #2058, opened 2020-09-21
Mention discontinuation uMatrix, mention uBlock advanced mode

dngray:
## Description

Resolves: https://github.com/privacytools/privacytools.io/issues/2057

#### Check List 

* https://deploy-preview-2058--privacytools-io.netlify.app/browsers/#addons

### DIFF
--- a/_includes/sections/browser-addons.html
+++ b/_includes/sections/browser-addons.html
@@ -171,7 +171,7 @@
 {% include cardv2.html
   title="uMatrix: Stop Cross-Site Requests"
   image="/assets/img/png/3rd-party/umatrix.png"
-  description="<strong>uMatrix</strong> gives you control over the requests that websites make to other websites. Many websites integrate features which let other websites track you, such as Facebook Like Buttons or Google Analytics. uMatrix allows 1st party scripts in its default configuration. If you want the default functionality of NoScript consider <a href=https://github.com/gorhill/uMatrix/wiki/How-to-block-1st-party-scripts-everywhere-by-default>blocking 1st party scripts everywhere by default</a>. This addon has been <strong><a href=https://github.com/uBlockOrigin/uMatrix-issues/issues/291#issuecomment-694988696>discontinued</a></strong>"
+  description="<strong>uMatrix</strong> gives you control over the requests that websites make to other websites. Many websites integrate features which let other websites track you, such as Facebook Like Buttons or Google Analytics. uMatrix allows 1st party scripts in its default configuration. If you want the default functionality of NoScript consider <a href=https://github.com/gorhill/uMatrix/wiki/How-to-block-1st-party-scripts-everywhere-by-default>blocking 1st party scripts everywhere by default</a>. <strong>This addon has been <a href=https://github.com/uBlockOrigin/uMatrix-issues/issues/291#issuecomment-694988696>discontinued</a>.</strong>"
   website="https://addons.mozilla.org/firefox/addon/umatrix/"
   privacy-policy="https://github.com/gorhill/uMatrix/wiki/Privacy-policy"
   forum="https://forum.privacytools.io/t/discussion-umatrix/271"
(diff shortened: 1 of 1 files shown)


Mikaela (approved):
LGTM, except that the descriptions are a bit of wall of text especially on mobile and I wonder if some line changes would make it more pleasant to look at.

freddy-m (approved):

3k2:
So what about noscript? far as I can tell it was removed due to fact that umatrix does the same thing now that umatrix isn't being maintained anymore is it good idea to add it back or not?

dngray:
> So what about noscript? far as I can tell it was removed due to fact that umatrix does the same thing now that umatrix isn't being maintained anymore is it good idea to add it back or not?

uBlock advanced can do the same as NoScript, you can disable JavaScript and then enable it for various things.

See [Advanced mode](https://github.com/gorhill/uBlock/wiki/Advanced-user-features) with [dynamic filtering](https://github.com/gorhill/uBlock/wiki/Advanced-user-features). Take a look the differences between [easy mode](https://github.com/gorhill/uBlock/wiki/Blocking-mode:-easy-mode), [medium mode](https://github.com/gorhill/uBlock/wiki/Blocking-mode:-medium-mode), or [hard mode](https://github.com/gorhill/uBlock/wiki/Blocking-mode:-hard-mode).

dngray:
One thing to add, uBlock currently works with Fenix (so if you use Firefox on Android), whereas uBlock does not.

3k2:
> > So what about noscript? far as I can tell it was removed due to fact that umatrix does the same thing now that umatrix isn't being maintained anymore is it good idea to add it back or not?
> 
> uBlock advanced can do the same as NoScript, you can disable JavaScript and then enable it for various things.
> 
> See [Advanced mode](https://github.com/gorhill/uBlock/wiki/Advanced-user-features) with [dynamic filtering](https://github.com/gorhill/uBlock/wiki/Advanced-user-features). Take a look the differences between [easy mode](https://github.com/gorhill/uBlock/wiki/Blocking-mode:-easy-mode), [medium mode](https://github.com/gorhill/uBlock/wiki/Blocking-mode:-medium-mode), or [hard mode](https://github.com/gorhill/uBlock/wiki/Blocking-mode:-hard-mode).

oh yes I forgot about that sorry! I feel kinda sad that umatrix isn't maintained anymore but I suppose it makes sense since ublock can do most of the things what umatrix did. So there's no point maintaining both which essentially do same.

dngray:
> So there's no point maintaining both which essentially do same

That's pretty much what the author said (Raymond Hill) said:

https://github.com/uBlockOrigin/uMatrix-issues/issues/291#issuecomment-694988696

unbranched:
In no way uBlock can replace uMatrix, or I'm missing some functionality:
-I set uBlock in advanced mode to block everything by default globally except images
-I visit a website that doesn't display correctly because of blocked css/third party images
-I allow the first party domain

Now the website works, but this implies enabled cosmetic elements and everything else, including scripts. What if I just want to display the website correctly while not running 1st party scripts? I can't.

I have no other choice than using ublock+noscript .

justjosias:
There's nothing to worry about for now. It will be a while before uMatrix becomes outdated to the point it doesn't make sense to use it.

lynn-stephenson:
We have no intention of recommending EOL software. Even if there are no huge issues with uMatrix.

computeristic:
@unbranched you can allow/block specific elements (images/CSS/DOM/scripts...) on any domain you visit using the logger on ublock. Check [this](https://github.com/gorhill/uBlock/wiki/The-logger) page on the wiki and pay special attention to the 
'Dynamic URL filtering rules'.

unbranched:
@computeristic Thanks, I missed that feature. That's kind of tedious though, compared to uMatrix.

LongJohn-Silver:
> wait, is there a config syntax to make it block eg. all scripts by default? That would be awesome.

Yes, you need do block first and third party scripts if you want to block all by default, although I recommend you to only block third one since otherwise you won't be able to load any website properly, and blocking javascript by default already does that.

Screenshot from 2021-01-20 12-51-09

PR #2081, opened 2020-10-07
The great browser section cleanup

dngray:
## Description

Resolves: https://github.com/privacytools/privacytools.io/pull/2005
Resolves: https://github.com/privacytools/privacytools.io/pull/1931
Resolves: https://github.com/privacytools/privacytools.io/issues/1326

#### Check List 


blacklight447-ptio:
Still not sure if outright removing Decentraleyes is the correct way to go. It may not help if people enable FPI, but if they don't then decentraleyes will atleast give you partial protection.

dngray:
> will atleast give you partial protection.

Not really no. The reason we would suggest removing it is because it [doesn't actually work](https://github.com/arkenfox/user.js/issues/948) as the resources are [horribly out of date](https://git.synz.io/Synzvato/decentraleyes/-/tree/master/resources).

@Thorin-Oakenpants does describe https://github.com/privacytools/privacytools.io/issues/1430#issuecomment-704335991 why FPI is really the only way to achieve these addons set out to do.

That being said, when LocalCDN is available for Fenix, we could revisit this. [According to the author](https://old.reddit.com/r/privacytoolsIO/comments/j6lv30/should_i_use_localcdn_instead_of_decentraleyes/g80zln0/) it works in a different mode of operation where:

> You can use LocalCDN in two ways. In both variants existing libraries are delivered offline via LocalCDN instead of loading online via one or more CDNs. Decentraleyes delivers 1 to 1 the requested version. LocalCDN could do that as well, but then the extension with all 123 frameworks would be 50 or 100 MB. That's why LocalCDN will upgrade the request. For example, if the website requests jQuery v1.7.0 but LocalCDN contains v1.7.1, the newer version will be used. This saves storage and allows me to integrate even more libraries. Now you have two options: If a library is missing, you can fetch it from the CDN (lower privacy) or block the request.

> The upgrade works for most websites. Unfortunately there are always exceptions because the internet is broken. There are over 100 different jQuery versions. Many websites use completely outdated technologies. If something doesn't work, just open a ticket on Codeberg so I can check and reference the changes to the code there. If libraries are missing, I'll of course integrate them quickly. Currently there are 30 CDNs and 123 frameworks in LocalCDN.

This would be infinitely more useful than Decentraleyes, even if your use case is simply to save bandwidth.

I think like other things, we really should be suggesting people do things that don't actually work. I have been using `privacy.firstparty.isolate` and have to say I [haven't found anything broken yet](https://www.ctrl.blog/entry/firefox-fpi.html). The only one there that is really meaningful is third party logins, and nobody really should be using social sso logins anyway as those are terrible for privacy. Yay for grouping all your activities to a particular Facebook profile for example!.

dngray:
I've decided with we will do separately https://github.com/privacytools/privacytools.io/issues/1257

dngray:
I've decided to tackle https://github.com/privacytools/privacytools.io/issues/1257 in it's own PR, not this one.

zero77:
@dngray 
I imagine you already have this covered.

It's worth adding that when it comes to anti fingerprinting.
It's better to spoof trackers than block as a fingerprint can still be made of you if you block.

Thorin-Oakenpants:
> It's worth adding that when it comes to anti fingerprinting.
> It's better to spoof trackers than block as a fingerprint can still be made of you if you block.

That's way too generalized, and makes no sense. I can't even tell what you mean

If you mean it's better to spoof to trackers than to block trackers: I disagree. Blocking the source of the FPing is the first step of many (but ultimately a game of whack a mole). Actual fingerprinting countermeasures are for when something gets through

If you mean disabling an API vs dealing with it, then the only reason to do that would be because there isn't a solution. For example, Tor Browser disable the web audio API. The entropy comes from floating points and the math libraries play a role. Since legitimate web audio API use is pretty much non-existant, then it's easier for RFP/Tor Uplift/TB to de-prioritize it and kick it upstream to the standards body. Meanwhile, all TB users are still exactly the same on that web audio metric: so it's effectively the same as if they were spoofing as far as entropy goes

It really depends on the metric. Generally speaking, you want anti-fingerprinting to cause as little breakage or side-effects as possible: but if there's no solution, it's better to disable the API and everyone will be the same, than allow the entropy to leveraged. Note: we're talking about sets of users: you cannot hide your engine, you cannot hide that you are FF vs TB, you cannot hide if you have RFP on or not, etc.

/end of rambling

github-account1111:
Should CDNs be un-nooped in uBO once I uninstall Decentraleyes, if FPI is on?

dngray:
> Should CDNs be un-nooped in uBO once I uninstall Decentraleyes, if FPI is on?

If you're using advanced mode/umatrix/noscript in advanced/hard mode [the noop rules were so that they could be served locally](https://git.synz.io/Synzvato/decentraleyes/-/wikis/Frequently-Asked-Questions#for-umatrix-and-ublock-origin-non-easy-mode-users).

If you're still using those addons in that way you'd need to allow them. [Noop just disables dynamic filtering](https://github.com/gorhill/uBlock/wiki/Dynamic-filtering:-rule-syntax#actions) (filter lists), for those CDNs.

github-account1111:
@dngray I'm sorry I meant disallow, not un-noop. Of those addons, I only use uBO (in Medium Mode, i.e. dynamic filtering is on). Decentraleyes requires to allow those CDNs globally. Do I need to disallow them and then noop one-by-one on the pages that use them like with all the other domains?

github-account1111:
@dngray why would I want to allow them? I still want to use the default blocking lists on everything, including CDNs (*especially* CDNs). Right now they're nooped globally (as per the Decentraleyes requirement that you linked).

I'm asking if I should un-noop them globally and start nooping on a per-page basis like I do with the other domains. That'd be kinda frustrating, since so many sites use them. But if I keep them nooped globally, would that be worse privacy-wise than doing the same but with Decentraleyes installed?

Asheq:
As far as I can tell, the following recommendations are _already_ the default values:

- `browser.send_pings`  = false
- `network.dns.disablePrefetchFromHTTPS`  = true
- `network.predictor.enable-prefetch`  = false

So, I think they can be removed for the sake of brevity, right?

rharish101:
Firefox 86 has introduced a concept called ["Total Cookie Protection"](https://blog.mozilla.org/security/2021/02/23/total-cookie-protection/) for both desktop and Android. The Temporary Containers add-on is now probably unnecessary. If this is true, then we can remove that section.

Further, they are introducing dynamic first-party isolation if Enhanced Tracking Protection is set to "strict". So `privacy.firstparty.isolate` can also be removed, as the new feature provides the same protection while having exceptions for certain third party login systems.

rusty-snake:
> Further, they are introducing dynamic first-party isolation if Enhanced Tracking Protection is set to "strict".

"Total Cookie Protection" _is_ dFPI.

> So privacy.firstparty.isolate can also be removed, as the new feature provides the same protection

FTR: AFAIK only if `privacy.partition.network_state=true` (default for FF 85+).

ETP=strict is not the default. If FPI will be removed, it must also be recommended to set this to strict, otherwise it's much lower protection. 

> The Temporary Containers add-on is now probably unnecessary. 

It's still "necessary" (whatevery this means), because it provides automatic clean within a session. dFPI only isolates different sites, while TC can also isolate the same site. Assuming that the automatic mode is used.

Victor239:
Does Cookie AutoDelete cover the same cleaning that Temporary Containers does? It allows automatically removing cookies, LocalStorage, cache, IndexedDB, plugin data and service workers.

rusty-snake:
https://github.com/stoically/temporary-containers/wiki/Comparison#cookies-autodelete

nourkagha:
New Firefox 87 update has introduced a new [default HTTP Referrer policy](https://blog.mozilla.org/security/2021/03/22/firefox-87-trims-http-referrers-by-default-to-protect-user-privacy/) and [SmartBlock](https://blog.mozilla.org/security/2021/03/23/introducing-smartblock/).

GintokiHub:
@dngray and all this fork of decentraleyes seems to be getting very frequent updates has anyone looked into it?
https://codeberg.org/nobody/LocalCDN

partulaj:
Firefox recently introduce [site isolation](https://blog.mozilla.org/security/2021/05/18/introducing-site-isolation-in-firefox/), it seem's to me that it would be a good addition to this PR

youdontneedtoknow22:
There are so many merge requests that aren't in the preview page, that I can't keep up with all of them.
There are some points I want to add:
1. Maybe we should let people set Firefox's protection to strict, and then tweak the about:config settings, that are still not tweaked. (People love to work with guis).
2. uBO: we should tell users that the default settings are okay. However, the recommended mode is the medium mode (sugessted by Raymond Hill). I personally lernt how to use it perfectly from a youtube video from (The Hated One) and I highly recommend to put a link for it. There's also a trick to activate that "green option", which is in my eyes important when using the medium mode. Some websites will break and you can't fix them without it.
3. Why use an addon (xbrowsersync) to increase the attack surface and cpu usage, when you can make a firefox account to synchronize between devices? Firefox already uses e2ee and you can set a 2FA for your account. If you really want to use it, this addon should be not under the "recommended Addon" section, but rather "Additional functionality"
4. On Windows, we should advice people, who saves their passwords in their browser (most people will do, believe me) to use a master password. Even if the device got comprissed or someone had physical access to the device, they can't access the passwords.
5. We should also state that Containers Addons won't isolate sites from each other better than FPI. But it will let people isolate their profiles/Accounts on the same websites from each other. (This gets asked really really alot on reddit and it has to stop)
6. uMatrix shouldn't be recommended anymore
7. Is Canvas Blocker necessary? I tested my canvas signature and I get a different one with a new window. It's probably due to (privacy.resistfingerprinting). We don't need to make that section any more complicated and make firefox buggier with all these addons.
8. We should write some possible "side effects" of the about:config tweaks, so that people already know what problems they will have and can easily fix them. We're working on it here:
https://github.com/privacytools/privacytools.io/issues/2347

9. Someone should get started on this, it has taken a really long time.

github-account1111:
Agree with 4-9 but:

>People love to work with guis

I think doing nothing (i.e. how it is now, just double-click a script that transfers the user.js) is preferable to working with a GUI.

>There's also a trick to activate that "green option", which is in my eyes important when using the medium mode. Some websites will break and you can't fix them without it.

Unless you can link those websites you're referring to (and they're *actually* broken until you allow a domain), that kinda tells me you haven't learned how to use it perfectly.

https://github.com/gorhill/ublock/wiki/Overview-of-uBlock's-network-filtering-engine

>Why use an addon (xbrowsersync) to increase the attack surface and cpu usage, when you can make a firefox account

Because it's always better to be platform-agnostic and work with your data yourself when it comes to privacy (and convenience in this case, as you're not locked into Firefox). Not to mention, it's a bad idea to sign into a Firefox account.. in Firefox. Don't do it. Also, CPU usage? lol

youdontneedtoknow22:
Copied from another comment:
ClearURLs isn't availabe on android, but uBO is. We probably need some more testing, but uBO can now replace ClearURLs with an Adguard List (both on mobile and desktop). And if it's done correctly then again: less cpu usage + less attack surface.
https://www.reddit.com/r/privacytoolsIO/comments/ooie4u/psa_ublock_origin_added_two_new_stock_filter/

youdontneedtoknow22:
> I think doing nothing (i.e. how it is now, just double-click a script that transfers the user.js) is preferable to working with a GUI. Besides, if they're already working with about:config, why overcomplicate it by introducing a new thing?

Because as I said, people will think it's advanced tweaks and can only done with you agree to "accept the risk of modifing these values". Doing it on in the gui takes even less time.
And PTIO isn't interessted in coping some user.js, they like to do their own thing.

> 
> > There's also a trick to activate that "green option", which is in my eyes important when using the medium mode. Some websites will break and you can't fix them without it.
> 
> Unless you can link those websites you're referring to (and they're _actually_ broken until you allow a domain), that kinda tells me you haven't learned how to use it perfectly.
> 
> https://github.com/gorhill/ublock/wiki/Overview-of-uBlock's-network-filtering-engine
> 
> Do you think gorhill would remove the option in the first place, if it were a good one?

I apologzie on this one. After playing a little with the green button, it didn't offer me any help. I was screwing up my uBO setup with other ways, that the grey buttion didn't seem to work on some websites.
> 
> > Why use an addon (xbrowsersync) to increase the attack surface and cpu usage, when you can make a firefox account
> 
> Because it's always better to be platform-agnostic and work with your data yourself when it comes to privacy (and convenience in this case, as you're not locked into Firefox). Not to mention, if you care at all about privacy, it's a bad idea to sign into a Firefox account.. in Firefox. Don't do it. Also, CPU usage? lol

If it's e2ee encrypted, I would rather trust Mozilla rather then that addon. An Account lets you sync your passwords too (when you're using Firefox Clockwise, which a lot of people will already do. I'm not going to discuss what it offers when comparing to Bitwarden and its addon).
And yeah, you won't be locked into firefox, but that's the recommended browser lol. If if the user wants to use something else, well that's why there's a section called "Additional functionality" instead of making the illusion that using this addon will make you "more private".

dngray:
In this PR, the plan is to move away from specifying any `about:config` defaults, and go over to providing screenshots with UI facing options, which is necessary for Google Play versions of Firefox anyway. Advanced users should be directed towards https://github.com/arkenfox/user.js

dngray:
> ClearURLs isn't availabe on android, but uBO is. We probably need some more testing, but uBO can now replace ClearURLs with an Adguard List (both on mobile and desktop). And if it's done correctly then again: less cpu usage + less attack surface.
https://www.reddit.com/r/privacytoolsIO/comments/ooie4u/psa_ublock_origin_added_two_new_stock_filter/

We should make that part of this PR.

ph00lt0:
> In this PR, the plan is to move away from specifying any `about:config` defaults, and go over to providing screenshots with UI facing options, which is necessary for Google Play versions of Firefox anyway. Advanced users should be directed towards https://github.com/arkenfox/user.js

@dngray please consider https://github.com/privacytools/privacytools.io/issues/2248 it comes with all right settings out of the box. I think this is a lot easier for the general user.

Another question, does anybody know if uBO also prevents ETag tracking? This currently is also a reason for listing ClearURLs.

rusty-snake:
> does anybody know if uBO also prevents ETag tracking?

It don't. The work etag does not even appear in it's codebase.

> ETag tracking

Who cares? It's isolated.
Do you block all cookies? And disable TLS Session tickets?

ph00lt0:
@rusty-snake if `privacy.firstparty.isolate` is on I believe you are right. But generally on mobile this won't be the case for most users as afaik this is not enabled by default. As @dngray also said about:config manually isn't very accessible to the average user. It might actually discourage people from doing so. Privacy should be easy and for everyone, also to make it more effective.

youdontneedtoknow22:
> 
> 
> In this PR, the plan is to move away from specifying any `about:config` defaults, and go over to providing screenshots with UI facing options, which is necessary for Google Play versions of Firefox anyway. Advanced users should be directed towards https://github.com/arkenfox/user.js

Is that a good decision tho?
Some tweaks can't be achieved using the UI, such as "privacy.resistFingerprinting = true" being the most important one to resist fingerprinting, and "beacon.enabled = false" and others.

I also wanted to link a discussion from firefox's github about FPI, dFPI (= Total Cookie Protection) and all these terms. To summarize it, dFPI is their way to implement FPI in a more web compatible way. We can enable dFPI just by setting ETP in the UI to strict (which is also availabe for Firefox on android).
https://github.com/mozilla/multi-account-containers/issues/1974
So "privacy.firstparty.isolate" is no longer needed, and as I understood from others, this will also isolate ETags. So ETag Stoppa and its replacment ClearURLs are also no longer needed I suppose.

rusty-snake:
Here's your reading: https://blog.mozilla.org/security/2021/01/26/supercookie-protections/

INANE for firefox under android, but I have no hint that this isn't the case for android.

github-account1111:
> And PTIO isn't interessted in coping some user.js, they like to do their own thing.

github-account1111:
> And PTIO isn't interessted in coping some user.js, they like to do their own thing.

Not _some_ user.js. Your own user.js.

>Doing it on in the gui takes even less time.

Not really. You have to dig through settings and look for the right buttons and checkmarks and stuff.

>I would rather trust Mozilla rather then that addon

That's a bad mindset. You should always trust local (an open source addon) vs cloud (Mozilla).

>And yeah, you won't be locked into firefox, but that's the recommended browser lol

Recommended where? On Windows? Mac? iOS? That's why I said platform-agnostic. For example, you probably shouldn't be using Firefox on Android. And you _definitely_ shouldn't be using it on iOS. What are you gonna do about your bookmarks there? Not to mention, some platforms don't even have Firefox.

Thorin-Oakenpants:
ph00lt0 said
> please consider Mull Browser (#2248) it comes with all right settings out of the box. I think this is a lot easier for the general user.

This is a complete parroting of what you said about Librewolf. I get that you're keen, but stop pushing obscure browsers and provide facts, not opinions - in the appropriate issue, not here. Why are these are all the right settings? How do **_you_** know? Where are your references and proof? What are your credentials/experience in all of this (optional but lends credence)? You also suggested a problematic extension Privacy Possum as an alternative, and one that has been abandoned for 3 years, for a problem that does not exist. And you keep making incorrect statements about a number of Firefox developments.

Instead of personally blocking me because you don't like my factual answers, you should read what I'm telling you. Blocking someone doesn't suddenly make your points correct

> does anybody know if uBO also prevents ETag tracking? This currently is also a reason for listing ClearURLs

Are you not able to check uBO yourself? And, no, etags are **_not currently_** a reason for listing ClearURLs. Etags are not even an issue. Neither is it the history API setting (this is a myth), nor the hyperlink auditing (you can use a pref). It is because it "clears" urls of tracking parameters. Side note: if uBO's new filters cover this, then ClearURLs could be probably be dropped IMO - needs a discussion, analysis elsewhere

rusty-snake: "ETag tracking" Who cares? It's isolated. Do you block all cookies? And disable TLS Session tickets?
ph00lt0: if `privacy.firstparty.isolate` is on I believe you are right. But generally on mobile this won't be the case

me right now: to add to rusty's comment "do you change your IP"?

It was already [pointed out](https://github.com/privacytools/privacytools.io/issues/2381#issuecomment-883058380) that etags are not an issue since FF85. **_More reading, less talking_**. [Here](https://groups.google.com/g/mozilla.dev.platform/c/uDYrtq1Ne3A) is a link to what network partitioning covers. This is enabled by default for all users, all platforms

If you still think etags are an issue, then please explain why, so I can explain why it isn't.

youdontneedtoknow22:
So I did some digging how does the new list from Adguard compare to ClearURLs. I believe the new list from adguard has a long way a head of it before it catches the list form ClearURLs (unless they copy their work, which makes more sense IMO). Also some people on Reddit pointed out that the list from adguard didn't remove the parameters from sites like bing (do they even have referrral parameters?) and some parameters from amazon.
Here's the post: https://www.reddit.com/r/uBlockOrigin/comments/oothk8/psa_ublock_origin_added_two_new_stock_filter/

And here's the list for ClearURLs: https://gitlab.com/anti-tracking/ClearURLs/rules/-/raw/master/data.min.json
And here's the list from Adguard: https://raw.githubusercontent.com/AdguardTeam/FiltersRegistry/master/filters/filter_17_TrackParam/filter.txt

You can compare the parameters for each site. I compared 2 or 3 (including amazon), and CleanURLs seems to have more parameters.
Also Bing isn't mentioned in the Adguard List (and its parameters aren't in the general parameter list in the beginning of their list) and lots and lots of other websites, like aliexpress for example.

Everything I said could be wrong and I may didn't understand the whole concept of their lists, so feel free to correct me (while still being polite, I'm trying to learn for myself and to protect my privacy and benefit others from this, just like most people who are spending their free time discussing such topics here)

PR #2146, opened 2020-12-05
LibreDNS doesn't support DNSSEC

dngray:
https://deploy-preview-2146--privacytools-io.netlify.app/providers/dns/


dngray:
Actually I tested this and it does.

PR #2167, opened 2020-12-26
Fix for tooltips on badge

Gusted:
## Description

Resolves: https://github.com/privacytools/privacytools.io/pull/2139#issuecomment-751338425 @dngray 

#### Check List 

* Netlify preview for the mainly edited page: https://deploy-preview-2167--privacytools-io.netlify.app/

### DIFF
--- a/assets/css/style.scss
+++ b/assets/css/style.scss
@@ -408,9 +408,12 @@ input#nav-toggle,
 }
 
 /* Vanilla CSS implementation for tooltip */
-[data-toggle="tooltip"]:hover {
+[data-toggle="tooltip"] {
   cursor: pointer;
   position: relative;
+}
+
+[data-toggle="tooltip"]:hover {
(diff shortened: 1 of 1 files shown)


dngray:
Thanks for this, will review!

dngray:
I had a look at this, any chance we can get it not bold



Like if you look how it was originally:

dngray:
Hmm, i think setting `font-weight: normal;` might be the right solution. Though I would have thought it would have been normal to begin with. Must have inherited the boldness from somewhere else.

Gusted:
> Must have inherited the boldness from somewhere else.

The badge itself seems to have an `bold` font-weight.

> Hmm, i think setting `font-weight: normal;` might be the right solution.

I couldn't think of anything else.

dngray (approved):

Issue #2144, opened 2020-12-01
📝 Correction | Invidious wording is incorrect - you are still communicating with Google


bkil:
## Description

If you open up a video from an Invidious instance and inspect the networking tab of the developer toolbar, you will clearly find that your video chunks are coming from `googlevideo.com`. Google not only sees your IP, but they know what you are watching, your connection speed, which parts you find interesting (seek/play/pause), whether you need to speed up or down for comprehension and may fingerprint you based on headers as well. I haven't checked what else they can do.

## Why I am making the suggestion

The existing wording is misleading: `you can keep watching videos without sharing your IP`.

## My connection with the software

No connection.

- [x] I will keep the issue up-to-date if something I have said changes or I remember a connection with the software.

bkil:
Oh, I found a hint in the documentation, this seems to be an optional feature:
- https://github.com/iv-org/documentation/blob/master/Always-use-%22local%22-to-proxy-video-through-the-server-without-creating-an-account.md

As this was not obvious, I think it would be beneficial to mention this.

Gusted:
Hi @bkil!

Please check my comment over at #2182 I think you might want to add some information to it.

Regards,
Gusted

bkil:
Thank you, I think the warning is adequate from my perspective.

Issue #2135, opened 2020-11-24
🌐 Website Issue | Latest commits not appearing


samuel-lucas6:
## Description
The last 5 commits (#2115, #2129, #2124, #2114, #2108) haven't appeared on the website yet. There appears to be some serious delay - the oldest was 25 days ago.

dngray:
Hmm this is an issue we'll have to look into. Not sure why this happens sometimes.

dngray:
Seems to have happened to the new email client page https://github.com/privacytools/privacytools.io/commit/5142ef18ab4d762b3570c6552a780685a1c20cb6

dngray:
Fixed

PR #2182, opened 2021-01-06
Add warning to Invidious

Gusted:
## Description

Resolves: #2144

As described in #2144 the current box is misleading, which I agree with. To clarify why I agree with this:
- As a new user(private window) and open a video you can see it will by default request the video from an google server `googlevideo.com` as seen in the [source code](https://github.com/iv-org/invidious/blob/168376b04671c6e362468267a3b07dd332b40702/src/invidious.cr#L211) where they add the CSP headers when the user has not enabled `local`.
- As seen in the [source code](https://github.com/iv-org/invidious/blob/168376b04671c6e362468267a3b07dd332b40702/src/invidious/helpers/helpers.cr#L38) the default preference for `local` which is the preference to see if the user wants proxied video, is false and by default not proxied if not otherwise changed in the preference *if the instance allows this, some have disabled this to be enabled*.
- As seen in the [source code](https://github.com/iv-org/invidious/blob/168376b04671c6e362468267a3b07dd332b40702/src/invidious.cr#L3479) it will look if the host parameter is in the request and use the custom host if the parameter is found. The call to this webpoint is handled in videos.cr that added the host parameter when user prefers(local preference) it.

I've now added an warning however, I think the whole `proxy` aspect to this can be removed as invidious [main goal](https://github.com/iv-org/invidious#invidious-is-an-alternative-front-end-to-youtube) is provide an front-end and not serve as proxy. I think it's good to make people aware of this as I've heard great things over invidious even to degoogle, however knowing that it's not by default proxied and some instances don't allow this is not common knowledge.

#### Check List 

* Netlify preview for the mainly edited page: https://deploy-preview-2182--privacytools-io.netlify.app/

### DIFF
--- a/_includes/sections/video-frontends.html
+++ b/_includes/sections/video-frontends.html
@@ -8,7 +8,8 @@
   title="Invidious"
   image="/assets/img/svg/3rd-party/invidious.svg"
   description='Invidious is an alternative front-end to YouTube. It is free software, with no advertising or Javascript dependency to play videos, with lots of other features that allow you to have a complete YouTube experience, sans Google.'
+  labels="color==warning::icon==fas fa-exclamation-triangle::link==https://github.com/iv-org/documentation/blob/master/Always-use-%22local%22-to-proxy-video-through-the-server-without-creating-an-account.md::text==Warning::tooltip==By default, Invidious will not proxy videos through the instance's proxy."
   website="https://invidio.us"
-  github="https://github.com/omarroth/invidious"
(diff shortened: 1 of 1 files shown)


freddy-m (approved):

LongJohn-Silver:
Was this fully merged? I can still see the same phrasing on the web site.

Gusted:
Reminds me of #2135

LongJohn-Silver:
Any ideas, @dngray?

PR #2185, opened 2021-01-10
Cleanup Javascript

Gusted:
- Remove sortable.js(not used).
- Cleanup redirects, I couldn't properly look at it.
- Made main.js ECMA6 ready.
- Made event functions into a declaration to ensure compatibility with possible older browser version.
- Removed useless `=== true/false`.
- tempSrc uses declaration now.

## Description

Resolves: none

For someone who spends a lot of time in Javascript, the javascript at privacytools.io was something to not prefer not look at it. So I've made some changes that should have none to little impact in the end result, It's just some code quality and code inconsistency fixed. You can refer to the Summary what exactly has been done. As the code was already targeting ECMA6 I updated some code that was readable in ECMA6 to make it more consistency. Thereby removed some useless `=== true/false` as JavaScript will automatic evaluate if the return result is true/false. I saw the `sortable.js` and saw already a security flaw(A variable was reinterpreted as HTML without escaping meta-characters), however fiddling around I saw it has no use-case and not used in production and thereby removed and fixed a little issue with `tempSrc` to define its a `const`. Looking at `redirect.js`, was giving this whole idea to clean it up and moving it to a object and do the `window.location` just once gives it a better 'maintainability status' to add/modify extra redirects. I've replaced some functions that were passed into events into `const x = () =>` so it will be used as declaration rather then a new function, so in the further if e.g. the removing of this event is needed it will use the declaration and not pass off a new function and gives you some good hours debugging(talking from experience). I didn't really saw a discussion/issue open for this, but I think everyone agrees to have better quality in the code.

Summary:
- Remove sortable.js(not used).
- Cleanup redirects, I couldn't properly look at it.
- Made `main.js` ECMA6 ready.
- Made event functions into a declaration to ensure compatibility with possible older browser version.
- Removed useless `=== true/false`.
- tempSrc uses declaration now.

#### Check List 


jonaharagon:
Didn't look at the other changes, but FYI sortable.js is definitely used in production. https://www.privacytools.io/providers/dns/ for example.

Gusted:
> Didn't look at the other changes, but FYI sortable.js is definitely used in production. https://www.privacytools.io/providers/dns/ for example. (I don't think there is any unused code in general but I may be wrong)

I thought so too, but the problem with this was when I checked at the network request, `sortable.js` was never requested and doing a quick search in the code doesn't mention any reference. Well the [netlify preview](https://deploy-preview-2185--privacytools-io.netlify.app/providers/dns/) on dns provider doesn't show anything wrong with it.

Issue #2186, opened 2021-01-10
🌐 Website Issue | Enable compression on text


Gusted:
## Description

Currently, `privacytools.io` only compress the html pages with gzip, you can confirm this by executing `curl -IL https://privacytools.io -H "Accept-Encoding: gzip, deflate, br" | grep "content-encoding"` whic will return `gzip`. If you look further and request an .css or .js you will see no compression is being applied. Compressing files has great benefits, lower latency(time to load the page) and lower network transfers(amount of data). I got confirmation over the development matrix that currently 'NGINX' is the server which has [good support for compression](https://docs.nginx.com/nginx/admin-guide/web-server/compression/) along-side the new brotli(br) compression. However it seems like the configuration for compression is the default one which doesn't include css(text/css)/javascript(application/javascript). This should be *simple* to add to the configuration to have compression on those files as well, which can be quite resourceful. In addition support for the [brotli](https://github.com/google/brotli) compression the [brotli module](https://github.com/google/ngx_brotli) can be added. 

As a side-note make sure the assets are enabled trough the static module of gzip/brotli, because it's not dynamic content. So they wouldn't be compressed on every request.

I've saw that @blacklight447-ptio is the current sysadmin so mentioning him.

Regards,
Gusted

jonaharagon:
> As a side-note make sure the assets are enabled trough the static module of gzip/brotli

That would require `.br` files for every assets in this repo though, which aren't currently being generated. I suppose that could be done as well however, with https://github.com/philnash/jekyll-brotli @dngray

Gusted:
> That would require `.br` files for every assets in this repo though, which aren't currently being generated. I suppose that could be done as well however, with https://github.com/philnash/jekyll-brotli @dngray

Having this the static enabled for most files that are text-based will be great for CPU Usage as it doesn't have to compress it every time when the server needs to serve it, as I remembered their are no dynamic pages in `privacytools.io`, it can be enabled for the HTML pages as well.

Gusted:
I'm not sure what happened last night, but I guess that `brotli` is now enabled? It will return on the assets a `br` content-encoding when possible. 
![image](https://user-images.githubusercontent.com/25481501/104156574-a6736d80-53e9-11eb-999f-474032d2eccf.png)

Gusted:
Hey @jonaharagon

Do you have any information what happened and how it's now enabled?

Regards,
Gusted

jonaharagon:
[...]

PR #2189, opened 2021-01-12
Enable Brotli compression

Gusted:
## Description

Resolves: #2186

As in the *discussion* #2186, adding the `jekyll-brotli` plugin to compress the static assets on building, so they don't have to be compressed each time when the server wants to serve them. Additionally server-side the static module has to be configured to use the `*.br` files. Some configuration has changed regarding `github-pages`, it will publish the `_site` to a github repo when the ENV was set to `production`. Removing it from the `jekyll-plugins` group and adding manually into the `_config.yml` fixed this issue. 

Summary:
- Compress file AOD.
- Let the server only use static files. 
- Don't add github-pages to 'jekyll-plugins' group, due to the fact github-pages will behave differently when enabling production ENV.
- Adds an extra 10s for the `host` to use `./build.sh` due to compressing all files.

#### Check List 


Gusted:
Succeeded by #2194

jonaharagon:
I didn't realize this was using `github-pages` instead of `jekyll` in the first place until your PR, so I think some wider changes were necessary 😆

Gusted:
>  didn't realize this was using `github-pages` instead of `jekyll` in the first place until your PR,

I tried to remove it, but got into some dependency hell. So I had left it to have a another look this weekend. Luckily your PR seems to include the 2 dependency that are needed :D

PR #2168, opened 2020-12-26
Update Ruby Dependencies

Gusted:
## Description

Resolves: https://github.com/privacytools/privacytools.io/pull/2166#issuecomment-751337663

In the issue linked above, it shows some errors that's being caused in the latest ruby-dev version at the time v2.7, this gives deprecation notices due to using old dependencies which still uses this behavior. 

#### Check List 

- [x] I understand that by not opening an issue about a software/service/similar addition/removal, this pull request will be closed without merging.

- [x] I have read and understand [the contributing guidelines](https://github.com/privacytools/privacytools.io/blob/master/.github/CONTRIBUTING.md).

- [x] The project is [Free Libre](https://en.wikipedia.org/wiki/Free_software) and/or [Open Source](https://en.wikipedia.org/wiki/Open-source_software) Software

* Netlify preview for the mainly edited page: https://deploy-preview-2168--privacytools-io.netlify.app/

* Code repository of the project (if applicable):

PR #2194, opened 2021-01-15
Various Jekyll/build improvements

jonaharagon:
**Changes:**
 - Specify specific Ruby version to use in `.ruby-version`, to ensure developers are using identical software to build machines and Netlify.
     - Set this value to 3.0.0, up from current builds using 2.6.2. Additionally I've added two gems that are now required for Jekyll Ruby 3 support.
     - We want to avoid issues like #2168 in the future, as updating Ruby locally without coordinating with site builders can break production builds.
     - Updated Ruby installation instructions in README.
 - Upgrade Bundler to 2.2.5
     - Locking to 2.0.1 in the first place was a workaround for a Netlify build bug which is now resolved.
 - Switched from `github-pages` gem to `jekyll` gem.
 - Upgrade Jekyll to 4.2.0
     - No longer depends on `ruby-sass`, which is EOL.
 - Adds brotli static file generation.
     - Merging this PR will automatically merge #2189 as well. Thx @Gusted 
 - Renames `build.sh` to `.buildrc`.
     - This new file is `source`d within the build scripts on production servers. Commands should be assumed to be running in the main directory (`$HOME/privacytools.io`) directly following a fresh pull from GitHub.


jonaharagon:
Blocked by https://github.com/netlify/build-image/issues/517

PR #2154, opened 2020-12-17
Adding freepost

freddy-m:
## Description

Resolves: #2001 

#### Check List 

- [x] I understand that by not opening an issue about a software/service/similar addition/removal, this pull request will be closed without merging.

- [x] I have read and understand [the contributing guidelines](https://github.com/privacytools/privacytools.io/blob/master/.github/CONTRIBUTING.md).

- [x] The project is [Free Libre](https://en.wikipedia.org/wiki/Free_software) and/or [Open Source](https://en.wikipedia.org/wiki/Open-source_software) Software

* Netlify preview for the mainly edited page: 

* Code repository of the project (if applicable): https://notabug.org/zPlus/freepost

Issue #2140, opened 2020-11-26
🆕 Software Suggestion | Lemmy


LongJohn-Silver:
## Basic Information

**Name:** Lemmy
**Category:** Social Networks/Social News Aggregators
**URL:** https://lemmy.ml/ https://github.com/LemmyNet/lemmy

## Description

Lemmy is similar to sites like Reddit, Lobste.rs, or Hacker News: you subscribe to forums you're interested in, post links and discussions, then vote, and comment on them. Behind the scenes, it is very different; anyone can easily run a server, and all these servers are federated (think email), and connected to the same universe, called the Fediverse.

For a link aggregator, this means a user registered on one server can subscribe to forums on any other server, and can have discussions with users registered elsewhere.

The overall goal is to create an easily self-hostable, decentralized alternative to Reddit and other link aggregators, outside of their corporate control and meddling.

Each Lemmy server can set its own moderation policy; appointing site-wide admins, and community moderators to keep out the trolls, and foster a healthy, non-toxic environment where all can feel comfortable contributing.

## Why I am making the suggestion

Recently Lemmy was able to release its federation so now different instances can be created and they can interact with each other, besides this, it is a libre and privacy respecting alternative to websites like Reddit or Lobsters while also aiming at not providing spaces for right wing organizations. Listing them would also help them to gain some recognition and with more people aware of the project they would probably get more people to help them with development.

As a side note, I think it would be a really awesome idea if the team behind Privacy Tools would add some official moderation in the !privacytools community in the official instance, and/or maybe even hosting an instance. This last part is up to discussion, though.

## My connection with the software

I am a user of the software and administrator in couple communities over there.

- [x] I will keep the issue up-to-date if something I have said changes or I remember a connection with the software.

Mikaela:
It's currently in worth mentioning at https://privacytools.io/providers/social-news-aggregator/

LongJohn-Silver:
@Mikaela Well, still, since now it has reached federation I think it deserves to be listed.

atomGit:
> ... while also aiming at not providing spaces for right wing organizations.

and what about the left-wing organizations?

wylel:
> while also aiming at not providing spaces for right wing organizations.

I want to point out that this is added by the requestor and not in any documentation from Lemmy at all. It would be nice if political views can be kept out of software recommendations.

Lemmy aims to do the same thing Mastodon has done, so it seems, which is fantastic.

LongJohn-Silver:
[...]

Issue #2001, opened 2020-07-25
🆕 Software Suggestion | FreePost


FreddyMarsden:
## Basic Information

**Name:** FreePost
**Category:** Social News Aggregator
**URL:** [freepo.st](https://freepo.st/)
**Source** [NotABug.org](https://notabug.org/zPlus/freepost)

## Description

FreePost is a community of people interested in free software and free culture. Maintained by [Peers](https://peers.community/), it is part of a highly motivated community of people with a strong interest in free culture. 

## Why I am making the suggestion

They have a good [Privacy Policy and ToS](https://freepo.st/freepost.cgi/about), and I think they would fit the site well.

## My connection with the software

I have an account with the site, though I have never communicated with any of the admins.

- [x] I will keep the issue up-to-date if something I have said changes or I remember a connection with the software.

LongJohn-Silver:
Cool little project, although it is so simple that it is a bit confusing to me. I'm all about being able to browse without JS but a little more CSS wouldn't hurt anyone. Just commenting on the aesthetics of it, I think it should be added.

PR #2322, opened 2021-06-01
Promote Lemmy and update broken links for Social Network Aggregators

lrq3000:
## Description

Resolves: #2154
Resolves: #2001
Resolves: #2140

This PR merges what #2154 did with on top some rewording and link fixes. It's a separate PR because I cannot edit #2154 and the original OP deleted the branch in their repo.

Note that Akasha is not online anymore (the website is, but it seems there is no instance of the app). So we may need to consider removing it from Worth Mentioning.

#### Check List 

* Code repository of the project (if applicable):

### DIFF
--- a/_includes/sections/social-news-aggregator.html
+++ b/_includes/sections/social-news-aggregator.html
@@ -1,4 +1,4 @@
-<h1 id="obb" class="anchor"><a href="#social-news-aggregator"><em class="fas fa-link anchor-icon"></em></a> Social News Aggregators</h1>
+<h1 id="obb" class="anchor"><a href="#social-news-aggregator"><i class="fas fa-link anchor-icon"></i></a> Social News Aggregators</h1>
 
 <div class="alert alert-warning" role="alert">
   <strong>If you are currently using a online bulletin board like Reddit, you should pick an alternative here.</strong>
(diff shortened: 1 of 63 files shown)


_includes/sections/social-news-aggregator.html
```diff
@@ -29,20 +29,20 @@ <h1 id="obb" class="anchor"><a href="#social-news-aggregator"><i class="fas fa-l
...
-forum="https://forum.privacytools.io/t/discussion-raddle/1258"
-gitlab="https://gitlab.com/postmill/"
-web="https://raddle.me"
+title="Lemmy"
+image="/assets/img/png/3rd-party/lemmy.svg"
+description="An <a href="https://github.com/dessalines/lemmy/blob/master/LICENSE">AGPL</a>-licensed <a href="https://www.wikipedia.org/wiki/Fediverse">federated</a> link aggregator using the <a href="https://en.wikipedia.org/wiki/ActivityPub">ActivityPub</a> protocol."
+website="https://lemmy.ml/"
+forum="https://forum.privacytools.io/t/"
+github="https://github.com/LemmyNet/lemmy/"
+web="https://lemmy.ml/"
```
dessalines, line 38: I'm not sure what the difference between web and website is here, but it'd probably be better to link to https://join.lemmy.ml/ , as it has the instance list. lemmy.ml is just the flagship instance.

lrq3000: No difference as it was before (both links used to point to dev.lemmy.ml), I didn't know about this link, I edited the website to point to https://join.lemmy.ml/ per your feedback, thank you very much!

Issue #1828, opened 2020-04-13
📝 Correction | rework the file-sharing site


DJCrashdummy:
## Description

*well... this issue is a mixture of a correction and a kind of suggestion:*
how about differnciating between [`file-sharing`](https://www.privacytools.io/software/file-sharing/) tools using 3rd-party services resp. servers and them who are not.

## Why I am making the suggestion

it makes a big difference if a 3rd party is involved and stores the data on its servers or not... similar to messengers which are centralized, federated or p2p.

IMHO tools like [Snapdrop](https://snapdrop.net/) (LAN-sharing with notifications) or [ShareDrop](https://www.sharedrop.io/) (possibility to share files between different networks) are at leasst *"worth mentioning"* because i know a bunch of people who won't setup neither **wormhole** nor **OnionShare** because of convenience and so still use unencrypted mails or other "curious" services for quick file-sharing.
i know, metadata are leaked, but IMHO it is still better than handing over the files itself to a 3rd party.

## My connection with the software

none... i'm just a FOSS- and privacy-enthusiast.

- [x] I will keep the issue up-to-date if something I have said changes or I remember a connection with the software.

## btw

what is **FreedomBox** doing at this site? on the one hand with FreedomBox itself you can't share any file, but on the other hand it's much more than file-sharing and thus would fit anywhere.

what about a general `self-hosting` site and then also add things like [YunoHost](https://yunohost.org/), [Sandstorm](https://sandstorm.io/) and [DPPM](https://dfabric.github.io/dppm/)?

lrq3000:
I have tested ShareDrop, and although it's very easy to use, it has some limitations.

First, it's [only fully compatible with Chrome browser, and only partially with Firefox](https://github.com/cowbell/sharedrop/issues/81) (depending on the network configuration, the Firefox browser may not send notifications).

Secondly, it has an unclear file size limit. The limit is not hard coded, but if the file is too big, it's not going to be transmitted fully. Magic Wormhole and the other solutions provided on PTIO are much more reliable in my experience.

I don't know about Snapdrop, it looks quite promising, but if it's limited to same LAN it's a big limitation.

DJCrashdummy:
> it's only fully compatible with Chrome browser, and only partially with Firefox (depending on the network configuration, the Firefox browser may not send notifications).

well... on firefox webrtc implementation is worked on right now, so perhaps it will get better with the next releases.

[...]

PR #2326, opened 2021-06-02
Add ShareDrop in File Sharing (Worth Mentioning)

lrq3000:
## Description

Resolves: #1828

#### Check List 

* Code repository of the project (if applicable): https://github.com/szimek/sharedrop

### DIFF
--- a/_includes/sections/file-sharing.html
+++ b/_includes/sections/file-sharing.html
@@ -43,4 +43,5 @@
   </li>
   <li><a href="https://github.com/schollz/croc">croc</a> - Easily and securely send arbitrary-sized files from one computer to another. Similar to Magic Wormhole but without dependencies.</li>
   <li><a href="https://freedombox.org/">FreedomBox</a> - Designed to be your own inexpensive server at home. It runs free software and offers an increasing number of services ranging from a calendar or XMPP server, to a wiki, or VPN.</li>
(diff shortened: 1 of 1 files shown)


youdontneedtoknow22:
I'm no expert but:
1- when the recipient had WebRTC blocked, he couldn't get the file
2- when the sender had WebRTC blocked, he couldn't send the file (even after enabling webrtc again, I couldn't send the file somehow lol. So someone might want to check that).
3- I couldn't use the site using Firefox for some reason. Maybe it's because one of the recommended tweaks? I'm not sure. It worked on Brave and Edge, but not Firefox (even when completely disabling uBO and Firefox enhanced tracking protection).
PTIO recommends using Firefox, and blocking WebRTC. So I'm not sure if that's the right website to recommend.

I tell my friends to use Tresorit with a password on their files. It's the only way nontech-savvy people would send me something (or recieve something from me). Other options could be: sending files using Signal or with a PGP encrypted email (automated for 2 protonmail or tutanota users)

lrq3000:
I tested ShareDrop extensively with both Chrome and Firefox. It actually
works better with Firefox, but indeed webrtc was enabled.

I think the recommendation about disabling WebRTC should be re-evaluated,
some of the safest new encrypted web technology such as encrypted
multiparty videoconferencing requires WebRTC.

In any case this recommendation is not mandatory but optional ("it's
probably a good idea to disable WebRTC") so this shouldn't imho prevent
inclusion, furthermore there is no requirement for coherence across all
tips/softwares recommended on PTIO, eg, several recommended plugins can
break javascript but lots of recommended softwares use javascript.

PR #2329, opened 2021-06-02
Add info tooltip for BitWarden to recommend registering through desktop clients

lrq3000:
## Description

Resolves: Suggestion by @ThracianKnight1907 at https://github.com/privacytools/privacytools.io/issues/1915#issuecomment-628417380

#### Check List 

### DIFF
--- a/_includes/sections/password-managers.html
+++ b/_includes/sections/password-managers.html
@@ -9,6 +9,7 @@
   title="Bitwarden - Cloud/Self-host"
   image="/assets/img/svg/3rd-party/bitwarden.svg"
   description="<strong>Bitwarden</strong> is a free and open-source password manager. It aims to solve password management problems for individuals, teams, and business organizations. Bitwarden is among the easiest and safest solutions to store all of your logins and passwords while conveniently keeping them synced between all of your devices. If you don't want to use the Bitwarden cloud, you can easily host your own Bitwarden server."
+  labels="color==info::icon==fas fa-info-circle::tooltip==Prefer registering via desktop clients rather than the website."
   website="https://bitwarden.com/"
   privacy-policy="https://bitwarden.com/privacy/"
   forum="https://forum.privacytools.io/t/discussion-bitwarden/1343"


freddy-m (dismissed):
LGTM

youdontneedtoknow22:
I don't understand that point.
Bitwarden would use a malicious javascript to get the account password from someone signing IN, because he could have some important passwords saved.
But to sign UP using the client, is not important because you already don't have passwords in your account, you're about to make an account.
So the info added should be: Avoid signin in your Bitwared account using the browser.  Sign up, set up your 2FA and never sign in again. Or am I missing something?

lrq3000:
It's not bitwarden the issue but keyloggers in malicious browser's
extensions for example. But yes i should also add sign is using app or
extension, thank you for the suggestion.

youdontneedtoknow22:
> It's not bitwarden the issue but keyloggers in malicious browser's extensions for example. But yes i should also add sign is using app or extension, thank you for the suggestion. Le mer. 9 juin 2021 à 02:19, youdontneedtoknow22 ***@***.***> a écrit :
> […](#)

If the Issue is a keylogger inside the browser, then the whole discussion with the jurisdiction of Bitwarden isn't relevant any more. I believe our friend there was refering to Bitwarden using a malicious javascript to steal the login information for a specific user, done by bitwarden. US Companies maybe forced to do such thing (Lavabit and Snowden Story)

Tbh I'm neither familiar with Keyloggers inside browser's addons nor Bitwarden (I use KeypassXC and Firefox Lockwise). I installed Bitwarden on 2 Keylogger Addons (not malicious, their job is literally to log keystrokes inside the browser). Those were:
Takker: https://addons.mozilla.org/en-US/firefox/addon/tackker-online-keylogger-tool/?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=search
Nifty Keylogger: https://addons.mozilla.org/en-US/firefox/addon/tackker-online-keylogger-tool/?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=search

Takker could log what I type in the urlbar and what I type inside a website. Nifty Keylogger only what I typed inside the website.
Both couldn't log what I typed inside the Bitwarden Addon. So I don't believe we should recommend signin in using the addon for this reason, but rather for malicious javascripts.

lrq3000:
Yes the original suggestion was made in the context of Bitwarden being compromised, but this suggestion is also beneficial for other threats such as keyloggers as you tested, so I think the variety of issues that this tip solves is a good argument to add it, that's why I made this PR :-)

About sign-in, are these keyloggers able to capture autofilled passwords using Bitwarden plug-in? Because that's why I thought the plug-in was safer, and intended to add another tip about.

lrq3000:
I have updated the tip per our discussion above. Please re-evaluate it.

youdontneedtoknow22:
Yup, I belive this fixes the issue with the potential malicious javascripts.

not relevant to Bitwarden but:
if one wants also to avoid keyloggers and other malicious stuff in firefox addons, they should just use the Addons with the recommended Badge on them (covers pretty much every aspect, like donwloading Youtube videos, Blocking ads, sticky notes, etc..). These will always be checked by mozilla developers, each update for their source code as well. So they would be secure (less vulnerabilites and less attack surface) and private (don't have malicious components like keyloggers).

Issue #2307, opened 2021-05-28
❌ Software Removal | Framadrop


JoshTheTofu:
## Why I am making the suggestion

I went to use FramaDrop from the link under the "Worth mentioning" banner within the "File Sharing" section and the website says that FramaDrop no longer exists.

## My connection with the software

I was a potential user of FramaDrop and nothing else.

- [x] I will keep the issue up-to-date if something I have said changes or I remember a connection with the software.

lrq3000:
It's not really down, it's just that FramaDrop doesn't offer the service themselves anymore, but community-hosted instances ("chatons") are still up and running and when you go to FramaDrop they list these instances. That's part of FramaDrop's plan to decentralize some services.

I searched if there is another "central" instance of LuFi but it seems not, so this link remains the best one since it lists instances.

JoshTheTofu:
Hi Irq3000

I now understand. I thought they were completely different services and not just instances.

I agree with you that keeping the link is definitely best.

lrq3000:
In fact I just found that by default Framadrop redirects to a french page by default, so I understand your confusion. I made a PR to link to the english page by default, this should avoid any confusion.

PR #2341, opened 2021-06-11
Framadrop update link to be in english by default

lrq3000:
## Description

Resolves: #2307

#### Check List 

### DIFF
--- a/_includes/sections/file-sharing.html
+++ b/_includes/sections/file-sharing.html
@@ -33,7 +33,7 @@
 <h3>Worth Mentioning</h3>
 
 <ul>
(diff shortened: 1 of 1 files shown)


youdontneedtoknow22:
https://github.com/privacytools/privacytools.io/issues/2406

I'm not sure which english version do you mean, but Disroot may be more trusted than the other providers. Also, we might want to add riseup to the list as they also offer a service to upload files.

lrq3000:
I'm not liking to a particular instance in this PR, the link is still to framasoft but in english, instead of french by default.

Issue #2157, opened 2020-12-18
📝 Correction | https://www.privacytools.io/providers/dns/ - Nixnet Servers Down


vindicatorr:
## Description
"uncensored.any.dns.nixnet.xyz" and "uncensored.ny1.dns.nixnet.xyz" do not work.
"uncensored.lv1.dns.nixnet.xyz" does work.

## Why I am making the suggestion
Simply pointing out the service does not appear to be maintained.
Anycast worked at one point. I moved to ny1, then that also stopped working. Now I'm on the final lv1.

## My connection with the software
User

I believe @Amolith maintains those servers (https://github.com/privacytools/privacytools.io/issues/1206 & https://docs.nixnet.services/User:Amolith). (I don't know if any of that will poke that person).

- [X] I will keep the issue up-to-date if something I have said changes or I remember a connection with the software.

Amolith:
> I believe @Amolith maintains those servers (#1206 & https://docs.nixnet.services/User:Amolith). (I don't know if any of that will poke that person).

Yes, I do maintain the DNS resolvers. I am aware of the reliability issues and have been working on a proper, long-term fix rather than simply slapping a band-aid on them for now.

Part of the issue that I *won't* be able to fix is that those servers are fairly lightweight yet receive a *lot* of traffic. I don't currently have the funds to upgrade them and keep them upgraded in the long run nor do I receive enough donations to cover the cost.

vindicatorr:
And now lv1 is also down. I'm out.

Amolith:
I'm sorry you're frustrated but the servers update, upgrade, and, when required, reboot autonomously to ensure timely security fixes. Never once have I guaranteed 100% uptime. If you'll check back, you'll notice that all three of them work perfectly fine.

* https://uncensored.ny1.dns.nixnet.xyz/dns-query?name=whoami.v4.powerdns.org&type=A
* https://uncensored.lv1.dns.nixnet.xyz/dns-query?name=whoami.v4.powerdns.org&type=A
* https://uncensored.lux1.dns.nixnet.xyz/dns-query?name=whoami.v4.powerdns.org&type=A

Amolith:
Note that I'm not trying to convince you to use my resolvers. Frankly, it would make maintaining them a little easier if fewer people did. I'm simply explaining their behaviour.

wylel:
@Amolith would this be something you would want pulled from the site because of this?

[...]

PR #2404, opened 2021-08-12
Remove NixNet from DNS providers: #2157

Amolith:


Amolith:
As mentioned a couple of times in #2157, these resolvers are most definitely not recommended any longer and they need to be removed from privacytools.io. There is a notice at the top of the page stating this as well.
https://docs.nixnet.services/NixNet_DNS

Amolith:
These servers will be shut down on September 10th.

ph00lt0:
@dngray ^^

